PR #15, opened 2019-01-28
Devel user func

ratnania:


spl/api/codegen/printing.py
```diff
@@ -3,6 +3,7 @@
 
 from sympde.printing.pycode import PythonCodePrinter as SympdePythonCodePrinter
 
+from pyccel.epyccel import get_source_function
```
yguclu: Is the function `get_source_function` actually used in this file?

ratnania: No, not anymore. The commented section in *_print_AppliedUndef* can be removed.

yguclu: Good, you can remove lines 6 and 34-38 and commit your changes. The pull request will be automatically updated.

PR #21, opened 2019-02-07
[api] use find function to create symbolic equations

ratnania:
add user friendly function to create equations, works for linear systems and newton iteration


ratnania:
- provide lhs and rhs as calls to BilinearForm and LinearForm

PR #34, opened 2019-04-08
Devel expr

saidctb:


psydac/api/tests/test_api_expr_2d_scalar.py
```diff
@@ -0,0 +1,73 @@
...
+    # ... discrete spaces
+    Vh = discretize(V, domain_h, degree=degree)
+    # ...
+    a = x*y+pi*div(F)+dot(F,F)
+    # ... dsicretize the bilinear form
+    ah = discretize(a, domain_h, Vh)
+    
+    Fh = VectorFemField(Vh) 
+    x = np.array(range(5))/5
+    y=ah(x,x,Fh)
```
yguclu: When evaluating `ah`, the fields should be passed as keyword arguments (they are free symbols).

PR #37, opened 2019-04-30
Devel feec

saidctb:


psydac/api/discretization.py
```diff
@@ -204,7 +206,116 @@ def solve(self, **kwargs):
 
         return driver_solve(self.linear_system, **settings)
 
+#==============================================================================
+class DiscreteDerham(BasicDiscrete):
```
yguclu: I know that the documentation is very scarce in this file, but can we add a docstring to class declarations at least?

psydac/api/discretization.py
```diff
@@ -204,7 +206,116 @@ def solve(self, **kwargs):
...
+    @property
+    def V2(self):
+        return self._V2
+        
+    @property
+    def V3(self):
+        return self._V3  
+
+#==============================================================================           
+def discretize_derham(V, domain_h, *args, **kwargs):
```
yguclu: Can we add a docstring to this important function? Also, I find that naming a de Rham complex as `V` is a little confusing, as we usually use this name for a function space.

psydac/core/bsplines.py
```diff
@@ -600,5 +600,21 @@ def basis_ders_on_quad_grid( knots, degree, quad_grid, nders ):
             ders = basis_funs_all_ders( knots, degree, xq, span, nders )
             basis[ie,:,:,iq] = ders.transpose()
 
+    if normalize:
```
yguclu: What kind of normalization is this? If we are requiring the basis functions to integrate to 1 instead of having the partition of unity property, then we should explain this carefully in the docstring of this function.

psydac/core/bsplines.py
```diff
@@ -600,5 +600,21 @@ def basis_ders_on_quad_grid( knots, degree, quad_grid, nders ):
...
+        x = scaling_matrix(degree, ne+degree, knots)
+        basis *= x[0]
+
     return basis
 
+#==============================================================================
+def scaling_matrix(p, n, T):
+    """Returns the scaling array for M-splines.
+    It is an array whose elements are (p+1)/(T[i+p+1]-T[i])
+
```
yguclu: We should add to this comment how the `scaling_matrix` is used to convert B-splines into M-splines and vice versa. Moreover, I do not see why we should call it "matrix", given that it is just a list of scalar values.

psydac/feec/derivatives.py
```diff
@@ -1,121 +1,248 @@
...
-from scipy import kron
-from scipy.sparse import identity
-from scipy.sparse import csr_matrix
+from psydac.linalg.stencil import StencilMatrix, StencilVectorSpace
+from psydac.linalg.kron    import KroneckerStencilMatrix as Stencil_kron
+from psydac.linalg.block   import ProductSpace, BlockVector, BlockLinearOperator, BlockMatrix
+from psydac.fem.vector     import ProductFemSpace
 
-def d_matrix(n):
+def d_matrix(n, p, P):
```
yguclu: Why is this function called `d_matrix`? Does the `d` in the name have to do with the *exterior derivative*?

And how is this function used in practice? Which object is going to call it?

More explanation in the docstring would be very helpful.

psydac/feec/derivatives.py
```diff
@@ -1,121 +1,248 @@
...
+    V2 = StencilVectorSpace([n-1], [p], [P])
+    M = StencilMatrix(V1, V2)
+    
+
+    for i in range(n):
+        M._data[p+i,p] = -1.
+        M._data[p+i,p+1] = 1.
+    return M
+    
+def identity(n, p, P):
```
yguclu: Why do we store an identity matrix in stencil form, rather than creating an `IdentityMatrix` class (or `IdentityLinearOperator`)? An object of such a class should require no storage at all (apart from some metadata).

psydac/feec/derivatives.py
```diff
@@ -1,121 +1,248 @@
...
+        Mat = BlockLinearOperator( Vh, Grad_Vh, blocks=mats )
+        self._matrix = Mat
+
+
+
+    def __call__(self, x):
+        return self._matrix.dot(x)
 # user friendly function that returns all discrete derivatives
-def discrete_derivatives(p, n, T):
+def discrete_derivatives(Vh):
```
yguclu: What does this function do?

psydac/feec/old/__init__.py
```diff
@@ -0,0 +1,4 @@
+__all__ = ['utilities','derivatives']
```
yguclu: Why do we need an **old** directory inside **psydac/feec**?

psydac/feec/derivatives.py
saidctb: the letter d here means derivative

psydac/feec/derivatives.py
saidctb: yes we could do that it's ,this is just a first implementation but yes i agree we should use something like `IdentityMatrix`

psydac/feec/old/__init__.py
saidctb: it will be deleted in the futur when we are sure that everything that was implemented in the old folder in updated in the new one

psydac/feec/derivatives.py
saidctb: nothing it's been deleted

psydac/api/discretization.py
saidctb: the name is changed V here is called Complex

psydac/api/discretization.py
saidctb: done

psydac/api/discretization.py
```diff
@@ -204,7 +206,116 @@ def solve(self, **kwargs):
...
+        elif dim == 2:
+            self._V0 = spaces[0]
+            self._V1 = spaces[1]
+            self._V2 = spaces[2]
+        elif dim == 3:
+            self._V0 = spaces[0]
+            self._V1 = spaces[1]
+            self._V2 = spaces[2]
+            self._V3 = spaces[3]
+
```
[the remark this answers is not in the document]
saidctb: it's been fixed

Issue #45, opened 2019-08-21
Do not use SymPDE's function print_expression


yguclu:
Instead, the class `SymbolicExpr` should be used.

PR #50, opened 2019-08-29
Use SymbolicExpr instead of print_expression and pycode.

yguclu:
This solves issue #45


yguclu:
This merge request require the changes in https://github.com/pyccel/sympde/pull/42

PR #51, opened 2019-08-30
Clean up installation procedure with pip

yguclu:
This is an effort to drastically simplify the installation procedure using pip, which now downloads all dependencies from PyPI (with the notable exception of `h5py` and `igakit`). Further, all deprecation warnings have been removed, and the source code passes Codacy's quality review (hence, this pull request solves #52).


psydac/cad/tests/test_geometry.py
```diff
@@ -154,9 +154,24 @@ def test_geometry_1():
...
+        'geo_1.h5',
+        'quart_circle.h5',
+        'quart_circle_0.h5',
+        'quart_circle_1.h5',
+        'circle.h5'
+    ]
+    for fname in filenames:
+        if os.path.exists(fname):
+            os.remove(fname)
+
```
ratnania: it would be better to export the geometry files in a tmp directory and remove its content, rather than doing it for each file.

psydac/api/ast/evaluation.py
```diff
@@ -905,7 +923,7 @@ def _initialize(self):
         # ...
 
         # ...
-        Nj             = ScalarTestFunction(space, name='Nj')
+        Nj             = element_of(space, name='Nj')
```
ratnania: should we keep the name Nj hard coded here? or check if there's an atom of the same name, if it is the case, then we create a random string?

psydac/api/ast/evaluation.py
```diff
@@ -860,7 +878,7 @@ def _initialize(self):
         weights_elements = []
         if self.is_rational_mapping:
             # TODO check if 'w' exist already
-            weights_pts = ScalarField(self.space, name='w')
+            weights_pts = element_of(self.space, name='w')
```
ratnania: should we keep the name w hard coded here? or check if there's an atom of the same name, if it is the case, then we create a random string?

psydac/api/ast/evaluation.py
yguclu: Good point, but I think we should address this in a separate merge request. Would you like to create an issue for this?

ratnania: you can do it and add a link to our discussion

yguclu: See issue #55

psydac/api/ast/evaluation.py
yguclu: See issue #55

psydac/cad/tests/test_geometry.py
yguclu: I actually find it preferable to have each module clean after itself. This way you can run `pytest` on a single file, and the geometry files will be removed at the end even if some tests fail.

PR #60, opened 2020-04-11
Multi-patch domain

saidctb:
In this PR we implemented the parallel Multi-patch solver for scalar equations with analytical mappings.
the following list of changes where introduced:
- add the analytical mapping
- add reverse axis in cart decomposition for the multi-patch
- Implement the interface integrals


psydac/api/ast/fem.py
```diff
@@ -315,6 +322,22 @@ def __init__(self, expr, terminal_expr, spaces, tag=None):
...
 
+
+        if is_broken:
+            if is_bilinear:
+                space_domain = spaces[0].domain
+            else:
+                space_domain = spaces.domain
+
+            if isinstance(domain, Interface):
+                i = space_domain.interior.args.index(domain.minus.domain)
```
yguclu, line 333: What is `domain.minus`? And what is the difference between `domain` and `spaces.domain`?

psydac/api/ast/nodes.py
```diff
@@ -1377,6 +1377,7 @@ def get_geometry_stmts(self, mapping):
 
         if mapping._expressions is not None:
             args += [ComputeLogical(WeightedVolumeQuadrature(l_quad))]
+            args += [ComputeLogical(SymbolicWeightedVolume(mapping))]
```
yguclu, line 1380: What is the `SymbolicWeightedVolume`?

psydac/api/tests/test_api_2d_scalar_mapping_multi_patch.py
```diff
@@ -0,0 +1,207 @@
...
+
+    I = domain.interfaces
+
+    kappa  = 10**2
+
+   # expr_I =(
+   #         - dot(grad(plus(u)),nn)*minus(v)  + dot(grad(minus(v)),nn)*plus(u) - kappa*plus(u)*minus(v)
+   #         + dot(grad(minus(u)),nn)*plus(v)  - dot(grad(plus(v)),nn)*minus(u) - kappa*plus(v)*minus(u)
+   #         - dot(grad(plus(v)),nn)*plus(u)   + kappa*plus(u)*plus(v)
+   #         - dot(grad(minus(v)),nn)*minus(u) + kappa*minus(u)*minus(v))
```
yguclu, line 63: Was this formula wrong? If so, can you please remove it?

psydac/api/tests/test_api_2d_scalar_multi_patch.py
```diff
@@ -0,0 +1,153 @@
...
+    error  = u - solution
+
+    I = domain.interfaces
+
+    kappa  = 10**3
+
+    #expr_I =- dot(grad(plus(u)),nn)*minus(v)  + dot(grad(minus(v)),nn)*plus(u) - kappa*plus(u)*minus(v)\
+    #        + dot(grad(minus(u)),nn)*plus(v)  - dot(grad(plus(v)),nn)*minus(u) - kappa*plus(v)*minus(u)\
+    #        - dot(grad(plus(v)),nn)*plus(u)   + kappa*plus(u)*plus(v)\
+    #        - dot(grad(minus(v)),nn)*minus(u) + kappa*minus(u)*minus(v)
```
yguclu, line 58: Same comment as before: can this be removed?

psydac/ddm/cart.py
```diff
@@ -91,8 +91,11 @@ def __init__( self, npts, pads, periods, reorder, comm=MPI.COMM_WORLD ):
         # ...
         # Know the number of processes along each direction
 #        self._dims = MPI.Compute_dims( self._size, self._ndims )
-        mpi_dims, block_shape = mpi_compute_dims( self._size, npts, pads )
-        self._dims = mpi_dims
+        if nprocs is None:
+            nprocs, block_shape = mpi_compute_dims( self._size, npts, pads )
```
yguclu, line 95: If the`nprocs` argument is given, can you check that it it contains something that makes sense?
For example, we should check that `len(nprocs) == len(npts)` (see assert at line 68)

psydac/ddm/cart.py
```diff
@@ -61,7 +61,7 @@ class CartDecomposition():
         (optional: default is MPI_COMM_WORLD).
 
     """
-    def __init__( self, npts, pads, periods, reorder, comm=MPI.COMM_WORLD ):
+    def __init__( self, npts, pads, periods, reorder, comm=MPI.COMM_WORLD, nprocs=None, reverse_axis=None ):
```
yguclu, line 64: What if we need to reverse more than one axis? Does that never happen?

psydac/linalg/block.py
```diff
@@ -162,6 +162,12 @@ def __isub__( self, v ):
             b1 -= b2
         return self
 
+    # ...
+    def update_ghost_regions( self, *, direction=None ):
+        for vi in self.blocks:
+            if not vi.ghost_regions_in_sync:
+                vi.update_ghost_regions()
```
yguclu, line 169: Shouldn't you use the `direction` argument in this method?

psydac/linalg/block.py
```diff
@@ -264,6 +270,8 @@ def dot( self, v, out=None ):
         else:
             out = BlockVector( self._codomain )
 
+        v.update_ghost_regions()
```
yguclu, line 273: Don't the linear operators `Lij` below call this method on each component `v[j]` already?

psydac/polar/dense.py
```diff
@@ -252,6 +256,10 @@ def __isub__( self, v ):
         self._data -= v._data
         return self
 
+    # ...
+    def update_ghost_regions( self, *, direction=None ):
+        pass
+
```
yguclu, line 262: Same comment as above. Please move this method to the section below.

psydac/polar/dense.py
```diff
@@ -194,6 +194,10 @@ def __init__( self, V, data ):
     def space( self ):
         return self._space
 
+    @property
+    def ghost_regions_in_sync(self):
+        return True
+
```
yguclu, line 200: If you look at the `Vector` base class [here](https://github.com/pyccel/psydac/blob/3a76c8de807ea73272a2be9ffa596fa6b7d9eca7/psydac/linalg/basic.py#L38) you will see that this method is not part of the abstract interface, so it should be moved to after line 265.

psydac/cad/geometry.py
```diff
@@ -23,7 +23,8 @@
 from psydac.fem.tensor       import TensorFemSpace
 from psydac.mapping.discrete import SplineMapping, NurbsMapping
 
-from sympde.topology import Domain, Line, Square, Cube
+from sympde.topology import Domain, Line, Square, Cube, NCubeInterior
```
yguclu, line 26: Do we still use classes `Line`, `Square` and `Cube` in this file? If not, please remove their import

psydac/api/grid.py
```diff
@@ -100,7 +100,7 @@ def ext(self):
         return self._ext
 #==============================================================================
 class BasisValues():
-    def __init__( self, V, grid, nderiv ):
+    def __init__( self, V, grid, nderiv , ext=None):
```
yguclu, line 103: Can you remind me of what `ext` means?

psydac/api/essential_bc.py
```diff
@@ -104,52 +146,126 @@ def apply_essential_bc_2d_StencilMatrix(V, bc, a):
...
 
     # assumes a 3D Tensor space
     # add asserts on the space if it is periodic
 
     V1,V2,V3 = V.spaces
 
-    s1, s2, s3 = a.space.starts
-    e1, e2, e3 = a.space.ends
+    s1, s2, s3 = a.domain.starts
+    e1, e2, e3 = a.domain.ends
```
yguclu, line 159: Assuming that `a.domain` is a discrete domain, what do `starts` and `ends` mean? And how are they different from the values of `starts` and `ends` in `a.space`?

psydac/api/essential_bc.py
```diff
@@ -104,52 +146,126 @@ def apply_essential_bc_2d_StencilMatrix(V, bc, a):
...
+def apply_essential_bc_1d_StencilInterfaceMatrix(V, bc, a):
+    """ Apply homogeneous dirichlet boundary conditions in 1D """
+
+    # assumes a 1D spline space
+    # add asserts on the space if it is periodic
+
+    # get order
+    order = bc.order
+    ext     = bc.boundary.ext
+    d_start = a._d_start
```
yguclu, line 202: What is `d_start`?

psydac/api/essential_bc.py
```diff
@@ -162,51 +278,90 @@ def apply_essential_bc_3d_StencilMatrix(V, bc, a):
 
     # get order
     order = bc.order
+    axis    = bc.boundary.axis
+    ext     = bc.boundary.ext
+    d_start = a._d_start
+    dim     = a._dim
```
yguclu, line 284: What is `dim`?

psydac/api/ast/fem.py
saidctb: space_domain is the global domain the other is where you are calculating the integral

psydac/api/ast/nodes.py
saidctb: it's the symbolic representation of the the weighted volume of a quadrature rule
`wvol = quad_1[k1]*quad_2[k2]` where `SymbolicWeightedVolume` is `wvol`

psydac/api/tests/test_api_2d_scalar_mapping_multi_patch.py
saidctb: it's not wrong, this formula exist in another paper

psydac/ddm/cart.py
saidctb: yes we should add it

psydac/ddm/cart.py
saidctb: it doesn't happen we always want to reverse one at a time , if we have an L shape patches we build the first patch then the second is reversed by one axis of the first aftet that the third is reversed by one axis of the second even though that the third is reversed by 2 axes of the first

psydac/linalg/block.py
saidctb: yes for some raison it locks if we do it in each component and this solves the problem, we also want to separate the communications from the calculations  to be able to optimize them in the future

psydac/polar/dense.py
saidctb: done

psydac/polar/dense.py
saidctb: done

psydac/cad/geometry.py
saidctb: yes we still them they are used in the file

psydac/api/grid.py
saidctb: the direction of the normal vector in the boundary we use it to determine the boundary

psydac/api/essential_bc.py
saidctb: before we had a.domain == a.codomain which mean a.space = a.domain
but if a.domain and a.codomain are different a.space don't have a meaning

psydac/api/ast/fem.py
yguclu: I see, but what is `domain.minus`?

psydac/api/ast/nodes.py
yguclu: Isn't that a `WeightedVolumeQuadrature` instead?

psydac/linalg/stencil.py
```diff
@@ -1038,12 +1038,23 @@ class StencilInterfaceMatrix(Matrix):
...
-    W : psydac.linalg.stencil.StencilVectorSpace
-        Codomain of the new linear operator.
+    W   : psydac.linalg.stencil.StencilVectorSpace
+          Codomain of the new linear operator.
+
+    s_d : int
+          The starting index of the domain.
 
+    s_c : int
+          The starting index if the codomain.
```
yguclu, line 1051: Typo: `if` >> `of`

psydac/linalg/stencil.py
```diff
@@ -1038,12 +1038,23 @@ class StencilInterfaceMatrix(Matrix):
...
+
+    dim : int
+          The axis of the interface.
+
+    pads: <list|tuple>
+          Padding of the lineair operator.
```
yguclu, line 1057: Typo: `lineair` >> `linear`

psydac/api/ast/fem.py
saidctb: an interface is a result of an intersection between two domains, domain_minus and domain_plus

psydac/api/ast/nodes.py
saidctb: this is `quad_1[k1]*quad_2[k2]` the `WeightedVolumeQuadrature`

psydac/linalg/stencil.py
saidctb: fixed

psydac/linalg/stencil.py
saidctb: fixed

psydac/linalg/block.py
yguclu: I am worried about this MPI deadlock: it should not occur unless we have a serious bug, which was maybe introduced by this PR, or had been here before and remained undetected for a very long time. We must investigate this.

psydac/api/ast/nodes.py
yguclu: Sorry, I cannot understand. Can you write the 2 expressions of `WeightedVolumeQuadrature` and `SymbolicWeightedQuadrature` again?

psydac/api/essential_bc.py
yguclu: I see! So our abstract LinearOperator does not have a `space` property anymore?

psydac/api/ast/nodes.py
saidctb: not both the `WeightedVolumeQuadrature` is used once in those lines but `SymbolicWeightedQuadrature` is used in other places for example we  can have something like
`wvol = quad_1[k1]*quad_2[k2]`
and then use it in a expression like
`v += wvol*(u_x*v_x + u_y*v_y)`

psydac/api/essential_bc.py
saidctb: yes

yguclu: Oh yes, there is no `space` anymore: https://github.com/pyccel/psydac/blob/6be6dffd7199ed93a4e83c24244d111ddc5d31bb/psydac/linalg/basic.py#L94

So we had a bug in these two lines of `essential_bc.py` before you changed them. Strange that we missed it.

psydac/linalg/block.py
yguclu: After some discussion we solved the mystery: only some processes in the MPI communicator have an `InterfaceLinearOperator` as an off-diagonal block in the block matrix. When the `dot` method of this operator is called on a vector `v` with `v.ghost_regions_in_sync == False` the method `v.update_ghost_regions()` is called, but not all processes contribute to this collective call. This results in a deadlock.

yguclu: Since `InterfaceLinearOperator` cannot safely call the `update_ghost_regions` method of a `StencilVector` object `v`, I recommend that it raises an error (MPI abort) if it finds that `v.ghost_regions_in_sync == False`.

examples/poisson_2d_multi_patch.py
```diff
@@ -0,0 +1,207 @@
...
+   # expr_I =(
+   #         - dot(grad(plus(u)),nn)*minus(v)  + dot(grad(minus(v)),nn)*plus(u) - kappa*plus(u)*minus(v)
+   #         + dot(grad(minus(u)),nn)*plus(v)  - dot(grad(plus(v)),nn)*minus(u) - kappa*plus(v)*minus(u)
+   #         - dot(grad(plus(v)),nn)*plus(u)   + kappa*plus(u)*plus(v)
+   #         - dot(grad(minus(v)),nn)*minus(u) + kappa*minus(u)*minus(v))
+
+    expr_I =- 0.5*dot(grad(plus(u)),nn)*minus(v)  + 0.5*dot(grad(minus(v)),nn)*plus(u)  - kappa*plus(u)*minus(v)\
+            + 0.5*dot(grad(minus(u)),nn)*plus(v)  - 0.5*dot(grad(plus(v)),nn)*minus(u)  - kappa*plus(v)*minus(u)\
+            - 0.5*dot(grad(minus(v)),nn)*minus(u) - 0.5*dot(grad(minus(u)),nn)*minus(v) + kappa*minus(u)*minus(v)\
+            - 0.5*dot(grad(plus(v)),nn)*plus(u)   - 0.5*dot(grad(plus(u)),nn)*plus(v)   + kappa*plus(u)*plus(v)
```
yguclu, line 71: Do we have a way to write these expressions using jumps and averages?

psydac/api/ast/fem.py
```diff
@@ -235,13 +236,18 @@ def __init__(self, expr, terminal_expr, spaces, tag=None):
...
+
         if isinstance(expr, LinearForm):
+
             is_linear           = True
             tests               = expr.test_functions
             fields              = expr.fields
             mapping             = spaces.symbolic_mapping
             is_rational_mapping = spaces.is_rational_mapping
             spaces              = spaces.symbolic_space
+            is_broken           = spaces.is_broken
```
yguclu, line 250: Does `is_broken` mean that the solution is not C0 across the interface between two patches?

saidctb: yes

examples/poisson_2d_multi_patch.py
saidctb: I have not tried it yet but it might work

saidctb: no it doesn't work we will need to fix it in `Sympde`

yguclu: Could you create an issue in Sympde about it?

saidctb: okay

psydac/fem/grid.py
```diff
@@ -127,7 +127,7 @@ def __init__( self, space, start, end, *, quad_order=None, nderiv=1, normalize=F
         # Quadrature data on extended distributed domain
         self._num_elements = ne
         self._num_quad_pts = len( u )
-        self._spans   = np.array( spans   ) - start
+        self._spans   = np.array( spans   )
```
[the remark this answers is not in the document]
saidctb: yes I have done it for the parallel case, we discussed that at a certain moment , so that the generated code stays the same for serial and parallel case but then i noticed that `example/poisson_2d_mapping` doesn't work anymore so I moved this change to `api/grid.py`

psydac/api/discretization.py
```diff
@@ -200,8 +199,8 @@ def assemble(self, **kwargs):
         self._linear_system = LinearSystem(M, rhs)
 
     def solve(self, **kwargs):
-        settings = _default_solver.copy()
-        settings.update(kwargs.pop('settings', {}))
+        settings = {k:kwargs[k] if k in kwargs else it for k,it in _default_solver.items()}
+        settings.update({it[0]:it[1] for it in kwargs.items() if it[0] not in settings})
```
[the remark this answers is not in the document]
saidctb: you can use solve like this now , `equation_h.solve(**settings)` instead of `equation_h.solve(settings={})`

psydac/fem/grid.py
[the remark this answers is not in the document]
yguclu: I see, thanks!

psydac/api/discretization.py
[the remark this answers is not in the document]
yguclu: OK

PR #72, opened 2020-07-07
Finite Element Exterior Calculus

yguclu:
- Add discrete DeRham sequence operators
- Add 1D/2D/3D push-forwards and pull-backs
- Add 1D/2D/3D global projectors
- Remove Analytical mapping from library, use the one from Sympde instead
- Modify the `VectorFemField` to be callable
- Fix bug in the assembly of periodic basis functions
- Add `KroneckerStencilMatrix` 
- Add Identity operators
- Add tests when pads are different from the self.domain.pads
- Add `transpose()` method to `KroneckerStencilMatrix` and `BlockMatrix`
- Fix `transpose()` method of `StencilMatrix` in the case where the pads are different from self.domain.pads
- Add `tostencil()` method to `KroneckerStencilMatrix`
- Add conversion to PETSc data structure
- Add 1D FEEC unit test w/ periodic and Dirichlet BCs (can also be run as a script)
- Add 3D FEEC unit test w/ periodic BCs

TODO (future work):
- Implement systematic strategy for FEEC with homogeneous Dirichlet BCs on V0 space
- Add 2D FEEC unit test w/ periodic and Dirichlet BCs
- Add 3D FEEC unit test w/ Dirichlet BCs
- Verify that everything works in parallel


yguclu:
> Ok, so I guess that at some point the 'normalization' argument will be called 'basis' ?

I wanted to call it `basis`, but in this low-level module the name `basis` is used to identify an array variable that contains the values of the basis functions. So for now I decided to call it `normalization` instead, in order to avoid confusion. But as you see it is not a boolean variable any more, instead it is a string which can take the possible values `'B'` or `'M'`.

In the `SplineSpace` class in **psydac/fem/splines.py**, which is higher level, I pass the argument `basis` to the constructor. Again, this takes one of the possible values `'B'` or `'M'`.

yguclu:
> A spline space of degree `p` is usually not a subspace of a spline space of degree `p+1` (because of the regularity constraint at the knots), so I guess this should be reformulated as :
> "
> Given the knot sequence of S a spline space of degree p, compute the knot
> sequence of S_0 a spline space of degree p+1 such that u' is in S for all u in S_0.
> Specifically, on bounded domains the first and last knots are repeated in the sequence, and in the periodic case the knot sequence is left unchanged
> "

You are correct, for S to be a subspace of S_0 we would have needed to duplicate all interior knots, which we do not do of course! I will update the docstring.

campospinto:
> > Ok, so I guess that at some point the 'normalization' argument will be called 'basis' ?
> 
> I wanted to call it `basis`, but in this low-level module the name `basis` is used to identify an array variable that contains the values of the basis functions. So for now I decided to call it `normalization` instead, in order to avoid confusion. But as you see it is not a boolean variable any more, instead it is a string which can take the possible values `'B'` or `'M'`.
> 
> In the `SplineSpace` class in **psydac/fem/splines.py**, which is higher level, I pass the argument `basis` to the constructor. Again, this takes one of the possible values `'B'` or `'M'`.

Ok it is clear like that. And indeed, here the two bases correspond to two different normalizations

psydac/mapping/discrete.py
```diff
@@ -11,7 +11,7 @@
 import yaml
 import yamlloader
 
-from psydac.mapping.basic import Mapping
+from sympde.topology.mapping  import Mapping
```
yguclu, line 14: Please add an empty line after this one in order to separate the external libraries from the internal library imports.

psydac/linalg/stencil.py
```diff
@@ -337,6 +337,11 @@ def toarray( self, *, with_pads=False ):
         index = tuple( slice(p,-p) for p in self.pads )
         return self._data[index].flatten()
 
+    # ...
+    def toarray_local( self ):
+        idx = tuple( slice(p,-p) for p in self.pads )
+        return self._data[idx].flatten()
```
yguclu, line 343: This function contains exactly the same code as function `toarray` in the serial case (see 5 lines above). In order to avoid code duplication, can we have one function call the other?

Further, `toarray_local` needs a proper docstring.

psydac/linalg/stencil.py
```diff
@@ -402,6 +407,23 @@ def _toarray_parallel_with_pads( self ):
 #        return out.flatten()
         return out.reshape(-1)
 
+    def topetsc( self ):
```
yguclu, line 410: Please add docstring.

psydac/linalg/stencil.py
```diff
@@ -402,6 +407,23 @@ def _toarray_parallel_with_pads( self ):
...
+        space = self.space
+        assert space.parallel
```
yguclu, line 412: I think it would be very useful if the user could use the PETSc data structure and solvers also in the serial case. Can this be done?

psydac/linalg/stencil.py
```diff
@@ -894,6 +916,59 @@ def _tocoo_no_pads( self ):
 
         return M
 
+    def tocoo_local( self ):
```
yguclu, line 919: Why do you need this function? By calling `tosparse(with_pads=False)` you will get exactly the same non-zero elements in the output sparse matrix, the only difference being that the indices will be global instead of local.

psydac/linalg/stencil.py
```diff
@@ -980,6 +1055,38 @@ def _tocoo_parallel_with_pads( self ):
 
         return M
 
+    # ...
+    def topetsc( self ):
```
yguclu, line 1059: This function needs a docstring.

As I mentioned before, we should really handle the serial case as well.

psydac/linalg/kron.py
```diff
@@ -102,13 +102,23 @@ def dot( self, x, out=None ):
...
     def __getitem__(self, key):
-        rows = key[:2]
-        cols = key[2:]
-        A = self._A1
-        B = self._A2
-        return A[rows[0],cols[0]]*B[rows[1],cols[1]]
+        raise NotImplementedError('TODO')
 
     # ...
     def tocoo( self ):
```
yguclu, line 124: We should remove `tocoo` and `tocsr` and just add the `tosparse` method instead. In linalg we only provide one sparse format of our choice, and we want to be able to change it whenever we want. The user is free to convert the sparse matrix to any other format.

psydac/linalg/kron.py
```diff
@@ -27,73 +29,71 @@ def kron_dot(starts, ends, pads, X, X_tmp, Y, A, B):
...
     # ...
     @property
     def codomain( self ):
-        return self._codomain
-
-    # ...
-    @property
-    def mats(self):
-        return self._mats
+        return self._space
```
yguclu, line 66: This new implementation assumes that domain and codomain are the same. This looks totally wrong to me.

psydac/linalg/block.py
```diff
@@ -201,6 +201,12 @@ def blocks( self ):
     def toarray( self ):
         return np.concatenate( [bi.toarray() for bi in self._blocks] )
 
+    def topetsc( self ):
```
yguclu, line 204: Please add docstring.

Can we make this work also in the serial case?

psydac/linalg/stencil.py
saidctb: I assumed that everything exist in scipy in the serial case, and the user can always call it with one process in the parallel case , but probably yes it can be done,  it needs a little bit of work, because I use the information provided by the cart in ordrer for me to create the petsc cart

psydac/linalg/stencil.py
saidctb: yes I need to local indices here

psydac/linalg/kron.py
saidctb: okay I don't remember why was this changed lol

psydac/api/ast/fem.py
```diff
@@ -64,7 +65,6 @@
 from itertools   import groupby
```
yguclu, line 65: Please reorganize the import order as mentioned in a previous comment.

psydac/linalg/stencil.py
saidctb: okay

psydac/api/ast/parser.py
```diff
@@ -432,6 +432,7 @@ def _visit_DefNode(self, expr, **kwargs):
 
         if self.backend['name'] == 'pyccel':
             a = build_types_decorator(arguments)
+            a = [String(str(i)) for i in a]
```
yguclu, line 435: Can we merge these two lines into one? This way we avoid redefining the variable `a`.

psydac/api/ast/parser.py
```diff
@@ -997,10 +987,10 @@ def _visit_ComputeKernelExpr(self, expr, op=None, lhs=None, **kwargs):
             ext     = target.ext if isinstance(target, Boundary) else 1
 
             J       = LogicalExpr(Jacobian(mapping), mapping=mapping, dim=dim, subs=True)
-            J       = SymbolicExpr(J)
-            values  = [ext * J.cofactor(i, j=axis) for i in range(J.shape[0])]
-            normalization = sum(v**2 for v in values)**0.5
-            values  = [v.simplify() for v in values]
+            J       = SymbolicExpr(J).inv()
```
yguclu, line 990: This variable should be called `J_inv`.

psydac/linalg/kron.py
saidctb: okay

psydac/linalg/kron.py
saidctb: in anyway, we have an assert that the domain is equal to the codomain

psydac/api/ast/parser.py
```diff
@@ -997,10 +987,10 @@ def _visit_ComputeKernelExpr(self, expr, op=None, lhs=None, **kwargs):
...
+            values  = ext * J[axis, :]
+            normalization = values.dot(values)**0.5
+            values  = [v for v in values]
```
yguclu, line 993: It appears that the original algorithm had a bug, and it was missing the determinant factor.

But I am not convinced by these changes. The previous version was explicitly computing one row of the inverse Jacobian matrix; instead you are now computing the whole inverse and then extracting the row, which is significantly slower.

psydac/api/discretization.py
```diff
@@ -280,132 +283,143 @@ def solve(self, **kwargs):
 
 #==============================================================================
 class DiscreteDerham(BasicDiscrete):
-    def __init__(self, *spaces):
+    """
+    Rerpresent the discrete De Rham sequence
```
yguclu, line 287: Typo: Rerpresent -> Represent

psydac/api/discretization.py
```diff
@@ -280,132 +283,143 @@ def solve(self, **kwargs):
...
+            raise NotImplementedError('only global projectors are available')
+
+        if self.dim == 1:
+            P0 = Projector_H1(self.V0)
+            P1 = Projector_L2(self.V1, nquads)
+            return P0, P1
+
+        elif self.dim == 2:
+            raise NotImplementedError('TODO')
+
```
yguclu, line 343: It should not be a problem to provide the projectors in 2D, right?

psydac/api/ast/parser.py
saidctb: yes it can be improved we actually have access to the inverse now by the new modifications in Sympde

psydac/api/discretization.py
saidctb: yes the only problem is we need to decide which sequence the user wants in 2d

psydac/api/grid.py
```diff
@@ -98,9 +98,10 @@ def axis(self):
     @property
     def ext(self):
         return self._ext
+
 #==============================================================================
 class BasisValues():
-    def __init__( self, V, grid, nderiv , ext=None):
+    def __init__( self, V, grid, nderiv , trial=False, ext=None):
```
yguclu, line 104: Can you please add a docstring to this class? I think it deserves one now

psydac/ddm/cart.py
```diff
@@ -226,6 +222,13 @@ def shape( self ):
     def subcomm( self ):
         return self._subcomm
 
+    #---------------------------------------------------------------------------
+    def topetsc( self ):
```
yguclu, line 226: I know that this file does not contain many docstrings... Nevertheless, can you please add one to this method at least?

psydac/ddm/petsc.py
```diff
@@ -0,0 +1,102 @@
...
+#===============================================================================
+class PetscCart:
+
+    def __init__(self, cart):
+        assert isinstance(cart, CartDecomposition)
+
+        try:
+            from petsc4py import PETSc
+        except ImportError:
+            raise
```
yguclu, line 17: Please add a docstring to this new class.

And can we raise a more intelligible error, where we say clearly that `petsc4py` must be installed with `pip` in order to use class `PetscCart`?

psydac/ddm/petsc.py
```diff
@@ -0,0 +1,102 @@
...
+    def create_Ao( self ):
+        cart    = self.cart
+        indices = self.indices
+        return self.petsc.AO().createBasic(indices, comm=cart.comm)
+
+    def create_LGMap( self ):
+        cart    = self.cart
+        indices = self.extended_indices
+        ao      = self.ao
+        return self.petsc.LGMap().create(ao.app2petsc(indices), comm=cart.comm)
```
yguclu, line 90: These four methods are used only by the `__init__` constructor, right? If that's correct, I think they should be "private" methods and their names should start with an underscore `_`

yguclu: A docstring would also be very useful...

psydac/ddm/petsc.py
```diff
@@ -0,0 +1,102 @@
...
+    def create_g2n(self, gvec, natural):
+        cart     = self.cart
+        indices  = self.indices
+        size     = self.local_size
+        start,_  = natural.getOwnershipRange()
+
+        from_is = self.petsc.IS().createStride(size, start, 1, comm=cart.comm)
+        to_is   = self.petsc.IS().createGeneral(indices, comm=cart.comm)
+
+        return self.petsc.Scatter().create(gvec, from_is, natural, to_is)
```
yguclu, line 101: What is this method used for?

psydac/feec/TODO.txt
```diff
@@ -0,0 +1,10 @@
+TODO:
+
+  - add core/interpolation.py (1D)
+  - add core/histopolation.py (1D)
+  - when choosing SplineSpace, choose if it should use N-basis or D-basis, remove normalize flag
+  - rename _init_collocation as _init_interpolation (also in TensorFemSpace)
+  - if N-basis, _interpolator uses collocation matrix
+  - if D-basis, _interpolator uses histopolation matrix
+
+  - remove normalize argument from discretize_space (and everywhere else)
```
yguclu, line 10: Can we please check if we have done everything in this TODO list? If so, I think we can remove this file altogether.

psydac/ddm/petsc.py
saidctb: it's more difficult than that, for it to work  the user needs to install petsc first and then petsc4py , I think the import error should be enough

psydac/feec/global_projectors.py
```diff
@@ -0,0 +1,472 @@
...
+
+        # build the rhs
+        self.func(*self.args, fun)
+
+        if len(self.N)==1:
+            rhs = self.rhs.toarray()
+            coeffs = array_to_stencil(self.N[0].solve(rhs), self.space.vector_space)
+            coeffs.update_ghost_regions()
+        else:
+            coeffs = kronecker_solve(solvers = self.N, rhs = self.rhs)
```
yguclu, line 64: As you can see [here](https://github.com/pyccel/psydac/blob/fb5f9d6552ced9ccc3d552f2fa6f8096bd70d6a8/psydac/linalg/kron.py#L417) this function accepts an `out` keyword argument. Hence this line should read
```python
kronecker_solve(solvers = self.N, rhs = self.rhs, out=coeffs)
```

psydac/feec/global_projectors.py
```diff
@@ -0,0 +1,472 @@
...
+
+        # build the rhs
+        self.func(*self.args, *fun)
+
+        self.rhs.update_ghost_regions()
+
+        coeffs    = BlockVector(self.space.vector_space)
+        coeffs[0] = kronecker_solve(solvers = self.DNN, rhs = self.rhs[0])
+        coeffs[1] = kronecker_solve(solvers = self.NDN, rhs = self.rhs[1])
+        coeffs[2] = kronecker_solve(solvers = self.NND, rhs = self.rhs[2])
```
yguclu, line 129: Same comment as above, `out=coeffs[i]` with `i=0,1,2` should be passed as an argument.

psydac/feec/global_projectors.py
```diff
@@ -0,0 +1,472 @@
...
+
+        # build the rhs
+        self.func(*self.args, *fun)
+
+        self.rhs.update_ghost_regions()
+
+        coeffs    = BlockVector(self.space.vector_space)
+        coeffs[0] = kronecker_solve(solvers = self.NDD, rhs = self.rhs[0])
+        coeffs[1] = kronecker_solve(solvers = self.DND, rhs = self.rhs[1])
+        coeffs[2] = kronecker_solve(solvers = self.DDN, rhs = self.rhs[2])
```
yguclu, line 196: Again, you can pass `out=coeffs[i]`

psydac/feec/global_projectors.py
```diff
@@ -0,0 +1,472 @@
...
+
+        # build the rhs
+        self.func(*self.args, fun)
+
+        if len(self.D) == 1:
+            rhs = self.rhs.toarray()
+            coeffs = array_to_stencil(self.D[0].solve(rhs), self.space.vector_space)
+            coeffs.update_ghost_regions()
+        else:
+            coeffs = kronecker_solve(solvers = self.D, rhs = self.rhs)
```
yguclu, line 258: Again, use `out=coeffs`

psydac/fem/splines.py
```diff
@@ -11,15 +11,37 @@
...
+        xleft  = xgrid[:-1]
+        xright = xgrid[1:]
+    elif xgrid[0] != xA:
+        xleft  = np.array([xgrid[-1] - (xB-xA), *xgrid[:-1]])
+        xright = xgrid
+    elif xgrid[-1] != xB:
+        xleft  = xgrid
+        xright = np.array([*xgrid[1:], xgrid[0] + (xB-xA)])
+
+    return xleft, xright
```
yguclu, line 43: Can you please check if this function is ever used in Psydac? (I think it's a leftover from a previous version of this file)

If it's not used, feel free to remove it.

psydac/ddm/petsc.py
saidctb: go from global ordering  to natural ordering

psydac/fem/splines.py
saidctb: okay

psydac/fem/tensor.py
```diff
@@ -418,52 +435,56 @@ def compute_interpolant( self, values, field ):
 
     def reduce_grid(self, axes=[], knots=[]):
```
yguclu, line 436: Default argument values must be immutable. Please use empty tuples `()`, or set them to `None`.

psydac/feec/tests/todo_test_derivatives.py
```diff
@@ -1,49 +1,105 @@
-from psydac.feec.derivatives import Grad, Curl, Div
-from psydac.fem.tensor       import TensorFemSpace
-from psydac.fem.splines      import SplineSpace
-from psydac.linalg.stencil   import StencilVector
-from psydac.linalg.block     import BlockVector,ProductSpace
-from psydac.fem.vector       import ProductFemSpace
-from psydac.feec.utilities   import Interpolation, interpolation_matrices
+import pytest
```
yguclu, line 1: I think that this file can be removed.

psydac/feec/tests/test_commuting_projections.py
```diff
@@ -0,0 +1,209 @@
+# -*- coding: UTF-8 -*-
```
yguclu, line 1: I think that we should also add tests in 1D and 2D.

psydac/feec/tests/OLD_test_differentiation_matrices.py
```diff
@@ -0,0 +1,417 @@
+import numpy as np
```
yguclu, line 1: We should remove this old file.

psydac/feec/OLD_derivatives.py
```diff
@@ -0,0 +1,318 @@
+# -*- coding: UTF-8 -*-
```
yguclu, line 1: We should remove this old file.

psydac/feec/struphy_example.py
```diff
@@ -0,0 +1,70 @@
+from mpi4py import MPI
```
yguclu, line 1: I think we have a more complete example in our unit tests, right? So we can remove this old file.

psydac/feec/pull_push.py
```diff
@@ -0,0 +1,215 @@
+from sympy import lambdify
```
yguclu, line 1: Every function in this file only works for 3D mappings. Maybe we should call it `push_pull_3d.py`?

We definitely need 1D and 2D versions as well.

Further, there are no docstrings...

psydac/api/tests/test_api_feec_3d.py
```diff
@@ -0,0 +1,206 @@
...
+    return error
+
+###############################################################################
+#            SERIAL TESTS
+###############################################################################
+
+#==============================================================================
+# 3D Maxwell's equations with "Collela" map
+#==============================================================================
+def test_maxwell_3d_1():
```
yguclu, line 156: We need to add a non-periodic test. A standing wave in a cavity would work nicely...

psydac/linalg/stencil.py
yguclu: I see. This function is almost identical to `tosparse`, and it would be nice to avoid such a large code duplication.

Is there not a way to have the `tosparse` function return what you need, maybe by passing the flag `local=True`?

psydac/linalg/kron.py
yguclu: I think that assert is new, it should not be there

psydac/api/ast/parser.py
yguclu: OK, let's fix it in this PR.

psydac/api/discretization.py
yguclu: Sure, but the user tells us explicitly which 2D DeRham sequence is needed when the symbolic DeRham sequence is created.

psydac/ddm/petsc.py
yguclu: But why are you using a try/except clause, if not for giving a nice error for the user??

psydac/ddm/petsc.py
yguclu: I don't think that this method is ever used in Psydac... Can you check?

psydac/api/tests/test_api_feec_3d.py
saidctb: this will need more work in order to apply boundary conditions and it's not something we currently have with the scipy format

psydac/feec/pull_push.py
saidctb: yes we only have the 3d

psydac/feec/global_projectors.py
saidctb: the coeffs[i] is not allocated so it's not gonna work

psydac/feec/pull_push.py
saidctb: we should keep the same name, and add the other cases later

psydac/linalg/kron.py
saidctb: okay this class is for 2d stencil matrix, it's not used at all, so it doen't really matter, we are using the general one, we just need to keep this one for the moment until we improve the general, and then we can delet it

psydac/api/discretization.py
saidctb: yes that's true, we can add it

psydac/ddm/petsc.py
saidctb: done

psydac/linalg/kron.py
yguclu: OK. Please comment it out so that it's clear that we are not using it. Or maybe put it in a separate file to be removed later.

saidctb: I will comment it out and when we don't need it, we can delete it

psydac/api/grid.py
yguclu: Thanks for adding the docstring. But the constructor's docstring should be placed outside of the `__init__` function. In other words, this should be the class' docstring.

psydac/ddm/petsc.py
yguclu: If you use a `try/except` block, you should give a proper error. Otherwise you can just remove the `try/except` block...

saidctb: yes that's true

psydac/linalg/stencil.py
yguclu: Please create a separate issue: "make petsc backend work in serial".

psydac/ddm/petsc.py
saidctb: done

psydac/api/grid.py
saidctb: done

psydac/feec/tests/test_commuting_projections.py
saidctb: done

saidctb (dismissed):

psydac/linalg/stencil.py
yguclu: Thanks for creating #79

psydac/ddm/petsc.py
```diff
@@ -0,0 +1,112 @@
...
+
+import numpy as np
+from itertools import product
+
+from .cart import CartDecomposition
+
+#===============================================================================
+class PetscCart:
+
+    def __init__(self, cart):
```
yguclu, line 11: Can you please add a docstring to this nice class?

As usual this should contain the `Parameters` section, and a list of the properties under the `Attributes` section (https://numpydoc.readthedocs.io/en/latest/format.html#class-docstring).

psydac/api/tests/test_api_feec_3d.py
yguclu: OK. Let's not worry about it in this pull request.

saidctb (approved):

psydac/ddm/petsc.py
yguclu: Let's leave this to issue #79. Things should be properly documented there.

PR #73, opened 2020-09-03
Devel logical expr

saidctb:
In this PR we made sure that Psydac works with the new version of Sympde and Pyccel where :
- we introduced the concept of a mapped domain and a logical domain
- the expression passed to the parser are defined in the logical domain only
- update the api tests
- support sympy 1.5 and 1.6


psydac/api/tests/test_api_glt_2d_scalar.py
```diff
@@ -6,7 +6,7 @@
 from sympde.core import Constant
 from sympde.calculus import grad, dot, inner, cross, rot, curl, div
 from sympde.calculus import laplace, hessian
-from sympde.topology import (dx, dy, dz)
+from sympde.topology import dx, dy, dz, dx1, dx2, dx3
```
yguclu, line 9: Please remove the import of `(dx, dy, dz)`, as they are never used in this file.

yguclu (changes requested):
Great job @saidctb! It only remains to remove `mapping` from this line,
```python
Vh = discretize(V, domain_h, mapping=mapping, degree=degree)
```
because the mapping is now attached to `domain_h`.

psydac/api/tests/test_api_glt_2d_scalar.py
saidctb: done

yguclu (dismissed):

psydac/api/ast/parser.py
```diff
@@ -5,22 +5,24 @@
 from sympy import Add
 from sympy import Abs
 from sympy import Symbol, Idx
+from sympy import Max
+from sympy import Basic
+from sympy.simplify import cse_main
```
yguclu, line 10: It appears that this function is not currently used in the module. Do you still want to import it?

saidctb: ah I forgot to push my commits

yguclu (approved):

PR #74, opened 2020-10-01
Add Pyccel and Numba backends

saidctb:
Psydac now can use Pyccel and Numba to accelerate its pure Python kernels. In addition:

- Drop Python-3.5 support, because Numba does not support it
- Add Python-3.8
- Add parallel tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,11 +25,11 @@ before_install:
   - python -m pip uninstall -y pyccel
   - python -m pip uninstall -y gelato
   - python -m pip install -r requirements.txt
-  - python -m pip install numba
 
 # command to install project
 install:
   - python -m pip install .
+  - python -m pip install numba
   - python -m pip freeze
 
 before_script:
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -25,6 +25,7 @@ before_install:
   - python -m pip uninstall -y pyccel
   - python -m pip uninstall -y gelato
   - python -m pip install -r requirements.txt
+  - python -m pip install numba
```
yguclu, line 28: While Pyccel is a hard dependency for Psydac, Numba is not. Instead, Numba is just a backend for the acceleration of our generated Python code. 

Therefore I think we should install Numba **after** Psydac, just below line 32. (This is what we do in Pyccel: see [here](https://github.com/pyccel/pyccel/blob/master/.travis.yml).)

yguclu (changes requested):

.travis.yml
saidctb: done

yguclu (approved):

PR #93, opened 2021-02-03
Rename ProductSpace as BlockVectorSpace

### LINKED ISSUE #92: Rename ProductSpace class as BlockVectorSpace
In the module `psydac.linalg.block` we define the class `ProductSpace`, but this name clashes with class `ProductSpace` from module `sympde.topology.space`. Therefore we should rename the Psydac class as `BlockVectorSpace`, which is consistent with the other names in the same module.

cc: @ratnania @campospinto @saidctb

yguclu:
Fixes #92. Class `ProductSpace` from module `psydac.linalg.block` caused a name clash with class `ProductSpace` from module `sympde.topology.space`. Therefore we have renamed it as `BlockVectorSpace`, which is consistent with the other names in the same module.

### DIFF
--- a/psydac/api/fem.py
+++ b/psydac/api/fem.py
@@ -5,47 +5,44 @@
 
 # TODO: avoid using os.system and use subprocess.call
 
-
-from sympde.expr     import BasicForm as sym_BasicForm
-from sympde.expr     import BilinearForm as sym_BilinearForm
-from sympde.expr     import LinearForm as sym_LinearForm
-from sympde.expr     import Functional as sym_Functional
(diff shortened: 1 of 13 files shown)


campospinto (approved):
New naming seems more consistent indeed: Vector -> VectorSpace /  BlockVector -> BlockVectorSpace ...

yguclu:
> New naming seems more consistent indeed: Vector -> VectorSpace / BlockVector -> BlockVectorSpace ...

In fact this was your suggestion in the first place! :smile:

saidctb (approved):

PR #96, opened 2021-02-04
Don't use SymPDE's obsolete classes for fields

yguclu:
The following classes of SymPDE are obsolete, as explained [here](https://github.com/pyccel/sympde/issues/50):

- `ScalarField`,
- `VectorField`,
- `IndexedVectorField`.

With this PR we stop using those classes in Psydac. Moreover:

- Remove unused classes `EvalQuadratureMapping` and `EvalQuadratureField` from `api.ast.evaluation`;
- Fix bug in `initialize` method of `GltKernel` when working with fields;
- Use SymPDE version 0.10.10 (fixes bug: fields were wrongly decomposed as tensor products of basis functions);
- Use GeLaTo version 0.11 (does not use obsolete classes).


psydac/api/ast/fem.py
```diff
@@ -284,52 +251,21 @@ def __init__(self, expr, terminal_expr, spaces, tag=None, **kwargs):
             atomic_expr_field = {f:[] for f in fields}
             for i_row in range(0, n_rows):
                 for i_col in range(0, n_cols):
-                    d            = get_max_logical_partial_derivatives(terminal_expr[i_row,i_col])
-                    nderiv       = max(nderiv, max(d.values()))
-                    atoms        = _atomic(terminal_expr[i_row, i_col], cls=atoms_types+_logical_partial_derivatives)
-                    fields_atoms = [atom for atom in atoms if is_scalar_field(atom) or is_vector_field(atom)]
```
saidctb, line 290: can you keep this part, we need it, it's a bug that must be fixed not deleted

psydac/api/ast/evaluation.py
```diff
@@ -22,557 +18,10 @@
 
 from .basic     import SplBasic
 from .utilities import build_pythran_types_header, variables
-from .utilities import filter_loops, filter_product, select_loops
 from .utilities import rationalize_eval_mapping
 from .utilities import compute_atoms_expr_mapping
 from .utilities import compute_atoms_expr_field
 
-#==============================================================================
-class EvalQuadratureMapping(SplBasic):
```
saidctb, line 31: can you keep this, we have a glt test which is marked `xfail`, it needs to be fixed

psydac/api/ast/evaluation.py
```diff
@@ -22,557 +18,10 @@
...
-            decorators = {'jit':[]}
-        elif self.backend['name'] == 'pythran':
-            header = build_pythran_types_header(self.name, func_args)
-
-        return FunctionDef(self.name, list(func_args), [], body,
-                           decorators=decorators,header=header)
-
-
-#==============================================================================
-class EvalQuadratureField(SplBasic):
```
saidctb, line 306: the same comment as above

psydac/api/ast/evaluation.py
```diff
@@ -22,557 +18,10 @@
...
-        elif self.backend['name'] == 'pythran':
-            header = build_pythran_types_header(self.name, func_args)
-
-        return FunctionDef(self.name, list(func_args), [], body,
-                           decorators=decorators,header=header)
-
-
-
-#==============================================================================
-class EvalQuadratureVectorField(SplBasic):
```
saidctb, line 444: same comment

psydac/api/ast/fem.py
yguclu: I have only removed one line, where `fields_atoms` is defined.

psydac/api/ast/evaluation.py
yguclu: We are not using this code anywhere in Psydac.
Are you sure that we need all of this just for the GLT expressions?
Can't we use the same machinery that we use everywhere else?

psydac/api/ast/fem.py
saidctb: yes we need it, to know what atoms are used

yguclu: But the functions `is_scalar_field` and `is_vector_field` do not exist anymore!

saidctb: they need to be changed accordingly, and check if the test function in each atom are part of the fields, something like
`fields_atoms = [a for a in atoms if  get_test_function(a) in fields]`

psydac/api/ast/evaluation.py
saidctb: the glt uses the old mechanism, what I implemented works only for fem

psydac/api/ast/fem.py
yguclu: The fact that the code was wrong does not change the fact that it is dead code. It is **never** used in all our tests...

If you add some failing tests I will put this code back :wink:

psydac/api/ast/evaluation.py
yguclu: I see. Do we need really a separate process for GLT expressions? What you have done for FEM cannot be reused to some extent?

saidctb: it can, but it needs time

psydac/api/ast/fem.py
saidctb: yes ,it should be fixed not deleted

saidctb (approved):

saidctb (approved):

psydac/api/ast/fem.py
yguclu: This should be fixed now.

saidctb (approved):

PR #100, opened 2021-02-12
Make DiscreteEquation.solve() return a FemField

yguclu:
Now the method `solve` of class `DiscreteEquation` from module `psydac.api.discretization` returns a `FemField` object instead of a `Vector` of coefficients.

### DIFF
--- a/psydac/api/tests/test_api_2d_fields.py
+++ b/psydac/api/tests/test_api_2d_fields.py
@@ -87,9 +87,7 @@ def run_field_test(filename, f):
 
     # Solve linear system
     # uh is the L2-projection of the analytical field "f"
-
-    x  = equation_h.solve()
-    uh = FemField( Vh, x )
+    uh = equation_h.solve()
 
     #+++++++++++++++++++++++++++++++
     l1   = LinearForm( v, integral(domain, F*v))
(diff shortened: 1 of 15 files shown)


saidctb (approved):

psydac/api/tests/test_api_2d_fields.py
```diff
@@ -162,15 +160,14 @@ def run_non_linear_poisson(filename, comm=None):
...
+    u_h = FemField(Vh)
 
     # Newton iteration
     for n in range(N):
 
         print()
         print('==== iteration {} ===='.format(n))
-        X = equation_h.solve(u=u_h)
-        du_h = FemField(Vh, X)
+        du_h = equation_h.solve(u=u_h)
```
yguclu, line 170: @campospinto What do you think of this? The result is a FemField now...

psydac/api/tests/test_api_2d_fields.py
```diff
@@ -181,7 +178,7 @@ def run_non_linear_poisson(filename, comm=None):
             break
 
         # update field
-        u_h.coeffs[:,:] += X[:,:]
+        u_h += du_h
```
yguclu, line 181: @campospinto Adding two FemFields is possible!

psydac/api/tests/test_api_2d_fields.py
campospinto: That's nice indeed. More consistent with the fact that u is a FemField also.

PR #103, opened 2021-02-24
Add H1 push-forwards for consistency

campospinto:
H1 'push-forward' added, for consistency

### DIFF
--- a/psydac/feec/pull_push.py
+++ b/psydac/feec/pull_push.py
@@ -20,7 +20,7 @@ __all__ = (
     # ----------------------
 #   'push_1d_h1',
     'push_1d_l2',
-#   'pull_2d_h1',
+    'push_2d_h1',
     'push_2d_hcurl',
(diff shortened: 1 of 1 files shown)


psydac/feec/pull_push.py
```diff
@@ -336,9 +336,22 @@ def fun(xi1, xi2, xi3):
...
+#==============================================================================
+
 #==============================================================================
 # 1D PUSH-FORWARD
 #==============================================================================
+def push_1d_h1(func, xi1):
+
+    value     = func(xi1)
+
+    return value
```
yguclu, line 353: Differently than in the other push-forwards, here there is no need to create a temporary variable called `value`. So we can just write
```python
def push_1d_h1(func, xi1):
    return func(xi1)
```
and similarly for the 2D and 3D cases.

yguclu (changes requested):

yguclu (approved):

PR #104, opened 2021-02-25
Fix bug in application of essential boundary conditions to linear systems

yguclu:
- Fix bug in `apply_essential_bc_{2|3}d_StencilMatrix` when `domain != codomain`;
- Add consistency checks to the same functions;
- Clean up functions`apply_essential_bc_{1|2|3}d_StencilVector` for consistency.

### DIFF
--- a/psydac/api/essential_bc.py
+++ b/psydac/api/essential_bc.py
@@ -10,92 +10,115 @@ from psydac.linalg.block   import BlockVector, BlockMatrix
 def apply_essential_bc_1d_StencilVector(V, bc, a):
     """ Apply homogeneous dirichlet boundary conditions in 1D """
 
-    # assumes a 1D spline space
-    # add asserts on the space if it is periodic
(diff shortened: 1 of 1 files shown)


saidctb:
this PR will create a conflict with what we have in the branch feec-multipatch, so it's probably better to have It there

yguclu:
I don't see why, I have only touched the low-level functions. I will take care of any conflicts anyway.

yguclu:
OK, I see the problem now. You've added some keywords to those functions. The problem is that those changes should have been done in a separate PR, not in a feature branch.

yguclu:
We cannot stop development in `devel` because of one feature branch. If you would like me to make further improvements to this PR I can do so. Let's discuss about it.

saidctb (approved):

campospinto:
Thanks Yaman for spotting and fixing this bug!

PR #107, opened 2021-03-10
Curl-conga

saidctb:
In this PR  we made the following changes:
- Add multipatch test for a system of equations


saidctb:
> here you should distinguish between the constant parameter, and the one that depends on h.
> So, write for instance :
> 
> ```
> kappa = 10**3 if kappa is None
> kappa_h = kappa*ncells[0]
> ```
> 
> (and maybe call it `alpha`, as in the papers on IP-DG Maxwell)

okay I can do that

PR #110, opened 2021-04-12
File generation for Igakit geometry 

saidctb:
Add file conversion and geometry manipulation capabilities to module `psydac.cad.geometry`:

- New function `import_geopdes_to_nurbs` reads GeoPDEs geometry file and converts it to igakit NURBS object;
- New function `refine_nurbs` refines igakit NURBS object through knot insertion and degree elevation;
- New function `export_nurbs_to_hdf5` takes Igakit geometry object and generates Psydac geometry file (hdf5);
- New file `psydac/cad/geometry/tests/test_geometry.py` contains unit tests.

Moreover:

- Add unit test to `test_2d_scalar_mapping.py` on 2d pipe domain (loads new geometry file `pipe.h5`);
- In class `FemAssemblyGrid` of module `psydac.fem.grid`, generalize calculation of local domain elements to take into account repeated knots in knot sequence (necessary for parallel runs).


psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
 
         # Close HDF5 file
         h5.close()
+
+#==============================================================================
+def export_geo(filename, patch, periodic=None, comm=None ):
```
yguclu, line 332: Please add a good docstring. What is the type of the `patch` argument?

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+    for d in range( patch.dim ):
+        group['knots_{}'.format( d )] = patch.knots[d]
+    group['points'] = patch.points
+    if rational:
+        group['weights'] = patch.weights
+
+    h5.close()
+
+#==============================================================================
+def refine(nrb, ncells=None, degree=None , periodic=None):
```
yguclu, line 407: Please add a good docstring, which should explain a few points:

- What is the type of the `nrb` argument?
- Is the output spline equal to the input one at each point?
- How do you use the information in `ncells` and `degree` to refine the knot sequence?
- Is the `periodic` argument actually used?

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+
+    if degree is not None:
+        for axis in range(0,nrb.dim):
+            d = degree[axis] - nrb.degree[axis]
+            nrb.elevate(axis, times=d)
+
+    return nrb
+
+#==============================================================================
+def create_geometry_file(nrb, filename, ncells, degree, periodic=None):
```
yguclu, line 429: Please add a complete docstring

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+    nrb = nrb.clone()
+    if ncells is not None:
+
+        for axis in range(0,nrb.dim):
+            ub = nrb.knots[axis][0]
+            ue = nrb.knots[axis][-1]
+            knots = list(np.linspace(ub,ue,ncells[axis]+2)[1:-1])
+            for k in nrb.knots[axis]:
+                if k in knots:
+                    knots.remove(k)
```
yguclu, line 418: Given that `knots` is a list of floating-point numbers, shouldn't we use a tolerance instead of checking for equality with `if k in knots`?

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+            nrb.refine(axis, knots)
+
+    if degree is not None:
+        for axis in range(0,nrb.dim):
+            d = degree[axis] - nrb.degree[axis]
+            nrb.elevate(axis, times=d)
```
yguclu, line 424: What happens if `degree` is smaller than the original spline degree?

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+
+    def _read_header(self, line):
+        chars = line.split(" ")
+        data  = []
+        for c in chars:
+            try:
+                data.append(int(c))
+            except:
+                pass
+        return data
```
yguclu, line 484: This method does not use `self`, so it should be a `@staticmethod`. Or just a separate function.

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+
+    def _extract_patch_line(self, lines, i_patch):
+        text = "PATCH " + str(i_patch)
+        for i_line,line in enumerate(lines):
+            r = line.find(text)
+            if r != -1:
+                return i_line
+        return None
```
yguclu, line 492: Same comment, this could be a `@staticmethod` or a separate function

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+
+    def _get_begin_line(self, lines, n_patchs):
+        list_begin_line = []
+        for i_patch in range(0, n_patchs):
+            r = self._extract_patch_line(lines, i_patch+1)
+            if r is not None:
+                list_begin_line.append(r)
+            else:
+                raise ValueError(" could not parse the input file")
+        return list_begin_line
```
yguclu, line 502: Same comment, this could be a `@staticmethod` or a separate function (once `_extract_patch_line` is)

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+        data  = []
+        for c in chars:
+            try:
+                data.append(int(c))
+            except:
+                try:
+                    data.append(float(c))
+                except:
+                    pass
+        return data
```
yguclu, line 515: Same comment, this could be a `@staticmethod` or a separate function

psydac/cad/geometry.py
```diff
@@ -326,3 +327,216 @@ def export( self, filename ):
...
+    # ...
+
+    extension = os.path.splitext(filename)[1]
+    if not extension == '.h5':
+        raise ValueError('> Only h5 extension is allowed for filename')
+
+    export_geo(filename, nrb, periodic=periodic)
+
+#==============================================================================
+class geopdes(object):
```
yguclu, line 445: 1. Use capitalized names for classes, for example `GeoPDEs` in this case.
2. Please add a docstring that explains what this class does
3. If I understand correctly the only goal of this class is to convert a GeoPDEs file to a IgaKit NURBS object, but do you actually need multiple instances of this "converter"? If not, then you do not need a class at all, and you can just write a single function...

psydac/fem/grid.py
```diff
@@ -156,6 +145,18 @@ def __init__( self, space, start, end, *, quad_order=None, nderiv=1 , pad=None):
...
+        # LOCAL GRID, PROPER (WITHOUT GHOST REGIONS)
+        #-------------------------------------------
+
+        # Local indices of first/last elements in proper domain
+        if space.periodic:
+            local_element_start = self._spans.searchsorted( degree + start )
+            local_element_end   = self._spans.searchsorted( degree + end   )
+        else:
+            local_element_start = self._spans.searchsorted( degree if start == 0   else 1 + start)
+            local_element_end   = self._spans.searchsorted( end if end   == n-1 else 1 + end )
```
yguclu, line 158: Could you explain why `numpy.searchsorted` should be used here, instead of `list.index`?

yguclu (changes requested):
Well done @saidctb, this is extremely useful!

I have a few comments and questions...

psydac/cad/geometry.py
saidctb: yes you are right I will just convert it to a function

psydac/fem/grid.py
saidctb: `numpy.searchsorted` works exactly like `list.index` if the element exist in the list otherwise it will search for the smallest index where the value inserted in it won't break the sorted array.
 when we have repeated knots in a knot sequence, the searched element may not be in the span , so the `list.index` will raise an error of not found element

psydac/cad/geometry.py
saidctb: it will raise an error of this kind
```python
   1137         if times is None: return self
   1138         times = int(times)
-> 1139         assert times >= 0
   1140         if times == 0: return self
   1141         axis = range(self.dim)[axis]
```

psydac/cad/geometry.py
saidctb: yes you are right

psydac/cad/geometry.py
saidctb: done, the type of patch is `igakit.nurbs.NURBS`

psydac/cad/geometry.py
saidctb: done, the periodic object is never used so I deleted it

psydac/cad/geometry.py
saidctb: done

psydac/cad/geometry.py
saidctb: done

psydac/cad/geometry.py
saidctb: I turned it into a function

psydac/cad/geometry.py
saidctb: I turned it into a function

psydac/cad/geometry.py
saidctb: I turned it into a function

psydac/cad/geometry.py
saidctb: I turned it into a function

psydac/cad/geometry.py
yguclu: Could we raise an error in Psydac with a nice error message, instead of relying on Igakit obsure errors?

psydac/cad/geometry.py
yguclu: I still have this question:
> Is the output spline equal to the input one at each point?

In other words, the spline function is still the same point-wise, even though it is now represented in a different basis?

saidctb: yes we use knot insertion, so the shape of the curve is not changed.

yguclu (changes requested):
I've had another look at module `psydac.cad.geometry`, and I'm a bit troubled by the amount of duplicated code.

Given that we have a `Geometry` class already, why don't we just add a static-method constructor to it, called `from_igakit_nurbs`, and then use the `export` method already existing to write to file?

This is probably also a good opportunity to clean up the `Geometry` class, if you see anything that you do not like. The `__init__` is far from optimal for sure...

psydac/cad/geometry.py
yguclu: Please add this comment to the description, I think it's very important

psydac/cad/geometry.py
saidctb: done

psydac/cad/geometry.py
yguclu: Thanks

saidctb:
> I've had another look at module `psydac.cad.geometry`, and I'm a bit troubled by the amount of duplicated code.
> 
> Given that we have a `Geometry` class already, why don't we just add a static-method constructor to it, called `from_igakit_nurbs`, and then use the `export` method already existing to write to file?
> 
> This is probably also a good opportunity to clean up the `Geometry` class, if you see anything that you do not like. The `__init__` is far from optimal for sure...

for that we will need to do the opposite of what we are doing now, we now export the igakit geometry to an hdf5 file, and then we read it and convert it to a Geometry object, your suggestion will do the opposite, it will read the igakit object then convert it to a Geometry object, you can then export it to an hdf5 file.
I think it's better to keep them as they are now, the Geometry class should be responsible only for reading and writing hdf5 files, otherwise, each time we add a new format we will need to add it to the class, these functions are functions that help the user, the user can directly create the hdf5 file without using our functions.

also creating a Geometry object directly  will make things more complicated, as we need to create all the proper spaces and different discrete mappings.

psydac/cad/geometry.py
```diff
@@ -326,3 +319,308 @@ def export( self, filename ):
...
+def export_geo(filename, patch, periodic=None, comm=None ):
+
+    """
+    This function exports an hdf5 geometry file from a single patch nurb object
+
+    Parameters
+    ----------
+
+    filename : <str>
+        the filename of the exported geometry
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
```diff
@@ -326,3 +319,308 @@ def export( self, filename ):
...
+
+    patch   : <igakit.nurbs.NURBS>
+        igakit geometry nurbs object that represent one patch
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
```diff
@@ -326,3 +319,308 @@ def export( self, filename ):
 
         # Close HDF5 file
         h5.close()
+
+#==============================================================================
+def export_geo(filename, patch, periodic=None, comm=None ):
+
+    """
+    This function exports an hdf5 geometry file from a single patch nurb object
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
saidctb: done

yguclu:
> > I've had another look at module `psydac.cad.geometry`, and I'm a bit troubled by the amount of duplicated code.
> > Given that we have a `Geometry` class already, why don't we just add a static-method constructor to it, called `from_igakit_nurbs`, and then use the `export` method already existing to write to file?
> > This is probably also a good opportunity to clean up the `Geometry` class, if you see anything that you do not like. The `__init__` is far from optimal for sure...
> 
> for that we will need to do the opposite of what we are doing now, we now export the igakit geometry to an hdf5 file, and then we read it and convert it to a Geometry object, your suggestion will do the opposite, it will read the igakit object then convert it to a Geometry object, you can then export it to an hdf5 file.
> I think it's better to keep them as they are now, the Geometry class should be responsible only for reading and writing hdf5 files, otherwise, each time we add a new format we will need to add it to the class, these functions are functions that help the user, the user can directly create the hdf5 file without using our functions.
> 
> also creating a Geometry object directly will make things more complicated, as we need to create all the proper spaces and different discrete mappings.

I see your point. You would like our `Geometry` class to be totally independent of other file formats, which makes a lot of sense from a design standpoint.

I, on the other hand, was thinking about making our file easier by using our `Geometry` class as a single HDF5 printer, instead of duplicating code as is happening now. But you're right to point out that this class requires additional information that what is strictly needed for printing.

psydac/fem/grid.py
yguclu: Interesting. I remember making sure that the `span` argument was correct also for repeated knots.

Could you give me a 1D knot sequence example where your fix is necessary?

saidctb: I had this problem with the pipe.h5 file when we run it in parallel,
where we have this knot sequence `[0.    0.    0.    0.125 0.125 0.25  0.25  0.375 0.375 0.5   0.5   0.625
 0.625 0.75  0.75  0.875 0.875 1.    1.    1.   ]`

yguclu (changes requested):
I think that we should give better (verbose) names to the new functions in file `psydac/cad/geometry.py`. For example:

- `geopdes` -> `import_geopdes_to_nurbs`
-  `refine` -> `refine_nurbs`
- `export_geo` -> `export_nurbs_to_hdf5`

Moreover, I really do not see the point of having a function `create_geometry_file` which combines `refine` and `export_geo`. This is just hiding the refinement step, which we should be shown explicitly instead.

Finally, there is no unit test for function `geopdes`, you should add one.

psydac/fem/grid.py
yguclu: I see, are you running `test_create_geometry_file` in parallel then?

psydac/cad/geometry.py
```diff
@@ -326,3 +319,311 @@ def export( self, filename ):
...
+
+    n_lines_per_patch = 3*n_dim + 1
+
+    list_begin_line = get_begin_line(lines, n_patchs)
+
+    nrb = read_patch(lines, 1, n_lines_per_patch, list_begin_line)
+
+    return nrb
+
+def read_header(line):
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
```diff
@@ -326,3 +319,311 @@ def export( self, filename ):
...
+    chars = line.split(" ")
+    data  = []
+    for c in chars:
+        try:
+            data.append(int(c))
+        except:
+            pass
+    return data
+
+def extract_patch_line(lines, i_patch):
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
```diff
@@ -326,3 +319,311 @@ def export( self, filename ):
...
+    text = "PATCH " + str(i_patch)
+    for i_line,line in enumerate(lines):
+        r = line.find(text)
+        if r != -1:
+            return i_line
+    return None
+
+def get_begin_line(lines, n_patchs):
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
```diff
@@ -326,3 +319,311 @@ def export( self, filename ):
...
+    list_begin_line = []
+    for i_patch in range(0, n_patchs):
+        r = extract_patch_line(lines, i_patch+1)
+        if r is not None:
+            list_begin_line.append(r)
+        else:
+            raise ValueError(" could not parse the input file")
+    return list_begin_line
+
+def read_line(line):
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
```diff
@@ -326,3 +319,311 @@ def export( self, filename ):
...
+        try:
+            data.append(int(c))
+        except:
+            try:
+                data.append(float(c))
+            except:
+                pass
+    return data
+
+def read_patch(lines, i_patch, n_lines_per_patch, list_begin_line):
```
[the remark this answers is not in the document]
saidctb: done

psydac/fem/grid.py
saidctb: I'm talking about the tests in  `test_2d_api_scalar_mapping.py`

setup.py
```diff
@@ -17,7 +17,7 @@
 
 NAME    = 'psydac'
 VERSION = mod.__version__
-AUTHOR  = 'Ahmed Ratnani, Jalal Lakhlili, Yaman Güçlü'
+AUTHOR  = 'Ahmed Ratnani, Jalal Lakhlili, Yaman Güçlü, Said Hadjout'
```
yguclu, line 20: :wink:

yguclu (approved):

yguclu:
@saidctb Good work!

This PR can be merged once the description is updated (this is needed because of the latest changes).

PR #114, opened 2021-04-15
Fix assembly code when multiple FemField objects are free arguments

### LINKED ISSUE #113: Wrong assembly code is generated when multiple FemField objects are free arguments
Consider this example

```python
from sympde.topology import Line
from sympde.topology import ScalarFunctionSpace
from sympde.topology import element_of
from sympde.expr     import BilinearForm
from sympde.expr     import LinearForm
from sympde.expr     import integral

from psydac.fem.basic          import FemField

def test_multiple_fields():

    domain = Line()
    V = ScalarFunctionSpace('V', domain)
    u = element_of(V, name='u')
    v = element_of(V, name='v')

    f1 = element_of(V, name='f1')
    f2 = element_of(V, name='f2')

    g = 0.5 * (f1**2 + f2)
    a = BilinearForm((u, v), integral(domain, u * v * g))
    l = LinearForm(v, integral(domain, g * v))

    ncells = (5,)
    degree = (3,)
    domain_h = discretize(domain, ncells=ncells)
    Vh = discretize(V, domain_h, degree=degree)
    ah = discretize(a, domain_h, [Vh, Vh])
    lh = discretize(l, domain_h, Vh)

    fh = FemField(Vh)
    fh.coeffs[:] = 1

    # Assembly call should not crash if correct arguments are used
    A = ah.assemble(f1=fh, f2=fh)
    b = lh.assemble(f1=fh, f2=fh)
```

The generated assembly functions, for both the linear and bilinear forms, yield the wrong result. This is because they only use the values of one of the fields:

```python
        l_vec_v_rj41a6be[ : ] = 0.0
        for i_basis_1 in range(0, 1 + test_v_p1, 1):
            for i_quad_1 in range(0, k1, 1):
                x1 = local_x1[i_quad_1]
                w1 = local_w1[i_quad_1]
                v_1 = global_test_basis_v_1[i_element_1,i_basis_1,0,i_quad_1]
                v_1_x1 = global_test_basis_v_1[i_element_1,i_basis_1,1,i_quad_1]
                wvol_M_Line_x1 = w1
                v = v_1
                v_x1 = v_1_x1
                f2 = arr_f2[i_quad_1]
                f2_x1 = arr_f2_x1[i_quad_1]
                l_vec_v_rj41a6be[i_basis_1] += v*wvol_M_Line_x1*(0.5*f2 + 0.5*f1**2)
```

Here `f2` is correctly extracted from `arr_f2` but the same does not happen for `f1`.

yguclu:
Fix bug in class `AST` from module `psydac.api.ast.fem`: in the generated code for the assembly of linear and bilinear forms, only the instructions for evaluating the last `FemField` argument were printed. We now print all the instructions required instead. This fixes #113.

### DIFF
--- a/psydac/api/ast/fem.py
+++ b/psydac/api/ast/fem.py
@@ -440,15 +440,17 @@ def _create_ast_bilinear_form(terminal_expr, atomic_expr_field,
     ind_dof_test  = index_dof_test.set_length(LengthDofTest(tests[0])+1)
     # ...........................................................................................
     eval_mapping = EvalMapping(ind_quad, ind_dof_test, d_tests[tests[0]]['global'], d_tests[tests[0]]['global'], mapping, geo, spaces[1], tests, nderiv, mask, is_rational_mapping)
-    g_stmts      = [eval_mapping]
 
+    eval_fields = []
     for f in fields:
         f_ex         = expand([f])
         coeffs       = [CoefficientBasis(i)    for i in f_ex]
         l_coeffs     = [MatrixLocalBasis(i)    for i in f_ex]
         ind_dof_test = index_dof_test.set_length(lengths_fields[f]+1)
         eval_field   = EvalField(atomic_expr_field[f], ind_quad, ind_dof_test, d_fields[f]['global'], d_fields[f]['global'], coeffs, l_coeffs, g_coeffs[f], [f], mapping, pads, nderiv, mask)
-        g_stmts     += [eval_field]
+        eval_fields += [eval_field]
+
+    g_stmts = [eval_mapping, *eval_fields]
 
     # sort tests and trials by their space type
     test_groups  = regroup(tests)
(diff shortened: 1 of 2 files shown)


psydac/api/ast/fem.py
```diff
@@ -474,8 +476,9 @@ def _create_ast_bilinear_form(terminal_expr, atomic_expr_field,
             for v in sub_tests+sub_trials:
                 stmts += construct_logical_expressions(v, nderiv)
 
-            if atomic_expr_field:
-                stmts += list(eval_field.inits)
```
saidctb, line 478: was the bug here?

saidctb (approved):

psydac/api/ast/fem.py
yguclu: Yes, it was only printing the instructions for the last field.

saidctb: I see

PR #115, opened 2021-04-27
Fix evaluation of non-uniform rational B-splines (NURBS)

saidctb:
- Improve `eval_field` to compute weighted fields
- Add tests for the evaluation of `NurbsMapping`

### DIFF
--- a/psydac/fem/basic.py
+++ b/psydac/fem/basic.py
@@ -82,7 +82,7 @@ class FemSpace( metaclass=ABCMeta ):
             Evaluation point(s) in logical domain.
 
         weights : StencilVector, optional
-            Weights of the basis functions, they belong to the same space of the field space coefficients.
+            Weights of the basis functions, such that weights.space == field.coeffs.space.
 
         Returns
         -------
@@ -105,7 +105,7 @@ class FemSpace( metaclass=ABCMeta ):
             Evaluation point(s) in logical domain.
 
         weights : StencilVector, optional
-            Weights of the basis functions, they belong to the same space of the field space coefficients.
+            Weights of the basis functions, such that weights.space == field.coeffs.space.
 
         Returns
         -------
--- a/psydac/fem/basic.py
+++ b/psydac/fem/basic.py
@@ -242,7 +242,7 @@ class FemField:
 
     # ...
     def gradient( self, *eta , weights=None):
-        """Evaluate gradient of field at location identified by logical coordinates eta."""
+        """Evaluate gradient of weighted field at location identified by logical coordinates eta."""
 
         if weights is not None:
             raise NotImplementedError('gradient evaluation is not available for a weighted field')
@@ -251,11 +251,7 @@ class FemField:
         
     # ...
     def divergence(self, *eta, weights=None):
-        """Evaluate divergence of vector field at location identified by logical coordinates eta."""
-
-        if weights is not None:
-            raise NotImplementedError('gradient evaluation is not available for a weighted field')
-
+        """Evaluate divergence of weighted vector field at location identified by logical coordinates eta."""
         return self._space.eval_field_divergence(self, *eta, weights=weights)
 
     # ...
--- a/psydac/fem/basic.py
+++ b/psydac/fem/basic.py
@@ -81,6 +81,9 @@ class FemSpace( metaclass=ABCMeta ):
         eta : list of float or numpy.ndarray
             Evaluation point(s) in logical domain.
 
+        weights : StencilVector, optional
+            Weights of the basis functions, they belong to the same space of the field space coefficients.
+
         Returns
         -------
         value : float or numpy.ndarray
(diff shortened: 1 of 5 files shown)


psydac/fem/basic.py
```diff
@@ -69,7 +69,7 @@ def is_product( self ):
     # Abstract interface: evaluation methods
     #---------------------------------------
     @abstractmethod
-    def eval_field( self, field, *eta ):
+    def eval_field( self, field, *eta , weights=None):
         """
         Evaluate field at location(s) eta.
```
yguclu, line 74: Please update docstring. It should explain that `weights.space == field.space`

psydac/fem/basic.py
```diff
@@ -89,7 +89,7 @@ def eval_field( self, field, *eta ):
         """
 
     @abstractmethod
-    def eval_field_gradient( self, field, *eta ):
+    def eval_field_gradient( self, field, *eta , weights=None):
         """
         Evaluate field gradient at location(s) eta.
```
yguclu, line 94: Same comment as above

psydac/fem/basic.py
```diff
@@ -230,19 +230,19 @@ def __getitem__(self, key):
         return self._fields[key]
 
     # ...
-    def __call__( self, *eta ):
+    def __call__( self, *eta , weights=None):
         """Evaluate field at location identified by logical coordinates eta."""
-        return self._space.eval_field( self, *eta )
+        return self._space.eval_field( self, *eta , weights=weights)
```
yguclu, line 235: Same comment as above

psydac/fem/basic.py
```diff
@@ -230,19 +230,19 @@ def __getitem__(self, key):
...
 
     # ...
-    def gradient( self, *eta ):
+    def gradient( self, *eta , weights=None):
         """Evaluate gradient of field at location identified by logical coordinates eta."""
-        return self._space.eval_field_gradient( self, *eta )
+        return self._space.eval_field_gradient( self, *eta , weights=weights)
```
yguclu, line 240: Same comment as above

psydac/fem/basic.py
```diff
@@ -230,19 +230,19 @@ def __getitem__(self, key):
...
         
     # ...
-    def divergence(self, *eta):
+    def divergence(self, *eta, weights=None):
         """Evaluate divergence of vector field at location identified by logical coordinates eta."""
-        return self._space.eval_field_divergence(self, *eta)
+        return self._space.eval_field_divergence(self, *eta, weights=weights)
```
yguclu, line 245: Same comment as above

psydac/fem/splines.py
```diff
@@ -239,10 +239,15 @@ def eval_field( self, field, *eta ):
         if self.basis == 'M':
             basis *= self._scaling_array[index]
 
-        return np.dot( field.coeffs[index], basis )
+        coeffs = field.coeffs[index].copy()
+
+        if weights:
+            coeffs *= wieghts[index]
```
yguclu, line 245: Typo: wieghts -> weights

I suspect that this code is never tested. Can you add a 1D unit test then?

yguclu (changes requested):
Excellent fix! I have a couple of comments

psydac/fem/basic.py
saidctb: okay

psydac/fem/basic.py
saidctb: done

psydac/fem/splines.py
saidctb: yes this part is never tested, how do we add a 1d test using `NurbsMapping`

psydac/fem/basic.py
yguclu: @saidctb Still missing

psydac/fem/basic.py
saidctb: done

psydac/fem/basic.py
saidctb: this case is not available yet, so I added a `NotImplementedError`

psydac/fem/basic.py
saidctb: done

psydac/fem/basic.py
yguclu: This is the base class, it is not its job to know whether the method exists or not. `NotImplementedError` is already raised by the subclasses, I believe.

psydac/fem/splines.py
yguclu: It should not be much different than a 2D test... You can create a 1D NURBS using Igakit

psydac/fem/basic.py
yguclu: This is the base class, it is not its job to know whether the method exists or not. NotImplementedError is already raised by the subclasses, I believe.

psydac/fem/basic.py
saidctb: ah okay

psydac/fem/splines.py
saidctb: we always return a TensorFemSpace when we use the `HDF5` files, so we will need to create everything by hand

psydac/fem/basic.py
saidctb: done

yguclu (approved):

psydac/fem/splines.py
yguclu: You're right, it does not really matter then.

PR #116, opened 2021-04-27
Accelerate matrix-vector multiplication with backends

saidctb:
To accelerate the dot product, we first generate pure Python code specific to the dimensions and the format of the  `StencilMatrix` object,  after that, we use one of the backends to accelerate the Python code.
This is an example of the generated code for a StencilMatrix with dimensions = `(16, 16, 7,  7)`.
```python
from pyccel.decorators import types

@types("real[:,:,:,:]", "real[:,:]", "real[:,:]")
def dot_product(mat, x, out):

    from itertools import product
    for i1,i2 in product(range(0, 16, 1),range(0, 16, 1)):
        v = 0.0
        for k1,k2 in product(range(0, 7, 1),range(0, 7, 1)):
            v += mat[3 + i1,3 + i2,k1,k2]*x[i1 + k1,i2 + k2]
        
        out[3 + i1,3 + i2] = v
```
however, in the parallel case, we pass the number of rows as an argument, where processes can have a different number of rows.
The previous example executed in parallel will generate the following Python code:

```python
from pyccel.decorators import types

@types("real[:,:,:,:]", "real[:,:]", "real[:,:]", int, int )
def dot_product(mat, x, out, n1, n2):

    from itertools import product
    for i1,i2 in product(range(0, n1, 1),range(0, n2, 1)):
        v = 0.0
        for k1,k2 in product(range(0, 7, 1),range(0, 7, 1)):
            v += mat[3 + i1,3 + i2,k1,k2]*x[i1 + k1,i2 + k2]
        
        out[3 + i1,3 + i2] = v
```

### DIFF
--- a/psydac/api/ast/linalg.py
+++ b/psydac/api/ast/linalg.py
@@ -1,10 +1,18 @@
 # coding: utf-8
 
+import sys
+import os
+import importlib
+
 from sympy import Mul, Tuple
-from sympy import Mod, Abs, Range
+from sympy import Mod, Abs, Range, Symbol
+from sympy import Function
 
 from pyccel.ast.core import Variable, IndexedVariable
 from pyccel.ast.core import For
+from pyccel.ast.core import Slice, String
+
+from pyccel.ast.datatypes import NativeInteger
 from pyccel.ast.core import Assign
 from pyccel.ast.core import AugAssign
 from pyccel.ast.core import Product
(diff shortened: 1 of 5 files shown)


saidctb:
@yguclu  can you test the file `test_stencil_matrix.py` in `linalg/tests` ?, I can't reproduce the error on my computer

yguclu:
> @yguclu can you test the file `test_stencil_matrix.py` in `linalg/tests` ?, I can't reproduce the error on my computer.
> you just need to install `fastcache` with pip

Yes I will check

yguclu (changes requested):
`fastcache` is a pretty old library that provides a "C implementation of Python 3 `functools.lru_cache` for Python 2.6, 2.7, 3.2, 3.3, 3.4". According to the project's [webpage](https://github.com/pbrady/fastcache):

> **Performance Warning**
> 
> As of Python 3.5, the CPython interpreter implements `functools.lru_cache` in C. It is generally faster than this library due to its use of a more performant internal API for dictionaries (and perhaps other reasons). Therefore this library is only recommended for Python 2.6-3.4

Therefore we should not use `fastcache`, but rather `functools.lru_cache` which is part of the Python 3 standard library. As an additional benefit, this will avoid the current issue with the broken wheel installation.

yguclu (approved):

PR #126, opened 2021-05-17
Fix `__div__` -> `__truediv__`

davschneller:
This PR updates the division operator to Python 3 (assuming that Python 2 support is obsolete/removed already),
i.e. it enables us to do e.g. `M /= a` where `M` is a `Matrix` or `Vector` object, and `a` is something that we can invert and have `M *= a` implemented, e.g. `a` could be a scalar. Note that we do not implement `__floordiv__` here (if that should be needed at all).

* `__div__` -> `__truediv__`, `__idiv__` -> `__itruediv__`
* Adds simple tests for `StencilMatrix` and `StencilVector` concerning some of their operators.

### DIFF
--- a/psydac/linalg/basic.py
+++ b/psydac/linalg/basic.py
@@ -95,10 +95,10 @@ class Vector( metaclass=ABCMeta ):
     #-------------------------------------
     # Methods with default implementation
     #-------------------------------------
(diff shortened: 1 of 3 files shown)


yguclu (approved):
Thanks for catching and fixing this bug!

PR #127, opened 2021-05-17
Add `StencilMatrix.dot` backend propagation

davschneller:
This PR lets the backend for `StencilMatrix.dot` propagate when multiplying, adding or transposing that said `StencilMatrix`.

* Propagates the `backend` parameter to all stencil matrices which are generated by `__mul__`, `__add__`, `copy`, `transpose`, etc. (in the case of `__add__` or `__sub__`, a warning is thrown if the backend of the left-hand side and right-hand side matrices differ; then the backend of the left-hand side is chosen by default)
* Makes it possible to use `PSYDAC_BACKEND_PYTHON` as backend
* `set_backend` can now be called again with another backend, the original args are stored
* Includes tests

### DIFF
--- a/psydac/linalg/stencil.py
+++ b/psydac/linalg/stencil.py
@@ -601,15 +601,15 @@ class StencilMatrix( Matrix ):
         args['gpads']        = tuple(V.pads)
         args['pads']         = tuple(self._pads)
 
-        self._args = args
+        self._dotargs_null = args
+        self._args = args.copy()
 
         self._func = self._dot
 
         if backend is None:
             backend = PSYDAC_BACKENDS.get(os.environ.get('PSYDAC_BACKEND'))
 
-        if backend:
-            self.set_backend(backend)
+        self.set_backend(backend)
 
     #--------------------------------------
     # Abstract interface
@@ -711,6 +711,10 @@ class StencilMatrix( Matrix ):
     @property
     def pads( self ):
         return self._pads
+    
+    @property
+    def backend(self):
+        return self._backend
 
     # ...
     def __getitem__(self, key):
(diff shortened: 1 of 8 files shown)


psydac/linalg/stencil.py
```diff
@@ -763,7 +767,7 @@ def __add__(self, m):
         assert m._domain   is self._domain
         assert m._codomain is self._codomain
         assert m._pads     == self._pads
-        w = StencilMatrix(self._domain, self._codomain, self._pads)
+        w = StencilMatrix(self._domain, self._codomain, self._pads, self._backend)
```
yguclu, line 770: Can you please check if the two matrices use the same backend, and if they don't raise a clear warning like the following?

```python
import warnings

if m._backend != self._backend:
    msg = 'Adding two matrices with different backends is ambiguous - defaulting to backend of first addend'
    warnings.warn(msg, category=RuntimeWarning)
```

psydac/linalg/stencil.py
```diff
@@ -776,7 +780,7 @@ def __sub__(self, m):
...
```
yguclu, line 783: Same comment as above

yguclu (commented):
Good job @davschneller!

I would like a runtime warning to be raised if we have an ambiguity in the result's backend.

yguclu (commented):
Thanks for the changes. Please update the PR's description...

davschneller:
Yes, it is now updated

yguclu (approved):
Waiting for [github actions problem](https://www.githubstatus.com/incidents/m16jzl31gnqt) to be fixed

PR #128, opened 2021-05-17
Zero borders for cartesian distribution in non-periodic cases

davschneller:
This PR adds some tiny enhancements to the cartesian distribution:

* Datatypes for non-used MPI send and receive operations are not created; and data at such boundaries is zeroed.
* Suppressed `print` statements in some testcases (and output is only done if `verbose` is true).
* Adds a new test for the `StencilMatrix` class, when reducing in some directions.


saidctb:
thanks @davschneller  for this PR, I'm currently working on the cart decomposition , so we might have some conflicts , it's maybe better to wait untill we discuss about this to decide what to do.

davschneller:
Sure; this issue can wait (it has a rather low priority right now anyways).

PR #129, opened 2021-05-31
Update PCG method to support `LinearSolver` and callable functions as input argument

davschneller:
This PR slightly modifies the `psydac.linalg.iterative_solvers.pcg`, so that we may use a `LinearSolver` object or a callable function instead of a string, or simply input `None` as preconditioner. (note that as of now, this PR modifies as little as necessary)

* If we input no preconditioner (i.e. `pc is None`), the regular `cg` method is called.
* If a `LinearSolver` is used, it utilizes the `out` parameter of the `solve` method, i.e. we do not allocate a new vector for each round.
* If a callable function is used, the parameter format is as for the functions which are addressable as strings. I.e. we have `function(A, r)`, where we want to solve `As = r`. `s` is always re-allocated.
* The two string parameters `'jacobi', 'weighted_jacobi'` are kept for compatibility reasons with the existing code for now.
* Includes an interface test.

### DIFF
--- a/psydac/linalg/iterative_solvers.py
+++ b/psydac/linalg/iterative_solvers.py
@@ -4,6 +4,7 @@ This module provides iterative solvers and precondionners.
 
 """
 from math import sqrt
+from psydac.linalg.basic import LinearSolver
 
 __all__ = ['cg', 'pcg', 'bicg', 'jacobi', 'weighted_jacobi']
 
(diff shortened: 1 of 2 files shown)


davschneller:
I also made a branch `devel-cginterface` which changes a bit more of the code, i.e. it merges the `cg` and `pcg` methods, and adds more type checks (and also, it adds an out parameter). If wanted, I can add the changes done there to this PR.

psydac/linalg/iterative_solvers.py
```diff
@@ -164,14 +165,24 @@ def pcg(A, b, pc, x0=None, tol=1e-6, maxiter=1000, verbose=False):
...
+        # for now, call the cg method here
+        return cg(A, b, x0=x0, tol=tol, maxiter=maxiter, verbose=verbose)
+    elif isinstance(pc, str):
+        pcfun = globals()[pc]
+        psolve = lambda r: pcfun(A, r)
+    elif isinstance(pc, LinearSolver):
+        s = b.space.zeros()
+        psolve = lambda r: pc.solve(r, out=s)
+    elif hasattr(pc, '__call__'):
+        psolve = lambda r: pc(A, r)
```
yguclu, line 178: Please update the function's docstring after these changes.

yguclu (changes requested):
This PR looks very good to me, and it can be merged once you update the docstring of function `pcg`.

yguclu:
> I also made a branch `devel-cginterface` which changes a bit more of the code, i.e. it merges the `cg` and `pcg` methods, and adds more type checks (and also, it adds an out parameter). If wanted, I can add the changes done there to this PR.

@davschneller If you want to create an additional PR for branch `devel-cginterface`, I will be happy to have a look at it after this PR is merged

davschneller:
Yes, thanks; I'll see if I'll make it

yguclu (approved):
I like these changes very much! in particular the fact that you avoid creating a temporary array at each call of the preconditioner.

PR #130, opened 2021-05-31
Add basic support for distributed multi-dimensional FFTs

davschneller:
This PR adds some basic multi-dimensional FFT methods which work in a distributed environment (i.e. it enables us to compute `fftn`, `ifftn`, `dctn`, etc. in a distributed/parallel setting for `StencilVector`s).

* Implements FFT, DCT, DST, and all their inverse methods (for compatibility).
* Internally, it uses a `KroneckerLinearSolver` to do the FFT/DCT/DST in each direction. Currently, it uses `scipy.fft` without additional padding for the one-dimensional FFTs; the normalization and worker parameters are forwarded to the methods. (future optimization, like padding, FFTW support, slab distribution or support for `mpi4py_fft` etc. are subject of future PRs)
* To support all this, we also propagated the `StencilVectorSpace.dtype` to the datatype of the temporary arrays.
* Includes basic tests.

### DIFF
--- a/psydac/linalg/fft.py
+++ b/psydac/linalg/fft.py
@@ -44,10 +44,10 @@ class DistributedFFTBase(LinearOperator):
         
         def solve(self, rhs, out=None, transposed=False):
             if out is None:
-                out = np.empty_like(v)
+                out = np.empty_like(rhs)
             
-            if out is not v:
-                out[:] = v
+            if out is not rhs:
+                out[:] = rhs
             
             self._function(out)
 
--- a/psydac/linalg/fft.py
+++ b/psydac/linalg/fft.py
@@ -142,10 +142,13 @@ class DistributedDCT(DistributedFFTBase):
     workers : Union[int, NoneType]
         Specifies the number of worker threads. By default set to the number of OpenMP threads, if given.
         See also the documentation of the corresponding scipy.fft.dct parameter.
+    
+    ttype : int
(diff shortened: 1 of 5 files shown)


davschneller:
Suggestions for other names (or removal of potentially redundant operations) are always welcome.

psydac/linalg/fft.py
```diff
@@ -0,0 +1,226 @@
...
+                out[:] = rhs
+            
+            self._function(out)
+
+            return out
+
+    def __init__(self, space, function):
+        assert isinstance(space, StencilVectorSpace)
+        onedimsolver = DistributedFFTBase.OneDimSolver(function)
+        self._isolver = KroneckerLinearSolver(space, [onedimsolver] * space.ndim)
```
yguclu, line 59: Here you are using the same 1D solver (`function`) along every direction. This will not work in the future if we use an FFT library where the plan can be precomputed, in which case this will be different for each direction. Given that this is the base class in this new module, we may want to be more careful.

yguclu (changes requested):
Good job! This is very user friendly. I have one remark.

yguclu (approved):

psydac/linalg/fft.py
yguclu: I agree. Thanks for this modification, that's sufficient for now!

PR #131, opened 2021-06-05
Propagate the `quad_order` parameter in the space discretization

davschneller:
This PR fixes a bug which could cause segmentation faults / out of bounds errors (depending on the backend used) when using the `quad_order` parameter with the `discretize` method.

* Discretization to a `TensorFEMSpace` now properly processes the `quad_order` parameter.
* (Bi)linear forms etc. need to have the same `quad_order` for now, as they use the underlying `TensorFEMSpace` quadrature grids. (allowing deviations here, e.g. by generating new quadrature grids on-demand may be done in a future PR) If we inserted a larger `quad_order` than there would be given to the space, we would get an out of bounds error (or segfault when compiling with pyccel; or just incorrect results), and if we chose a too small `quad_order` here, we would only use some quadrature points from the larger quadrature grid.
* Includes tests.

### DIFF
--- a/psydac/fem/tensor.py
+++ b/psydac/fem/tensor.py
@@ -36,7 +36,8 @@ class TensorFemSpace( FemSpace ):
         self._spaces = tuple(args)
 
         npts    = [V.nbasis   for V in self.spaces]
-        pads    = [V.degree   for V in self.spaces]
+        pads    = [V._pads    for V in self.spaces]
+        degree  = [V.degree   for V in self.spaces]
         periods = [V.periodic for V in self.spaces]
         basis   = [V.basis    for V in self.spaces]
 
@@ -71,9 +72,11 @@ class TensorFemSpace( FemSpace ):
         # Shortcut
         v = self._vector_space
 
+        self._quad_order = kwargs.pop('quad_order', [sp.degree for sp in self.spaces])
+
         # Compute extended 1D quadrature grids (local to process) along each direction
-        self._quad_grids = tuple( FemAssemblyGrid( V,s,e, nderiv=V.degree)
-                                  for V,s,e in zip( self.spaces, v.starts, v.ends ) )
+        self._quad_grids = tuple( FemAssemblyGrid( V,s,e, nderiv=V.degree, pad=p, quad_order=q)
+                                  for V,s,e,p,q in zip( self.spaces, v.starts, v.ends, v.pads, self._quad_order ) )
 
         # Determine portion of logical domain local to process
         self._element_starts = tuple( g.indices[g.local_element_start] for g in self.quad_grids )
@@ -344,6 +347,10 @@ class TensorFemSpace( FemSpace ):
     @property
     def spaces( self ):
         return self._spaces
+    
+    @property
+    def quad_order( self ):
+        return self._quad_order
 
     @property
     def quad_grids( self ):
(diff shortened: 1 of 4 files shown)


saidctb (approved):

psydac/fem/tensor.py
```diff
@@ -611,31 +620,11 @@ def reduce_degree(self, axes, basis='B'):
         periods = [V.periodic for V in spaces]
         # create new Tensor Vector
         if v.cart:
-
-            tensor_vec = TensorFemSpace(*spaces, comm=v.cart.comm)
             red_cart = v.cart.remove_last_element(axes)
-            v = StencilVectorSpace(red_cart)
-            tensor_vec._vector_space = v
```
yguclu, line 618: Could you explain these changes?

psydac/fem/tensor.py
```diff
@@ -611,31 +620,11 @@ def reduce_degree(self, axes, basis='B'):
...
-       # Compute extended 1D quadrature grids (local to process) along each direction
-       
-        tensor_vec._quad_grids = tuple( FemAssemblyGrid( V,s,e,quad_order=q, pad=q )
-                                  for V,s,e,q in zip( spaces, v.starts, v.ends, self.degree ) )
-
-        tensor_vec._element_starts, tensor_vec._element_ends =  self._element_starts, self._element_ends
-
-        # Compute limits of eta_0, eta_1, eta_2, etc... in subdomain local to process
-        tensor_vec._eta_limits = tuple( (space.breaks[s], space.breaks[e+1])
-        for s,e,space in zip( self._element_starts, self._element_ends, spaces ) )
```
yguclu, line 636: Is this code obsolete?

yguclu (changes requested):

psydac/fem/tensor.py
davschneller: I think it might be by now (the code is very similar to the constructor, and hence I thought it would be a good idea to unify as much as possible there; so that future changes are easier doable):

* by initializing the new `TensorFemSpace` with the new `cart`, we only really change the `self.global_element_starts`, `self.global_element_ends` parameters (which are currently not used anywhere, even not in some branches I checked).
* the `_quad_grids` assembly should be the same as when we initialize with the new cart anyways here (we have pads now, after all)
* The `_element_starts` and `_element_ends` will still do their job as described (no double computation), as long as the `FemAssemblyGrid` works by distributing each index (from the `indices` array) onto a unique process (they may change though by reducing the degree). Cf. also the test included in #132 which checks exactly this. Besides, these parameters are currently only used for plotting, and computing the `_eta_limits`, as it seems.
* The `_eta_limits` then are constructed in the same way as in the constructor, i.e. if the previous statements are true, then the `_eta_limits` should be correct as well. Also, it is currently not used outside the `TensorFemSpace` class, as far as I checked.

Of course, if I may have forgotten something (and some of these changes were incorrect), I can revert some of these changes; like the `_element_starts` and `_element_ends`, if necessary.

psydac/fem/tensor.py
davschneller: See the comment below

psydac/fem/tensor.py
yguclu: Thanks, I've just double checked and I think that you are right.

yguclu (approved):

PR #132, opened 2021-06-05
Fix distributed/parallel spline assembly when using DeRham complexes

davschneller:
This PR fixes a bug in relation to the `FemAssemblyGrid` when `pad` does not equal `degree`. Previously, there seemed to be some discrepancies on multiple processes (except for the first one) at least on a periodic grid which caused a segmentation fault or an out of bounds error in the kernel, depending on if pyccel was used or not. Now, the spans are elevated on all processes, if `pad==degree+1`. (i.e. the padding parameter is used instead of the degree parameter) Henceforth, this PR does:

* Make DeRham work in parallel;
* Take the pad parameter directly from the `SplineSpace`;
* (sort of) fix `local_element_start` and `local_element_end` in the corner cases, where `end < degree`;
* Include small tests;
* Add parallel unit tests to files `test_api_feec_1d.py` and `test_api_feec_2d.py`, with check on L2 error norm.

### DIFF
--- a/psydac/api/tests/test_api_feec_1d.py
+++ b/psydac/api/tests/test_api_feec_1d.py
@@ -1,6 +1,8 @@
 # coding: utf-8
 # Copyright 2020 Yaman Güçlü
 
+import pytest
+
 #==============================================================================
 # TIME STEPPING METHOD
 #==============================================================================
@@ -70,7 +72,9 @@ def run_maxwell_1d(*, L, eps, ncells, degree, periodic, Cp, nsteps, tend,
     from sympde.expr     import BilinearForm
 
     from psydac.api.discretization import discretize
-    from psydac.api.settings       import PSYDAC_BACKEND_GPYCCEL
+
+    # for now, switch to the Python backend, to be able to detect out of bounds errors
+    from psydac.api.settings       import PSYDAC_BACKEND_PYTHON
     from psydac.feec.pull_push     import push_1d_l2
 
     #--------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


davschneller:
Ok, I tried it. But I had to disable the correctness test (the assertions) of E and B for now; as the `eval_field` did not seem to function that properly (maybe #131 would help here; I'd have to see). And, as expected; the mass matrix assembly really failed on `devel` due to an out of bounds error. On this branch, it runs through; but again, the correctness checks may have to wait a bit...

Commit will come shortly

yguclu:
> Ok, I tried it (by simply parallelizing the existing unit tests). But I had to disable the correctness test (the assertions) of E and B for now; as the `eval_field` did not seem to function that properly (maybe #131 would help here; I'd have to see). And, as expected; the mass matrix assembly really failed on `devel` due to an out of bounds error. On this branch, it runs through; but again, the correctness checks may have to wait a bit...
> 
> Commit will come shortly

I don't think that we have a parallel implementation for `eval_field`, so we cannot easily check the max-norm of the error. But it should be possible to assemble the L2-norm of the error instead.

davschneller:
Ok, I tried to make the `eval_field` work with MPI as well (locally on this branch; but it can be easliy moved to another one; also it might require separate tests). With it, the 2D periodic and dirichlet test and the 1D periodic and penalized dirichlet now pass also in parallel.
The 1D dirichlet strong case probably fails due to other reasons (maybe since it takes a differentiation matrix, transposes it, and turns it into a StencilMatrix; cf. the comment to #134).

yguclu (approved):

PR #133, opened 2021-06-07
Add Null operators, enhance Identity operators

davschneller:
This tiny PR adds operators which only return zero. Also, we update the `IdentityLinearOperator` classes a bit.

* `IdentityLinearOperator` got a little bit of documentation, and also supports now non-zero `out` types. Same goes for `IdentityMatrix` and `IdentityStencilMatrix`. The latter now also supports multiple dimensions.
* Adds `NullLinearOperator`, `NullMatrix`, `NullStencilMatrix`. All of them can have different domain and codomain spaces (the `NullStencilMatrix` needs their padding to be the same still), and always output zero. They also support the `out` vector.
* Includes some basic tests.

### DIFF
--- a/psydac/linalg/basic.py
+++ b/psydac/linalg/basic.py
@@ -5,7 +5,7 @@
 from abc   import ABCMeta, abstractmethod
 from numpy import ndarray
 
-__all__ = ['VectorSpace', 'Vector', 'LinearOperator', 'LinearSolver', 'Matrix', 'IdentityElement', 'NullElement']
+__all__ = ['VectorSpace', 'Vector', 'LinearOperator', 'LinearSolver', 'Matrix']
 
 #===============================================================================
 class VectorSpace( metaclass=ABCMeta ):
--- a/psydac/linalg/basic.py
+++ b/psydac/linalg/basic.py
@@ -8,20 +8,6 @@ from numpy import ndarray
 __all__ = ['VectorSpace', 'Vector', 'LinearOperator', 'LinearSolver', 'Matrix', 'IdentityElement', 'NullElement']
 
 #===============================================================================
-class IdentityElement(metaclass=ABCMeta):
-    """
-    Dummy type for a general identity element.
-    """
-    pass
-
-#===============================================================================
-class NullElement(metaclass=ABCMeta):
-    """
-    Dummy type for a general null element.
-    """
-    pass
-
-#===============================================================================
 class VectorSpace( metaclass=ABCMeta ):
     """
     Generic vector space V.
--- a/psydac/linalg/basic.py
+++ b/psydac/linalg/basic.py
@@ -5,7 +5,21 @@
 from abc   import ABCMeta, abstractmethod
 from numpy import ndarray
 
-__all__ = ['VectorSpace', 'Vector', 'LinearOperator', 'LinearSolver', 'Matrix']
+__all__ = ['VectorSpace', 'Vector', 'LinearOperator', 'LinearSolver', 'Matrix', 'IdentityElement', 'NullElement']
+
+#===============================================================================
+class IdentityElement(metaclass=ABCMeta):
+    """
+    Dummy type for a general identity element.
+    """
+    pass
+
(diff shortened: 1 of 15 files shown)


psydac/linalg/basic.py
```diff
@@ -5,7 +5,21 @@
...
+    Dummy type for a general identity element.
+    """
+    pass
+
+#===============================================================================
+class NullElement(metaclass=ABCMeta):
+    """
+    Dummy type for a general null element.
+    """
+    pass
```
yguclu, line 22: The naming of these classes does not convince me, as I do not understand what "element" means here, neither in the name of the classes nor in the docstrings.

More importantly these classes are not used anywhere in Psydac, so they seem completely unnecessary to me. I recommend that we postpone their introduction to a future PR dedicated to the abstract linear operators.

yguclu (changes requested):

psydac/linalg/basic.py
davschneller: Ok, they're gone for now.

psydac/linalg/identity.py
```diff
@@ -26,9 +26,32 @@ def codomain( self ):
...
+        """
         assert isinstance( v, Vector )
         assert v.space is self.domain
-        # find a way to handle out not None
+
+        if out is not None and out is not v:
+            assert isinstance(out, Vector)
+            assert v.space is out.space
+            out *= 0.0
+            out += v
```
yguclu, line 52: Is this the place where you would like to use a `copyto` method? Since the `Vector` base class already has a `copy` method, maybe we could just add the optional argument `out` to it. What do you think?

psydac/linalg/identity.py
```diff
@@ -37,28 +60,83 @@ class IdentityMatrix( Matrix, IdentityLinearOperator ):
...
-        n = V.npts[0]
-        p = p or V.pads[0]
+    def __init__(self, V, pads=None):
+        assert pads is None or len(pads) == V.ndim
 
-        super().__init__(V, V, pads=(p,))
+        super().__init__(V, V, pads=pads)
 
-        self._data[:, p] = 1.
+        idslice = tuple([slice(None) for _ in range(V.ndim)] + list(self.pads))
```
yguclu, line 107: Very minor comment: I think that this can be simply written as
```python
idslice = (slice(None),) * V.ndim + (*self.pads)
```
or
```python
idslice = (*((slice(None),) * V.ndim), *self.pads)
```
But if you want to avoid creating temporary lists or tuples, then using a generator is the way to go:
```python
idslice = (*(slice(None) for _ in range(V.ndim)), *self.pads)
```

On my computer the second option is the faster.

psydac/linalg/identity.py
davschneller: Yes, that sounds great! (and yes, this is the place where I would use it) Shall I add it to this PR already? Or to a separate one?

yguclu: Please just make a separate small PR. I'm going to merge this one as soon as you update its description!

psydac/linalg/identity.py
davschneller: Thanks for the hint!

yguclu (approved):

psydac/linalg/identity.py
```diff
@@ -37,28 +60,83 @@ class IdentityMatrix( Matrix, IdentityLinearOperator ):
...
-        n = V.npts[0]
-        p = p or V.pads[0]
+    def __init__(self, V, pads=None):
+        assert pads is None or len(pads) == V.ndim
 
-        super().__init__(V, V, pads=(p,))
+        super().__init__(V, V, pads=pads)
 
-        self._data[:, p] = 1.
+        idslice = tuple(*((slice(None),) * V.ndim), *self.pads)
```
yguclu, line 107: I think you should remove `tuple` from this line

yguclu (changes requested):

yguclu (approved):

PR #135, opened 2021-06-16
Add repeated knots

saidctb:


psydac/api/ast/fem.py
```diff
@@ -211,36 +278,41 @@ def __init__(self, expr, terminal_expr, spaces, tag=None, **kwargs):
             is_linear           = True
             tests               = expr.test_functions
             fields              = expr.fields
-            mapping             = spaces.symbolic_mapping
-            is_rational_mapping = spaces.is_rational_mapping
+            is_broken           = spaces.symbolic_space.is_broken
```
campospinto, line 281: What means 'broken' in this scope ?

saidctb: broken is used to denote if the space is broken

psydac/polar/c1_linops.py
```diff
@@ -89,7 +89,7 @@ def dot( self, v, out=None ):
         return out
 
     # ...
-    def toarray( self ):
+    def toarray( self, *, order='C' ):
```
yguclu, line 92: The `order` argument is not used by this function... Shouldn't you pass it to `np.zeros` on line 101?

psydac/polar/c1_linops.py
```diff
@@ -167,7 +167,7 @@ def __isub__(self, m):
     #-------------------------------------
     # Other properties/methods
     #-------------------------------------
-    def tocoo( self ):
+    def tocoo( self, *, order='C' ):
```
yguclu, line 170: As I said before, I don't understand what `order` means for a COO matrix...

psydac/polar/c1_linops.py
```diff
@@ -260,7 +260,7 @@ def dot( self, v, out=None ):
         return out
 
     # ...
-    def toarray( self ):
+    def toarray( self, *, order='C' ):
```
yguclu, line 263: The `order` argument is not used

psydac/polar/c1_linops.py
```diff
@@ -280,8 +280,8 @@ def toarray( self ):
         return a
 
     # ...
-    def tosparse( self ):
-        return self.tocoo()
+    def tosparse( self, *, order='C' ):
+        return self.tocoo(order=order)
```
yguclu, line 284: The `order` argument is not used, plus I cannot see its meaning for a COO matrix

psydac/polar/c1_linops.py
```diff
@@ -336,7 +336,7 @@ def __isub__(self, m):
     #-------------------------------------
     # Other properties/methods
     #-------------------------------------
-    def tocoo( self ):
+    def tocoo( self, *, order='C' ):
```
yguclu, line 339: Same comment

psydac/polar/dense.py
```diff
@@ -259,7 +259,7 @@ def __isub__( self, v ):
     #-------------------------------------
     # Other properties/methods
     #-------------------------------------
-    def toarray( self ):
+    def toarray( self, *, order='C' ):
```
yguclu, line 262: The `order` argument is not used

psydac/polar/dense.py
```diff
@@ -318,11 +318,11 @@ def dot( self, v, out=None ):
         return out
 
     # ...
-    def toarray( self ):
+    def toarray( self, *, order='C' ):
         return self._data.copy()
```
yguclu, line 322: The `order` argument is not used

psydac/polar/dense.py
```diff
@@ -318,11 +318,11 @@ def dot( self, v, out=None ):
...
 
     # ...
-    def tosparse( self ):
+    def tosparse( self, *, order='C' ):
```
yguclu, line 325: The `order` argument is not used, plus I cannot see its meaning for a COO matrix

yguclu (commented):
Good job Said, thanks for this massive PR! I am submitting a first partial review

psydac/api/essential_bc.py
```diff
@@ -94,12 +94,18 @@ def apply_essential_bc_BlockMatrix(a, bc):
...
             i = bc.position + i_loc
             js = [ij[1] for ij in keys if ij[0] == i]
             for j in js:
                 apply_essential_bc(a[i, j], bc)
+
+    elif bc.position is not None and not bc.variable.space.is_broken:
+        i = bc.position
+        js = [ij[1] for ij in keys if ij[0] == i]
+        for j in js:
+            apply_essential_bc(a[i, j], bc)
```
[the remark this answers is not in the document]
campospinto: Indeed a comment here would be useful, also because the meaning of `position` does not seem to be documented, neither here or in sympde

psydac/api/tests/navier_stockes.py
```diff
@@ -0,0 +1,280 @@
...
+
+    boundaries = {1: {'axis': 0, 'ext': -1},
+                  2: {'axis': 0, 'ext':  1},
+                  3: {'axis': 1, 'ext': -1},
+                  4: {'axis': 1, 'ext':  1}}
+
+    return tuple(boundaries[i] for i in args)
+#==============================================================================
+def run_navier_stokes_2d(domain, f, ue, pe, *, ncells, degree):
+
```
campospinto, line 57: You may specify here that this solves the steady-state problem

psydac/api/essential_bc.py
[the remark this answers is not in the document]
saidctb: the `EssentialBC` object has two attributes `position` and `index_component`  one refers to the position of the variable in the list of variables that defines the trial functions,  the other to the index of vector function component.
if we have the following trial functions  `(u,h,p)` the position of `h` is `1` and the index component of `h[0]` is `0`.

psydac/linalg/basic.py
```diff
@@ -144,11 +144,11 @@ class Matrix( LinearOperator ):
...
     #-------------------------------------
     @abstractmethod
-    def toarray( self ):
+    def toarray( self, *, order='C', with_pads=False ):
         """ Convert to Numpy 2D array. """
 
     @abstractmethod
-    def tosparse( self ):
+    def tosparse( self, *, order='C', with_pads=False ):
         """ Convert to any Scipy sparse matrix format. """
```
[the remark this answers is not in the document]
saidctb: I added those arguments because of Codacy , it complains if we change the signature

psydac/fem/basic.py
```diff
@@ -114,6 +119,42 @@ def eval_field_gradient( self, field, *eta , weights=None):
...
+    def __rmul__(self, a):
+        if a.is_product:
+            spaces = list(a.spaces)
+        else:
+            spaces = [a]
+
+        if self.is_product:
+            spaces += list(self.spaces)
+        else:
+            spaces += [self]
```
yguclu, line 150: Again, could we compress this code into two lines?

```
spaces = [*(   a.spaces if    a.is_product else    [a]),
          *(self.spaces if self.is_product else [self])]
```

psydac/linalg/block.py
```diff
@@ -500,15 +500,13 @@ class BlockMatrix( BlockLinearOperator, Matrix ):
...
+    def toarray( self, *, order='C', with_pads=False ):
+        """ Convert to Numpy 2D array. """
+        return self.tosparse(order=order).toarray()
 
     # ...
-    def tosparse( self ):
-        """ Convert to Scipy sparse matrix.
-        """
+    def tosparse( self, *, order='C', with_pads=False ):
+        """ Convert to any Scipy sparse matrix format. """
```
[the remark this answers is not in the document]
saidctb: yes the keyword argument is not used in all of them, I added them because of codacy, we may need to raise a not implemented error when they are used

psydac/polar/c1_linops.py
saidctb: the same comment as above, it's not implemented , I added them because of codacy , we can either implement them or raise a not implemented error

psydac/fem/basic.py
```diff
@@ -300,3 +341,5 @@ def __isub__(self, other):
         assert self._space is other._space
         self._coeffs -= other._coeffs
         return self
+
+from .vector import ProductFemSpace
```
yguclu, line 345: I guess that you're importing `ProductFemSpace` here at the end of the file in order to avoid issues with this cyclic dependency (vector -> basic -> vector). A better way to do this is to move this import to the methods that actually use `ProductFemSpace`, namely `__mul__` and `__rmul__`.

psydac/fem/basic.py
```diff
@@ -114,6 +119,42 @@ def eval_field_gradient( self, field, *eta , weights=None):
...
+        if self.is_product:
+            spaces += list(self.spaces)
+        else:
+            spaces += [self]
+
+        space = ProductFemSpace(*spaces)
+
+        if a.symbolic_space and self.symbolic_space:
+            space._symbolic_space =  a.symbolic_space * self.symbolic_space
+        return space
```
yguclu, line 156: Please note that methods `__mul__` and `__rmul__` are not `abstractmethod`s, and therefore they do not belong to the "abstract interface" section. So they should be moved to another section, which I would call "Concrete methods"

psydac/core/bsplines.py
```diff
@@ -482,7 +482,9 @@ def breakpoints( knots, degree ):
         Abscissas of all breakpoints.
 
     """
-    return np.unique( knots[degree:-degree] )
+    knots = np.array(knots)
+    diff  = np.append(True, np.diff(knots[degree:-degree])>1e-15)
+    return knots[degree:-degree][diff]
```
yguclu, line 487: I do not agree with this approach of hiding comparison tolerances inside the code. How can one know if `1e-15` is sufficiently small? If you want to do this you should provide the optional argument `tol` for the user to change.

In any case I don't think that the `breakpoints` function should take care of this. The original implementation was assuming that a knot sequence containing repeated knots is usually created by copying the same knot value multiple times, so that they will be identical no matter the precision. If you are worried about this you can add the following two things:

1. A dedicated function that "sanitizes" the knot sequence by applying a user-provided tolerance;
2. A clear comment in the docstring of `breakpoints`, which states that the input knot sequence must be properly sanitized.

Another approach, which I think is bullet-proof, is to use knot sequences without repeated knots in conjunction with an integer array containing the multiplicities of each knot.

psydac/core/bsplines.py
```diff
@@ -575,7 +578,7 @@ def elements_spans( knots, degree ):
 
     ie = 0
     for ik in range( degree, nk-degree ):
-        if knots[ik] != knots[ik+1]:
+        if knots[ik+1]-knots[ik]>=1e-15:
```
yguclu, line 581: See my previous comment to the `breakpoints` function: don't hide tolerances inside this function

psydac/core/bsplines.py
```diff
@@ -506,6 +508,7 @@ def greville( knots, degree, periodic ):
         Abscissas of all Greville points.
 
     """
+
```
yguclu, line 511: Useless white space change

psydac/core/bsplines.py
```diff
@@ -603,6 +606,9 @@ def make_knots( breaks, degree, periodic ):
     periodic : bool
         True if domain is periodic, False otherwise.
 
+    multiplicity: int
+        multiplicity of the knots in the knot sequence
```
yguclu, line 610: You should add "we assume that the same multiplicity applies to each interior knot"

Moreover, you should check that `1 <= multiplicity <= degree + 1`

psydac/core/bsplines.py
```diff
@@ -656,23 +665,35 @@ def elevate_knots(knots, degree, periodic):
     periodic : bool
         True if domain is periodic, False otherwise.
 
+    multiplicity : int
+        multiplicity of the knots.
```
yguclu, line 669: See my previous comment to the `make_knots` function

psydac/core/bsplines.py
```diff
@@ -656,23 +665,35 @@ def elevate_knots(knots, degree, periodic):
...
-        right = knots[-1]
+        left  = [knots[0],*knots[:degree+1]]
+        right = [knots[-1],*knots[-degree-1:]]
+
+        diff   = np.append(True, np.diff(knots[degree+1:-degree-1])>1e-15)
+        if len(knots[degree+1:-degree-1])>0:
+            unique = knots[degree+1:-degree-1][diff]
+            knots  = np.repeat(unique, multiplicity)
+        else:
+            knots = knots[degree+1:-degree-1]
```
yguclu, line 694: See my previous comment to the `breakpoints` function: don't hide tolerances inside this function

yguclu (changes requested):

psydac/linalg/basic.py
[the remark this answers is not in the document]
yguclu: Have you tried to add `**kwargs` to the methods of the base class? Does Codacy still complain?

psydac/api/essential_bc.py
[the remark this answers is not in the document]
yguclu: I see. So this is not a multi-patch domain, but rather a system of equations.

So why did you need to add this case? Was is not treated already? If so, was there a bug then?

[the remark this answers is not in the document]
saidctb: yes it was not treated , I had this problem in 1d where all the variables are scalars

psydac/linalg/basic.py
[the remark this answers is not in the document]
saidctb: I didn't try this but I will check

psydac/fem/basic.py
```diff
@@ -114,6 +119,42 @@ def eval_field_gradient( self, field, *eta , weights=None):
...
+
+        if self.is_product:
+            spaces = list(self.spaces)
+        else:
+            spaces = [self]
+
+        if a.is_product:
+            spaces += a.spaces
+        else:
+            spaces += [a]
```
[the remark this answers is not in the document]
saidctb: okay

psydac/linalg/stencil.py
```diff
@@ -570,14 +613,15 @@ def __init__( self, V, W, pads=None , backend=None):
         if pads is not None:
             for p,vp in zip(pads, V.pads):
                 assert p<=vp
-        
+
         self._pads     = pads or tuple(V.pads)
-        dims           = [e-s+2*p+1 for s,e,p in zip(W.starts, W.ends, W.pads)]
-        diags          = [2*p+1 for p in self._pads]
+        dims           = [e-s+2*mi*p+1 for s,e,p,mi in zip(W.starts, W.ends, W.pads, W.multiplicity)]
+        diags          = [compute_diag_len(p, md, mc) for p,md,mc in zip(self._pads, V.multiplicity, W.multiplicity)]
```
yguclu, line 619: As I recommended in a previous comment, here we can pass the multiplicities of domain and codomain as arguments to the constructor of `StencilMatrix` instead of storing them inside the `StencilVectorSpace` objects `V` and `M`.

psydac/ddm/cart.py
```diff
@@ -61,7 +61,7 @@ class CartDecomposition():
         (optional: default is MPI_COMM_WORLD).
 
     """
-    def __init__( self, npts, pads, periods, reorder, comm=MPI.COMM_WORLD, nprocs=None, reverse_axis=None ):
+    def __init__( self, npts, pads, periods, reorder, comm=MPI.COMM_WORLD, multiplicity=None, nprocs=None, reverse_axis=None ):
```
yguclu, line 64: I am really glad that you found a way to make Psydac work in parallel with knot sequences of non-unit multiplicity, but I think that it is excessive to have the `multiplicity` integer values appear also in this module, whose only purpose is to distribute an N-dimensional `StencilVector` across multiple processes. Do we really need to specialize `CartDecomposition` so much?

What are we going to do if we decide to support knot sequences where each knot can have a different multiplicity? Are we going to change every class in Psydac once again?

yguclu (changes requested):

psydac/ddm/cart.py
saidctb: if it was just one space, I would not have included this information in the cart, the problem is that we can have different spaces with different multiplicities , so we need to make sure that the decomposition makes sense.
when we use the method `reduce_elements`  to create the Derham sequence , we can reduce the space by just `1` element or `n` elements or possibly `m*n` such that `m` is the multiplicity and `n` the number of breakpoints.
we need to make a decomposition such that the break points owned by  each process unique and are identical between  between the spaces

saidctb: in the case of a general knot sequence, we will for sure need to change everything , maybe not as much as in this case, but we will need to update the generated assembly code, the linear algebra methods like (dot, transpose), maybe the data structure, and also the decomposition need to make sense so that the breakpoints are unique in each process

yguclu: Since we are distributing the B-spline coefficients, and hence the basis functions of a given spline space, we can think of assigning to each process a different knot sequence, which defines a different set of B-splines. By so doing `p + 1` knots are repeated between adjacent processes, and if the multiplicity of every interior knot is `1` there will be `p + 1` repeated breakpoints.

Given the above, I cannot see what you mean by "distributing the breakpoints" and making them "unique in each process". Do you mean that you want the breakpoints on a given process to remain unchanged after reducing the space degree?

@ratnania We may need a short discussion is needed in order to understand each other.

saidctb: yes that's what I mean

yguclu: OK.

When we assemble a mass (or stiffness) matrix by adding "element mass matrices", we are essentially decomposing an integral over the domain into a sum of integrals over each element. The requirement of having domain and codomain share the same breakpoints on a given process seems to simplify this process... is that the reason why you want that property?

psydac/cad/geometry.py
```diff
@@ -441,12 +442,22 @@ def refine_nurbs(nrb, ncells=None, degree=None):
     degree : <list>
         degree in each direction
 
+    multiplicity : <list>
+        multiplicity of each knot in the knot sequence in each direction
+
+    tol : <float>
+        tolerance
```
yguclu, line 449: Tolerance for what? Please explain...

psydac/cad/geometry.py
```diff
@@ -475,7 +486,87 @@ def refine_nurbs(nrb, ncells=None, degree=None):
...
+        nrb = nrb.refine(axis, knots)
     return nrb
+
+def refine_knots(knots, ncells, degree, multiplicity=None, tol=1e-9):
+    """
+    This function refines the knot sequence.
+    It contructs a new grid based on the new number of cells, and it adds the new break points to the nrb grid,
+    such that the total number of cells is equal to the new number of cells.
+    We use knot insertion to construct the new knot sequence , so the geometry is identical to the previous one.
+    It also elevates the degree of the nrb object based on the new degree.
```
yguclu, line 504: The `refine_knots` function looks very useful indeed!

But in order to avoid code duplication, I think that function `refine_nurbs` should call this function now.

yguclu (changes requested):

psydac/ddm/cart.py
saidctb: yes this is one of the reasons, but also for the dot product to work we need that , otherwise, we will need more ghost regions

yguclu (commented):
Why are all GLT unit tests marked as XFAIL now?

psydac/api/ast/nodes.py
```diff
@@ -42,6 +42,20 @@ class Zeros(Function):
...
+        else:
+            return Basic.__new__(cls, a1, a2)
+
+    @property
+    def a1(self):
+        return self._args[0]
+
+    @property
+    def a2(self):
+        return self._args[1]
```
yguclu, line 58: I don't think that `a1` and `a2` are expressive names for these properties. Doesn't `FloorDiv` inherit the `args` property from the `Function` class already?

psydac/api/ast/nodes.py
```diff
@@ -111,16 +159,25 @@ class IndexDofTrial(IndexDof):
...
-index_dof_trial = IndexDofTrial()
-index_deriv     = IndexDerivative()
+index_element        = IndexElement()
+index_quad           = IndexQuadrature()
+index_dof            = IndexDof()
+index_dof_test       = IndexDofTest()
+index_dof_trial      = IndexDofTrial()
+index_deriv          = IndexDerivative()
+index_outer_dof_test = IndexOuterDofTest()
+index_inner_dof_test = IndexInnerDofTest()
```
yguclu, line 179: Are these singletons? Could you explain why they are needed and how they are used?

psydac/api/ast/nodes.py
```diff
@@ -166,7 +223,7 @@ class EvalField(BaseNode):
        This function computes atomic expressions needed
        to evaluate  EvaluteField/VectorField final expression
     """
-    def __new__(cls, atoms, q_index, l_index, q_basis, l_basis, coeffs, l_coeffs, g_coeffs, tests, mapping, pads, nderiv, mask=None):
+    def __new__(cls, atoms, q_index, l_index, q_basis, l_basis, coeffs, l_coeffs, g_coeffs, tests, mapping, nderiv, mask=None):
```
yguclu, line 226: Since this is a very important class, I think it is a good time for adding a `Parameters` section to its docstring!

yguclu (changes requested):
It is quite difficult to navigate module `psydac.api.ast.nodes` without docstrings.

At least for those classes whose constructors have a long list of arguments, we should specify the type of each argument. Can you please do it?

psydac/api/grid.py
```diff
@@ -2,25 +2,71 @@
...
+                    p,w = get_points_weights(spaces, i, e)
+                    points[-1]  = np.concatenate((p, points[-1]))
+                    weights[-1] = np.concatenate((w, weights[-1]))
+                elif e>V1.quad_grids[i].indices[-1]:
+                    p,w = get_points_weights(spaces, i, e)
+                    points[-1]  = np.concatenate((points[-1],p))
+                    weights[-1] = np.concatenate((weights[-1],w))
+                else:
+                    raise ValueError("Could not contsruct indices")
+
```
yguclu, line 63: Are you constructing a quadrature grid that works for the product of any two spaces? If so, is this a bug-fix?

Can this construction handle our quadrature rule for histopolation? I think we had a problem as an histopolation interval could contain some knots.

psydac/api/grid.py
```diff
@@ -2,25 +2,71 @@
...
+        if e in s.quad_grids[axis].indices:
+            i = np.where(s.quad_grids[axis].indices==e)[0][0]
+            return s.quad_grids[axis].points[i:i+1], s.quad_grids[axis].weights[i:i+1]
 #==============================================================================
 class QuadratureGrid():
-    def __init__( self, V , axis=None, ext=None):
+    def __init__( self, V , axis=None, ext=None, trial_space=None):
+
+        n_elements          = []
+        indices             = []
```
yguclu, line 24: What is contained in the `indices` list?

psydac/api/tests/navier_stockes.py
```diff
@@ -0,0 +1,280 @@
+# -*- coding: UTF-8 -*-
```
yguclu, line 1: Please change the file name: "Navier Stokes" has no "c"

yguclu (changes requested):

psydac/core/bsplines.py
saidctb: when we have repeated knots, we have the choice to use `numpy.unique` to get the break points, or the way that I did it,I'm not sure if it's good to use numpy.unique with float numbers , that's why I used this method with a small enough tolerance.
so I suggest to add a keywrod argument that specifies the tolerance

psydac/linalg/basic.py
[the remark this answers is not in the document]
saidctb: done

psydac/linalg/block.py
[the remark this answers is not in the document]
saidctb: it's fixed

psydac/polar/c1_linops.py
saidctb: fixed

psydac/polar/c1_linops.py
saidctb: fixed

psydac/polar/c1_linops.py
saidctb: same as above

psydac/polar/c1_linops.py
saidctb: same as above

psydac/polar/c1_linops.py
saidctb: same as above

psydac/polar/dense.py
saidctb: it's been deleted

psydac/polar/dense.py
saidctb: it's been deleted

psydac/polar/dense.py
saidctb: it's been deleted

psydac/api/tests/navier_stockes.py
saidctb: done

psydac/fem/vector.py
```diff
@@ -39,6 +40,9 @@ def __init__( self, *spaces ):
         self._ncells = ncells[0]
         # ...
 
+        # ...
+        self._symbolic_space      = None
```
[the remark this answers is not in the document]
saidctb: done

psydac/fem/vector.py
```diff
@@ -195,6 +209,8 @@ def __init__( self, *spaces):
         self._vector_space = BlockVectorSpace(*v_spaces)
         # ...
 
+        self._symbolic_space      = None
```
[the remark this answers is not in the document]
saidctb: fixed

psydac/fem/basic.py
```diff
@@ -9,7 +9,6 @@
 from abc import ABCMeta, abstractmethod
 
 from psydac.linalg.basic import Vector
-
```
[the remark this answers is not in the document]
saidctb: fixed

psydac/fem/basic.py
```diff
@@ -56,6 +55,7 @@ def mapping( self ):
     def vector_space( self ):
         """Topologically associated vector space."""
 
+
```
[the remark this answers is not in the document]
saidctb: fixed

psydac/fem/basic.py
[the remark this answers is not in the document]
saidctb: done

psydac/fem/basic.py
saidctb: done

psydac/fem/basic.py
saidctb: done

psydac/fem/basic.py
saidctb: done

psydac/core/bsplines.py
saidctb: we had a conversation about this, we can either fix the tolerance to a number or add a kesword argument, that the user can change

psydac/core/bsplines.py
saidctb: fixed

psydac/core/bsplines.py
saidctb: done

psydac/core/bsplines.py
saidctb: done

psydac/core/bsplines.py
saidctb: done

psydac/linalg/utilities.py
```diff
@@ -1,15 +1,13 @@
...
 from psydac.linalg.stencil import StencilVectorSpace, StencilVector
 from psydac.linalg.block   import BlockVector, BlockVectorSpace
 
-import numpy as np
-
 __all__ = ['array_to_stencil']
 
 def array_to_stencil(x, Xh):
     """ converts a numpy array to StencilVector or BlockVector format"""
-
```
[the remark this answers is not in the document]
saidctb: done

psydac/linalg/stencil.py
```diff
@@ -51,18 +60,24 @@ def __init__( self, *args, **kwargs ):
             self._init_serial  ( *args, **kwargs )
 
     # ...
-    def _init_serial( self, npts, pads, periods, dtype=float ):
+    def _init_serial( self, npts, pads, periods, multiplicity=None, dtype=float ):
+
+        if multiplicity is None:
+            multiplicity = tuple(1 for i in pads)
```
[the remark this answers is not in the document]
saidctb: we need this information in the `StencilVectorSpace` because we need to set it for the cart in the paralle case, also the `StencilVector` and `StencilMatrix` use  this information.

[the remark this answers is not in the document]
saidctb: done for the naming

psydac/linalg/stencil.py
saidctb: we discussed this previously, the information is part of the space, we need that in the decomposition

psydac/ddm/partition.py
```diff
@@ -94,8 +94,8 @@ def mpi_compute_dims_uniform( mpi_size, npts ):
 
     nprocs = [1]*len( npts )
 
-    mpi_factors   = factorint( mpi_size )
-    npts_factors  = [factorint( n ) for n in npts]
+    mpi_factors   = factorint( int(mpi_size) )
+    npts_factors  = [factorint( int(n) ) for n in npts]
```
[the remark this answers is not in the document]
saidctb: I had a problem when the type is a numpy type instead of a the int of Python

psydac/fem/splines.py
```diff
@@ -53,7 +53,7 @@ class SplineSpace( FemSpace ):
         Set to "M" for M-splines (have unit integrals)
 
     """
-    def __init__( self, degree, knots=None, grid=None,
+    def __init__( self, degree, knots=None, grid=None, multiplicity=None, parent_multiplicity=None,
                   periodic=False, dirichlet=(False, False), basis='B', pads=None ):
```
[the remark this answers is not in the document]
saidctb: done

psydac/fem/splines.py
```diff
@@ -66,11 +66,23 @@ def __init__( self, degree, knots=None, grid=None,
             raise ValueError('Either knots or grid must be provided.')
 
         if knots is None:
-            knots = make_knots( grid, degree, periodic )
+            if multiplicity is None:
+                multiplicity = 1
```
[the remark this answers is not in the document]
saidctb: done

psydac/fem/splines.py
```diff
@@ -66,11 +66,23 @@ def __init__( self, degree, knots=None, grid=None,
...
 
+        indices = np.where(np.diff(knots[degree+1:-degree-1])>1e-15)[0]
+
+        if len(indices)>0:
+            multiplicity = np.diff(indices).max()
+        else:
+            multiplicity = max(1,len(knots[degree+1:-degree-1]))
+
+        if parent_multiplicity is None:
+            parent_multiplicity = 1
```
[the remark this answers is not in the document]
saidctb: yes you are right, we should check the consisentcy of the multiplicities, we should have `parent_multiplicity >= multiplicit`

psydac/fem/splines.py
```diff
@@ -102,7 +115,8 @@ def __init__( self, degree, knots=None, grid=None,
         self._ext_greville  = greville(elevate_knots(knots, degree, periodic), degree+1, periodic)
         self._scaling_array = scaling_array
 
-        self._histopolation_grid = unroll_edges(self.domain, self.ext_greville)
+        self._parent_multiplicity  = parent_multiplicity
+        self._histopolation_grid   = unroll_edges(self.domain, self.ext_greville)
```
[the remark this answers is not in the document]
saidctb: yes the change was there

psydac/fem/splines.py
```diff
@@ -229,6 +246,10 @@ def vector_space( self ):
     def is_product(self):
         return False
 
+    @property
+    def symbolic_space( self ):
+        return self._symbolic_space
```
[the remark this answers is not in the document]
saidctb: yes we add it in `discretize`, this is only needed in the api,  we do not needed in the conga method for example where we don't use the api, that's why we keep it optional.
I will add a setter property.

psydac/fem/grid.py
```diff
@@ -45,15 +45,17 @@ class FemAssemblyGrid:
         points (default: 1).
 
     """
-    def __init__( self, space, start, end, *, quad_order=None, nderiv=1 ):
+    def __init__( self, space, start, end, *, quad_order=None, nderiv=1, parent_start=None, parent_end=None):
```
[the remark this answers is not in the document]
saidctb: done

psydac/fem/grid.py
```diff
@@ -108,11 +115,16 @@ def __init__( self, space, start, end, *, quad_order=None, nderiv=1 ):
                     weights.append( glob_weights[k] )
                     indices.append( k )
                     ne += 1
-        
+
+
+        m = multiplicity if multiplicity>1 else 0
```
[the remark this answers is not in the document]
saidctb: this we need when the trial and the test spaces are diffrent and the parent multiplicity is `>1`, where we need more basis function than the usual

psydac/fem/grid.py
```diff
@@ -135,7 +146,7 @@ def __init__( self, space, start, end, *, quad_order=None, nderiv=1 ):
         self._quad_rule_x  = u
         self._quad_rule_w  = w
 
-        #-------------------------------------------
+        #---------------------- ---------------------
```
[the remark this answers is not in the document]
saidctb: yes

psydac/fem/grid.py
```diff
@@ -91,8 +93,13 @@ def __init__( self, space, start, end, *, quad_order=None, nderiv=1 ):
 
         if pad==degree:
             current_glob_spans  = glob_spans
+            current_start = start
+            current_end   = end
```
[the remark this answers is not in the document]
saidctb: done

psydac/fem/tensor.py
```diff
@@ -141,6 +147,10 @@ def vector_space(self):
     def is_product(self):
         return False
 
+    @property
+    def symbolic_space( self ):
+        return self._symbolic_space
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
```diff
@@ -184,11 +184,11 @@ def read( self, filename, comm=MPI.COMM_WORLD ):
...
                 if dtype == 'SplineMapping':
                     mapping = SplineMapping.from_control_points( tensor_space,
-                                                                 patch['points'] )
+                                                                 patch['points'][..., :pdim] )
 
                 elif dtype == 'NurbsMapping':
                     mapping = NurbsMapping.from_control_points_weights( tensor_space,
-                                                                        patch['points'],
+                                                                        patch['points'][..., :pdim],
                                                                         patch['weights'] )
```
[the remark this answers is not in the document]
saidctb: This is needed in case where we have 3d coordinates for a 2d geometry, where the last coordinate is fixed.

psydac/cad/geometry.py
saidctb: done

psydac/cad/geometry.py
```diff
@@ -475,7 +486,87 @@ def refine_nurbs(nrb, ncells=None, degree=None):
...
+        total number of cells in each direction
+
+    degree : <list>
+        degree in each direction
+
+    multiplicity : <list>
+        multiplicity of each knot in the knot sequence in each direction
+
+    tol : <float>
+        tolerance
```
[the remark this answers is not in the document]
saidctb: done

psydac/cad/geometry.py
saidctb: they are not identical, one takes optinal argument the other not, `refine_nurbs` will be used in `caid` in https://github.com/pyccel/caid/blob/5bc5428007c8642412762f80c36e8531e56cd15e/tests/test_cad_geometry.py#L8, that's why I kept them seperate.

saidctb:
> Why are all GLT unit tests marked as XFAIL now?

Some are not working, I didn't have time to fix them after the major changes that we had.

psydac/api/tests/navier_stockes.py
saidctb: done

psydac/api/ast/nodes.py
saidctb: those represent the indices used in the loops and the arrays, and yes they are singletons.

psydac/api/grid.py
saidctb: yes it's used for any two spaces and yes it's a bug fix.
I'm not familiar with the problem in the histopolation, don't we use the same spaces always in the histopolation case?

psydac/api/grid.py
saidctb: it represents the union of the indices calculated by `FemAssemblyGrid`, we make sure that they match between all the spaces.

psydac/linalg/block.py
```diff
@@ -500,12 +500,12 @@ class BlockMatrix( BlockLinearOperator, Matrix ):
...
     # Abstract interface
     #--------------------------------------
-    def toarray( self, *, order='C', with_pads=False ):
+    def toarray( self, *, order='C' ):
         """ Convert to Numpy 2D array. """
         return self.tosparse(order=order).toarray()
 
     # ...
-    def tosparse( self, *, order='C', with_pads=False ):
+    def tosparse( self, *, order='C' ):
```
yguclu: It's interesting that Codacy does not complain about the fact that these methods have a signature different from that of the base class, which is
```
def toarray(self, **kwargs):
```
So did Codacy complain if you kept the same signature in the classes `StencilVector` and `StencilMatrix`?

psydac/fem/grid.py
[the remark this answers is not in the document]
yguclu: What about the case where the current space is not a reduction, with `trial_space==test_space` and `pad==degree`? You get then

* if `multiplicity == 1`, then `m = 0`
* if `multiplicity > 1`, then `m = multiplicity`

So why is that? You should really explain the role of `m` in the code below, as I wrote before.

psydac/ddm/cart.py
```diff
@@ -72,13 +72,13 @@ def __init__( self, npts, pads, periods, reorder, comm=MPI.COMM_WORLD, multiplic
...
+        shifts = tuple(shifts) if shifts else (1,)*len(npts)
 
         # Store input arguments
         self._npts         = tuple( npts    )
         self._pads         = tuple( pads    )
         self._periods      = tuple( periods )
-        self._multiplicity = multiplicity
+        self._shifts = shifts
         self._reorder      = reorder
         self._comm         = comm
```
yguclu: Please align the equal signs, as it improves readability.

Specifically it allows developers and reviewers to scroll through the code much faster.

psydac/core/bsplines.py
```diff
@@ -476,14 +476,17 @@ def breakpoints( knots, degree ):
     degree : int
         Polynomial degree of B-splines.
 
+    tol: float
+        Maximum tolerance between the break points.
```
yguclu, line 480: I think it's best to be very explicit in this low-level module. So I suggest to add "If the distance between two knots is less than `tol`, we assume that they are repeated knots which correspond to the same break point."

psydac/core/bsplines.py
```diff
@@ -666,7 +671,11 @@ def elevate_knots(knots, degree, periodic, multiplicity=1):
         True if domain is periodic, False otherwise.
 
     multiplicity : int
-        multiplicity of the knots.
+        Multiplicity of the knots in the knot sequence, we assume that the same 
+        multiplicity applies to each interior knot.
+
+    tol: float
+        Maximum tolerance between the break points.
```
yguclu, line 678: See previous comment

psydac/cad/geometry.py
```diff
@@ -446,7 +446,7 @@ def refine_nurbs(nrb, ncells=None, degree=None, multiplicity=None, tol=1e-9):
         multiplicity of each knot in the knot sequence in each direction
 
     tol : <float>
-        tolerance
```
yguclu: Don't you mean this instead?
```python
"""
tol : <float>
    Minimum distance between two break points.
"""
```
To which I would add 

"Two knots with a distance less than `tol` are considered to be repeated knots, which correspond to the same break points"

psydac/cad/geometry.py
```diff
@@ -519,7 +519,7 @@ def refine_knots(knots, ncells, degree, multiplicity=None, tol=1e-9):
...
+        Maximum tolerance accepted between two break points.
```
yguclu, line 522: See comment above

psydac/api/discretization.py
```diff
@@ -227,12 +227,12 @@ def discretize_space(V, domain_h, *args, **kwargs):
...
             Vh = reduce_space_degrees(V, Vh, basis=basis)
 
-        Vh._symbolic_space = V
+        Vh.symbolic_space = V
         g_spaces[inter]    = Vh
 
     Vh = ProductFemSpace(*g_spaces.values())
 
-    Vh._symbolic_space      = V
+    Vh.symbolic_space      = V
```
yguclu: Well done! Using property setters is good practice.

yguclu (changes requested):

psydac/linalg/block.py
saidctb: it did not complain

psydac/ddm/cart.py
saidctb: done

psydac/core/bsplines.py
saidctb: done

psydac/core/bsplines.py
saidctb: done

psydac/cad/geometry.py
saidctb: done

psydac/cad/geometry.py
saidctb: done

psydac/api/grid.py
yguclu: OK then, if this is a bug fix it is important to add a point about it to the PR description!

Indeed there is only one space involved in the histopolation, so the issue there was a bit different: we were performing a simple Gauss quadrature within each interval, which was defined by the Greville points of the parent space.

Since the histopolation was used as part of a projection operator, it should be exact when applied to a spline of the space you are projecting to. But unfortunately this was not the case, because quadrature intervals could contain knots and Gauss quadrature does not work for piecewise polynomials!

I hope I could explain myself...

psydac/api/grid.py
yguclu: Please add a comment to the code!

psydac/fem/grid.py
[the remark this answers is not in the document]
saidctb: it is used in the case of `pad==degree` because we can't know if the `trial_space==test_space`, here we compute the basis function of each space independently of each other.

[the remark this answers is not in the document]
saidctb: in the case where `multiplicity`>1 and the `test_space != trial_space`, we will need more basis functions (m more basis functions) otherwise the computation will be false, otherwise, the computation is the same as before

psydac/linalg/stencil.py
```diff
@@ -366,7 +365,7 @@ def __str__(self):
         return txt
 
     # ...
-    def toarray( self, *, order='C', with_pads=False ):
+    def toarray( self, **kwargs ):
```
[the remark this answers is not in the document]
saidctb: done

psydac/linalg/stencil.py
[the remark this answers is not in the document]
yguclu: This is good enough for now. We'll improve in a future refactoring of Psydac

psydac/fem/splines.py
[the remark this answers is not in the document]
yguclu: I see that this was done. Thanks!

psydac/cad/geometry.py
yguclu: OK

psydac/linalg/block.py
yguclu: If it did not complain, why did you not keep this signature in those classes?
```python
def toarray(self, *, with_pads=False):
```
This is more informative than `**kwargs`, which is only needed in the base class.

yguclu (commented):
Thanks for the latest changes.

I have just checked every conversation in this PR, and closed many discussions... but many are still open, please have a look!

psydac/linalg/stencil.py
```diff
@@ -15,6 +15,15 @@
...
 
+#===============================================================================
+def compute_diag_len(p, md, mc, return_padding=False):
+    n = ((np.ceil((p+1)/mc)-1)*md).astype('int')
+    ep = np.minimum(0, n-p)
+    n = n-ep + p+1
+    if return_padding:
+        return n.astype('int'), (-ep).astype('int')
+    else:
+        return n.astype('int')
```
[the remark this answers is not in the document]
saidctb: done

psydac/linalg/stencil.py
```diff
@@ -131,6 +149,15 @@ def zeros( self ):
...
 
+    def reduce_elements(self, axes, elements):
+        assert not self.parallel
+        npts         = [n-ne for n,ne in zip(self.npts, elements)]
+        multiplicity = [max(1,m-1) for m in self.multiplicity]
+
+        v = StencilVectorSpace(npts=npts, pads=self.pads, periods=self.periods, multiplicity=multiplicity)
+        v._parent_starts = self.starts
+        v._parent_ends   = self.ends
+        return v
```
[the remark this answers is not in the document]
saidctb: done

psydac/api/ast/nodes.py
saidctb: yes, `args` are inherited from the `Function` class,  so I have changed them to `arg1` and `arg2` to be more clear.

psydac/api/ast/nodes.py
```diff
@@ -265,7 +325,10 @@ def __new__(cls, quads, indices_basis, q_basis, l_basis, mapping, components, sp
         for test,mapp,comp in zip(test_atoms, mapping_atoms, components):
             test    = AtomicNode(test)
             val     = ProductGenerator(MatrixQuadrature(mapp), quads)
-            rhs     = Mul(CoefficientBasis(comp),test)
+            if is_rational:
+                rhs     = Mul(CoefficientBasis(comp),CoefficientBasis(weight),test)
+            else:
+                rhs     = Mul(CoefficientBasis(comp),test)
```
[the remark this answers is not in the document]
saidctb: yes, this is the bug fix

psydac/linalg/block.py
saidctb: for some reason, it does not complain about the Block Matrix, but it does for the StencilMatrix

psydac/api/ast/nodes.py
```diff
@@ -357,7 +428,7 @@ def __new__(cls, target, dummies):
             dummies = [dummies]
         dummies = Tuple(*dummies)
 
-        if not isinstance(target, (ArrayNode, MatrixNode)):
+        if not isinstance(target, (ArrayNode, MatrixNode, Expr)):
             raise TypeError('expecting an ArrayNode')
```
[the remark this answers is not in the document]
saidctb: done

psydac/api/ast/nodes.py
saidctb: done

psydac/api/ast/nodes.py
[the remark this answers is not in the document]
saidctb: I have added it

psydac/api/grid.py
saidctb: I see, but this class is only used for the API, so we don't use this part of the code to calculate the histopolation matrix

psydac/api/grid.py
saidctb: done

saidctb:
> Thanks for the latest changes.
> 
> I have just checked every conversation in this PR, and closed many discussions... but many are still open, please have a look!

Thanks!, I have fixed the remaining ones, please take a look.

psydac/api/ast/nodes.py
```diff
@@ -294,9 +329,41 @@ class RAT(Basic):
...
+            The dummy variable for the test functions
+
+        nderiv  : <int>
+            Maximum number of derivatives
+
+        mask    : int,optional
+            The fixed direction in case of a boundary integral
+
+        is_irational: bool,optional
+            True if the mapping is irational
```
yguclu, line 364: Same typos on both lines: `irational` -> `rational` (without the leading `i`)

psydac/linalg/stencil.py
```diff
@@ -16,14 +16,31 @@
...
+
+        shifts_domain : tuple_like (int)
+         Shifts of the domain along each direction
+
+        mshifts_codomainc : tuple_like (int)
+         Shifts of the codomain along each direction
+
+        return_padding : bool
+            Return the new padding if True
+        """
```
yguclu, line 35: Nice docstring!

Please add the `Returns` section, which is missing, and fix a few typos:
- lenght -> length (swap positions of `t` and `h`)
- mshifts_codomainc -> shifts_codomain (remove leading `m` and trailing `c`)

Also, please add an empty line before the `Parameters` section.

yguclu (changes requested):
Thanks for the changes! I have found a couple of typos

psydac/api/ast/nodes.py
saidctb: lol my bad

saidctb: done

psydac/linalg/stencil.py
saidctb: done

psydac/linalg/stencil.py
```diff
@@ -27,17 +28,25 @@ def compute_diag_len(pads, shifts_domain, shifts_codomain, return_padding=False)
...
          Shifts of the codomain along each direction
 
         return_padding : bool
             Return the new padding if True
-        """
+    
+        Returns
+        -------
+        n : (int)
+         Diagonal lenght of the stencil matrix
```
yguclu, line 40: Again the same typo!! 🤣 You may want to use a spell checker...

lenght -> length

psydac/linalg/stencil.py
```diff
@@ -27,17 +28,25 @@ def compute_diag_len(pads, shifts_domain, shifts_codomain, return_padding=False)
...
+
+        ep : (int)
+          Starting index of the non zero elements
```
yguclu, line 43: Is this the padding? If so please explain it in the docstring

yguclu (changes requested):

psydac/linalg/stencil.py
saidctb: lol , done

psydac/linalg/stencil.py
saidctb: yes ,it's the padding, I have added it to the docstring

yguclu (approved):

PR #139, opened 2021-08-28
Do not require old Pyccel version 0.10.1 at installation

yguclu:
Remove hard dependency of Psydac AST on obsolete Pyccel version 0.10.1 by copying the relevant Pyccel files into the submodule `psydac.pyccel`.

While Psydac uses this old version of Pyccel to generate Python code, the acceleration is performed by the version of Pyccel installed on the machine.

### DIFF
--- a/psydac/pyccel/ast/core.py
+++ b/psydac/pyccel/ast/core.py
@@ -16,7 +16,7 @@ from sympy import preorder_traversal
 from sympy.simplify.radsimp   import fraction
 from sympy.core.compatibility import with_metaclass
 from sympy.core.assumptions   import StdFactKB
-from sympy.core.singleton     import Singleton, S
+from sympy.core.singleton     import S
 from sympy.core.function      import Function, Application
 from sympy.core.function      import Derivative, UndefinedFunction as sp_UndefinedFunction
 from sympy.core.function      import _coeff_isneg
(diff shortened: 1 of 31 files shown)


psydac/pyccel/ast/core.py
```diff
@@ -0,0 +1,5703 @@
...
+
+from .numbers        import BooleanTrue, BooleanFalse, Integer as Py_Integer, ImaginaryUnit
+from .functionalexpr import GeneratorComprehension as GC
+from .functionalexpr import FunctionalFor
+
+from ..errors.errors import Errors
+from ..errors.messages import *
+
+errors = Errors()
+
```
saidctb, line 50: We don't need to copy all the content of the file, we only keep what we import from this file

saidctb (changes requested):

psydac/pyccel/ast/core.py
yguclu: Yes, this can be done for every file in principle. But I don't think it is a useful investment of time given that this is an "intermediate" temporary solution. 😉

yguclu:
@saidctb We are running into this error, which stems from Pyccel when used in parallel:
```
ModuleNotFoundError: No module named 'tblib'
```
I think we should add it to the requirements of Psydac.

psydac/pyccel/ast/core.py
yguclu: Also, given that we use the `import *` syntax, there is no simple way to know what we really use from that module.

saidctb (approved):

PR #154, opened 2022-01-13
Update README file

anlavandier:
- Use GitHub Flavored Markdown instead of RestructuredText
- Add instructions for installation of dependencies
- Explain how to launch the unit tests


README.rst
```diff
@@ -18,7 +18,10 @@ Requirements
 
 - All *Python* dependencies can be installed using::
 
-    sudo -H pip3 install -r requirements.txt
+    export CC="mpicc"
+    export HDF5_MPI="ON"
+    export HDF5_DIR=/usr/lib/x86_64-linux-gnu/hdf5/openmpi
```
yguclu, line 23: This path is for Ubuntu, it will be different on other operating systems

README.rst
```diff
@@ -37,7 +40,15 @@ Uninstall
...
 
     python3 -m pip uninstall psydac
+
+Running tests
+*********
+- **Setup**::
     
+   export PSYDAC_MESH_DIR=/path/to/psydac/mesh/
+   python -m pytest --pyargs psydac -m "not parallel"
+   python /path/to/psydac/mpi_tester.py --pyargs psydac -m "parallel"
```
yguclu, line 50: We should probably change `python` with `python3`

anlavandier: I didn't what made more sense because I assume that most people only have python 3 and use python as a way to launch the python3 executable.

yguclu (approved):
Good job @anlavandier! Thanks

Issue #170, opened 2022-02-11
Add automatic testing for examples


yguclu:
The scripts in the `examples` directory are not tested by our CI setup, therefore they can become obsolete and useless. We should restructure them so that they include at least one unit test (e.g. a simulation with small parameters) which will be run by Pytest.

PR #183, opened 2022-02-16
Update signature of Mapping methods

### LINKED ISSUE #182: Update signature of Mapping methods
The signatures of the evaluation methods of class `Mapping` require an array_like `eta` instead of a variable number of arguments `*eta` (one for each component). The latter solution allows for passing array inputs, and is consistent with the evaluation methods of the `TensorFemSpace` class.

Hence the current signatures

```python
def __call__( self, eta ):
def jac_mat( self, eta ):
def metric( self, eta ):
def metric_det( self, eta ):
```

should be changed to

```python
def __call__(self, *eta):
def jac_mat(self, *eta):
def metric(self, *eta):
def metric_det(self, *eta):
```

Further, the function calls should be updated throughout the library.

anlavandier:
fixes #182

### DIFF
--- a/psydac/cad/tests/test_geometry.py
+++ b/psydac/cad/tests/test_geometry.py
@@ -195,9 +195,9 @@ def test_export_nurbs_to_hdf5(ncells, degree):
     eta2 = refine_array_1d( new_pipe.breaks(1), 10 )
 
     pcoords1 = np.array( [[new_pipe(e1,e2) for e2 in eta2] for e1 in eta1] )
-    pcoords2 = np.array( [[mapping([e1,e2]) for e2 in eta2] for e1 in eta1] )
+    pcoords2 = np.array( [[mapping(e1,e2) for e2 in eta2] for e1 in eta1] )
 
-    assert np.allclose(pcoords1[...,:domain.dim], pcoords2, 1e-15, 1e-15)
+    assert np.allclose(pcoords1[..., :domain.dim], pcoords2, 1e-15, 1e-15)
 
 @pytest.mark.parametrize( 'ncells', [[8,8], [12,12], [14,14]] )
 @pytest.mark.parametrize( 'degree', [[2,2], [3,2], [2,3], [3,3], [4,4]] )
--- a/psydac/mapping/discrete.py
+++ b/psydac/mapping/discrete.py
@@ -126,17 +126,17 @@ class SplineMapping:
     #--------------------------------------------------------------------------
     # Abstract interface
     #--------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


psydac/cad/tests/test_geometry.py
```diff
@@ -195,9 +195,9 @@ def test_export_nurbs_to_hdf5(ncells, degree):
     eta2 = refine_array_1d( new_pipe.breaks(1), 10 )
 
     pcoords1 = np.array( [[new_pipe(e1,e2) for e2 in eta2] for e1 in eta1] )
-    pcoords2 = np.array( [[mapping([e1,e2]) for e2 in eta2] for e1 in eta1] )
+    pcoords2 = np.array( [[mapping(e1,e2) for e2 in eta2] for e1 in eta1] )
```
yguclu, line 198: Any hope that this will work with Numpy broadcasting? Something like the following:
```suggestion
    ee1, ee2 = np.meshgrid(eta1, eta2, indexing='ij', sparse=True)
    pcoords1 = new_pipe(ee1, ee2)
    pcoords2 =  mapping(ee1, ee2)
```

yguclu (commented):
Can you please try and run the examples in the `psydac/examples` directory and see if they work? This is also related to issue #170...

yguclu:
There are several unit tests across Psydac that contain some plotting utilities, which are only used when these files are run from the command line. I think some of them also use the evaluation routines of the `Mapping` class, so you should probably run a search.

psydac/cad/tests/test_geometry.py
anlavandier: It doesn't because `TensorFemSpace.eval_field` doesn't accept arrays because `psydac.core.bsplines.find_span` and `basis_funs` do not either.

anlavandier:
> > Can you please try and run the examples in the `psydac/examples` directory and see if they work? This is also related to issue #170...
> 
> `poisson_2d_mapping` uses classes that are not in `psydac` anymore namely:
> 
> ```python
> from psydac.mapping.analytical         import AnalyticalMapping, IdentityMapping
> from psydac.mapping.analytical_gallery import Annulus, Target, Czarny
> ```
> 
> Because `visualize_mass_matrix`and test_mass_matrix`rely on`poisson_2d_mapping` they also do not work. The rest work fine.

The same outdated classes are used in the `visual_tests` of the `mapping/tests` sub-directory.

psydac/cad/tests/test_geometry.py
yguclu: Right! I forgot that these are spline mappings...

PR #187, opened 2022-02-21
Fix 2D Poisson example

yguclu:
Make sure that the low-level example script `psydac/examples/poisson_2d_mapping.py` works with the latest version of Psydac:

- Use `CallableMapping` from `sympde.topology.callable_mapping` instead of `AnalyticalMapping` from `psydac.mapping.analytical` (which does not exist anymore)
- Use gallery of analytical mappings from `SymPDE`
- Use `Geometry` class from `psydac.cad.geometry` to read and write HDF5 geometry files
- Use new signature of mapping evaluation methods
- Bugfix in library: change (eta) to (*eta) when calling jac_mat and metric

### DIFF
--- a/examples/poisson_2d_mapping.py
+++ b/examples/poisson_2d_mapping.py
@@ -623,13 +623,14 @@ def main( *, test_case, ncells, degree, use_spline_mapping, c1_correction, distr
     map_analytic = model.mapping
 
     if use_spline_mapping:
-        map_discrete = SplineMapping.from_mapping( V, map_analytic )
-        mapping = map_discrete
+        map_discrete = SplineMapping.from_mapping( V, map_analytic.symbolic_mapping )
+        map_discrete.jacobian = map_discrete.jac_mat  # needed after changes in mapping classes
         # Write discrete geometry to HDF5 file
         t0 = time()
-        mapping.export( 'geo.h5' )
+        map_discrete.export( 'geo.h5' )
         t1 = time()
         timing['export'] += t1-t0
+        mapping = map_discrete
     else:
         mapping = map_analytic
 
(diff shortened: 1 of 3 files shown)


examples/poisson_2d_mapping.py
```diff
@@ -22,15 +26,26 @@
...
 
-        sym = type(mapping).symbolic
+        sym = mapping.symbolic_mapping
 
-        self._eta        = sym.eta
-        self._metric     = sym.metric    .subs( mapping.params )
-        self._metric_det = sym.metric_det.subs( mapping.params )
+        self._eta        = sym.logical_coordinates
+        self._metric     = sym.metric_expr    .subs(mapping.params)
+        self._metric_det = sym.metric_det_expr.subs(mapping.params)
```
yguclu, line 48: @saidctb Can a `CallableMapping` object have free parameters? The old `AnalyticalMapping` stored a dictionary of `(symbol, value)` pairs in `params`, but this looks empty now.

saidctb: yes, it can have free parameters, but was never tested,  we usually fix the free parameters when we create an instance of the Mapping class

yguclu: Do you mean [here](https://github.com/pyccel/sympde/blob/7e2a5622bb2eb3bf4a592db9113b1c9f9fdd4dcf/sympde/topology/callable_mapping.py#L20-L32)? I do not know how these lines could work:
```python
        constants    = mapping.constants
        params       = {a.name:a for a in constants}
        params.update( kwargs )
        for p in params.values():
            assert not isinstance(p, Symbol)


        if params:
            subs = {a:params[a.name] for a in constants}
            expressions = expressions.subs(subs)
            jac         = jac.subs(subs)
            inv_jac     = inv_jac.subs(subs)
            metric      = metric.subs(subs)
            metric_det  = metric_det.subs(subs)
```
What type of objects are in `mapping.constants`?

yguclu: I have just created issue #189 for discussing about this. I will now remove the commented lines and proceed with merging this small PR.

saidctb (approved):

examples/poisson_2d_mapping.py
saidctb: okay, sounds good

PR #208, opened 2022-03-21
Fixing `PostProcessManager.export_to_vtk`

anlavandier:
This PR fixed `PostProcessManager.export_to_vtk` which did not do anything.
It consists of:
*  Putting starred expressions where they need to be
* Adding a `debug` argument to `export_to_vtk` that returns what is computed in the function. 
* Adding `assert` statements to check if the right fields are exported.


psydac/api/tests/test_postprocessing.py
```diff
@@ -47,6 +47,7 @@ def test_OutputManager():
 
     try:
         Om.export_fields(uh_static=uh)
+        raise ValueError
     except AssertionError:
         pass
```
yguclu, line 52: Is the purpose of this piece of code to check that an assertion is raised? If so, there is a specific way to do so with Pytest:

```suggestion
    with pytest.raises(AssertionError):
        Om.export_fields(uh_static=uh)
```
If the correct exception is not raised, the test will fail.

psydac/api/postprocessing.py
```diff
@@ -909,7 +916,7 @@ def export_to_vtk(self, filename_pattern, grid, npts_per_cell=None, snapshots='n
 
         smart_eval_dict = {}
         for i, snapshot in enumerate(snapshots):
-            self.load_snapshot(snapshot, tuple(fields.keys()))
+            self.load_snapshot(snapshot,*fields.keys())
```
yguclu, line 919: Missing empty space after the comma

psydac/api/postprocessing.py
```diff
@@ -791,32 +789,36 @@ def load_snapshot(self, n, *fields):
...
             Number of leading zeros in the time indexing of the files. 
             Only used if ``snapshot`` is not ``'none'``. 
 
-        fields: dict
+        fields : dict
             Dictionary with the fields to export as keys and the name under which to export them as values
+        
+        debug : bool, default=False
+            If true, returns ``(mesh, pointData_list)`` where ``mesh`` is ``(x_mesh, y_mesh,  z_mesh)``
+            and ``pointData_list`` is the list of all the pointData dictionnaries.
```
yguclu, line 821: Typo: dictionnaries -> dictionaries

yguclu (changes requested):

psydac/api/postprocessing.py
anlavandier: Fixed in 3e32950

yguclu (approved):

PR #216, opened 2022-04-29
Use Comm.Allreduce instead of Comm.allreduce

saidctb:
The `allreduce` function of mpi4py uses serialization which is slower than using `Allreduce` function that uses numpy arrays.

To illustrate that we run this [script](https://github.com/pyccel/psydac/blob/devel-allreduce/examples/test_cg_perf.py) using the old implementation vs the new implementation. 
 
```
using Comm.allreduce ...
number of procs :: 72
CG info      :: {'niter': 5003, 'success': True, 'res_norm': 8.064672720977876e-14}
Elapsed time :: 3.9431095123291016

using Comm.Allreduce ...
number of procs :: 72
CG info      :: {'niter': 5003, 'success': True, 'res_norm': 7.648109028693814e-14}
Elapsed time :: 1.0101466178894043
```

### DIFF
--- a/examples/test_cg_perf.py
+++ b/examples/test_cg_perf.py
@@ -1,53 +1,119 @@
+import pytest      
 import time
+import matplotlib.pyplot as plt
 import numpy as np
 
-from mpi4py       import MPI
-from psydac.ddm.cart       import CartDecomposition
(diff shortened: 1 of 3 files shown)


yguclu (approved):
Impressive speed-up! Well done for spotting this issue!!

Issue #54, opened 2019-09-26
Remove Fortran source files from repository


yguclu:
The directory **psydac/core** contains several files with extensions **f90**, **F90**, **c** and **pyf**, which are used to create a Python extension module `.so` with a certain number of low-level Fortran spline routines.

We should re-implement all those routines in pure Python, so that the installation of PSYDAC will not necessarily require a Fortran compiler.

In the future we could recover Fortran-level efficiency for these low-level routines by using Pyccel on the annotated Python code during the library setup. This would be an optional optimization, which the user could skip.

yguclu:
As mentioned by @ratnania in duplicated issue #86:

> I don't think we still need these files.
> BTW, I am implementing the missing Python functions such as the refinement matrix needed for the Multigrid.

That's good to know! Feel free to create a new PR with the new functions.

PR #217, opened 2022-05-05
Remove unused Fortran Files

anlavandier:
The fortran files in `psydac/core/external` are only used in `psydac/utilities/integrate.py` which itself isn't used anywhere. For this reason, those fortran files can be removed from the library. 
Fixes issue #54


yguclu (approved):
We can now merge this PR. Thank you @anlavandier!

PR #222, opened 2022-05-23
Add Jacobian matrix pointwise evaluation to `NurbsMapping`

anlavandier:
This PR simply adds pointwise evaluation of the Jacobian related quantities for the `NurbsMapping` class. 

`psydac/core/tests/test_kernels.py` doesn't pass because of an error is the kernel functions that has been fixed in #214


yguclu (approved):
Thanks for looking into NURBS and adding this feature!

PR #248, opened 2022-11-07
Accelerate tosparse methods

saidctb:
Add low level `tosparse` methods in `linalg/kernels.py`


psydac/api/ast/linalg.py
```diff
@@ -106,7 +106,8 @@ def __hashable_new__(cls, ndim, block_shape, comm_id=None, **kwargs):
 
         # Generate random tag, unique for all processes in MPI communicator
         tag = random_string(8)
-        if comm is not None and comm.size>1:
+        parallel = True
+        if comm is not None and comm.size>1 and parallel:
```
yguclu, line 110: I don't understand why you set `parallel = True` before the if block, rather than inside it

psydac/api/ast/linalg.py
```diff
@@ -457,7 +465,8 @@ def __hashable_new__(cls, ndim, comm_id=None, **kwargs):
...
```
yguclu, line 469: See previous question about `parallel`

psydac/api/ast/linalg.py
```diff
@@ -419,15 +420,22 @@ def _compile_pyccel(self, mod, backend, verbose=False):
...
+                           bcast        = True,
+                           folder       = _PYCCEL_FOLDER,
+                           verbose      = verbose)
+        else:
+            fmod = epyccel(mod,
+                           accelerators = accelerators,
+                           compiler     = compiler,
+                           fflags       = fflags,
+                           folder       = _PYCCEL_FOLDER,
+                           verbose      = verbose)
```
yguclu, line 438: I believe that `bcast = True` is the default. Could you not avoid the code duplication and write simply
```suggestion
        fmod = epyccel(mod,
                       accelerators = accelerators,
                       compiler     = compiler,
                       fflags       = fflags,
                       comm         = self.comm if parallel else None,
                       folder       = _PYCCEL_FOLDER,
                       verbose      = verbose)
```
Also, are you always OK with using the MPI rank 0 in the communicator to accelerate the code? We pass `root = 0` by default.

psydac/api/ast/linalg.py
```diff
@@ -574,14 +583,22 @@ def _compile_pyccel(self, mod, backend, verbose=False):
...
```
yguclu, line 601: See previous comment about code duplication

psydac/api/ast/parser.py
```diff
@@ -486,14 +486,14 @@ def _visit_DefNode(self, expr, **kwargs):
...
-            arguments += [self._visit(i, **kwargs) for i in constants]
-
         if f_args:
             f_args     = [self._visit(i, **kwargs) for i in f_args]
             f_args     = [tuple(arg.values())[0] if isinstance(arg, dict) else arg for arg in f_args]
             arguments += flatten(f_args)
 
+        if constants:
+            arguments += [self._visit(i, **kwargs) for i in constants]
+
```
yguclu, line 496: So you changed the order of the arguments? I think that it should be documented somewhere, maybe in the docstring of this method? Suggestions are welcome

psydac/api/discretization.py
```diff
@@ -271,7 +271,11 @@ def discretize_space(V, domain_h, *, degree=None, multiplicity=None, knots=None,
     connectivity = construct_connectivity(domain)
     if isinstance(domain_h, Geometry) and all(domain_h.mappings.values()):
         # from a discrete geoemtry
-        mappings  = [domain_h.mappings[inter.logical_domain.name] for inter in interiors]
+        if interiors[0].name in domain_h.mappings:
+            mappings  = [domain_h.mappings[inter.name] for inter in interiors]
+        else:
+            mappings  = [domain_h.mappings[inter.logical_domain.name] for inter in interiors]
+
```
yguclu, line 278: Can you explain this change?

psydac/api/equation.py
```diff
@@ -93,7 +93,7 @@ def l2_boundary_projection(equation):
     test_dict = dict(zip(u, v))
 
     # Compute product of (u, v) using dot product for vector quantities
-    product  = lambda f, g: (f * g if isinstance(g, ScalarFunction) else dot(f, g))
+    product  = lambda f, g: (f * g if g.atoms(ScalarFunction) else dot(f, g))
```
yguclu, line 96: Please explain

psydac/linalg/block.py
```diff
@@ -1194,7 +1194,10 @@ def set_backend(self, backend):
             dm.append(self._blocks[key]._dotargs_null['dm'])
 
         if self.domain.parallel:
-            comm = self.codomain.spaces[0].cart.comm if isinstance(self.codomain, BlockVectorSpace) else self.codomain.cart.comm
+            if interface:
+                comm = self.domain.spaces[0].interfaces[d_axis, d_ext].cart.local_comm if isinstance(self.domain, BlockVectorSpace) else self.domain.interfaces[d_axis, d_ext].cart.local_comm 
+            else:
+                comm = self.codomain.spaces[0].cart.comm if isinstance(self.codomain, BlockVectorSpace) else self.codomain.cart.comm
```
yguclu, line 1200: Please explain these changes. Is this a bug fix?

psydac/mapping/discrete_gallery.py
```diff
@@ -15,9 +15,9 @@
 
 class Collela3D( Mapping ):
 
-    expressions = {'x':'2.*(x1 + 0.1**sin(2.*pi*x1)*sin(2.*pi*x2)) - 1.',
-                   'y':'2.*(x2 + 0.1**sin(2.*pi*x1)*sin(2.*pi*x2)) - 1.',
-                   'z':'2.*x3  - 1.'}
+    _expressions = {'x':'2.*(x1 + 0.1*sin(2.*pi*x1)*sin(2.*pi*x2)) - 1.',
+                    'y':'2.*(x2 + 0.1*sin(2.*pi*x1)*sin(2.*pi*x2)) - 1.',
+                    'z':'2.*x3  - 1.'}
```
yguclu, line 20: Clearly we were not using this class before.
Should we not add it to `sympde/topology/analytical_mapping.py`?

psydac/linalg/kernels.py
```diff
@@ -0,0 +1,163 @@
...
+                ii3 = s3+i3
+                I   = ii3*nr1*nr2 + ii2*nr1 + ii1
+                for j1 in range(ncl1):
+                    for j2 in range(ncl2):
+                        for j3 in range(ncl3):
+                            value = A[i1+pp1,i2+pp2,i3+pp3,j1,j2,j3]
+                            if abs(value)<1e-15:continue
+                            jj1 = ((ii1//cm1)*dm1+j1-dp1)%nc1
+                            jj2 = ((ii2//cm2)*dm2+j2-dp2)%nc2
+                            jj3 = ((ii3//cm3)*dm3+j3-dp3)%nc3
```
yguclu, line 155: The modulo operator `%` is expensive, and we might want to avoid it when the boundary conditions are not periodic. Maybe put a TODO item in a comment?

psydac/linalg/kernels.py
```diff
@@ -0,0 +1,163 @@
...
+            for i3 in range(nrl3):
+                ii1 = s1+i1
+                ii2 = s2+i2
+                ii3 = s3+i3
+                I   = ii3*nr1*nr2 + ii2*nr1 + ii1
+                for j1 in range(ncl1):
+                    for j2 in range(ncl2):
+                        for j3 in range(ncl3):
+                            value = A[i1+pp1,i2+pp2,i3+pp3,j1,j2,j3]
+                            if abs(value)<1e-15:continue
```
yguclu, line 152: Why hard-coding this tolerance here?

psydac/linalg/kernels.py
```diff
@@ -0,0 +1,163 @@
...
+                            jj2 = ((ii2//cm2)*dm2+j2-dp2)%nc2
+                            jj3 = ((ii3//cm3)*dm3+j3-dp3)%nc3
+
+                            J   = jj3*nc1*nc2 + jj2*nc1 + jj1
+
+                            rows[ind] = I
+                            cols[ind] = J
+                            data[ind] = value
+                            ind += 1
+    return ind
```
yguclu, line 163: Does the `ind` variable represent the number of non-zero elements? If so, I suggest to rename it.

yguclu (changes requested):

psydac/linalg/kernels.py
saidctb: yes, it represents the number of non zero elements

psydac/api/ast/linalg.py
saidctb: I will delete this part, I had a strange bug, and as a quick fix I did those changes, but I have fixed it properly

psydac/linalg/kernels.py
saidctb: should we allow the user to set it?

psydac/linalg/kernels.py
saidctb: ah yes, this can be improved

psydac/linalg/kernels.py
yguclu: I would say so. But in any case I am not convinced that we even need to remove the small entries. In all situations I can think of, our entries are either (relatively) large or exactly zero.

yguclu:
@saidctb: what is the status of this PR? As soon as you address the comments we could merge it...

psydac/api/ast/linalg.py
saidctb: it deleted in [0a922eb](https://github.com/pyccel/psydac/pull/248/commits/0a922ebe5711d1e83d5f158155d0b4ca99315bae)

psydac/api/ast/linalg.py
saidctb: it's deleted in [0a922eb](https://github.com/pyccel/psydac/pull/248/commits/0a922ebe5711d1e83d5f158155d0b4ca99315bae)

psydac/api/ast/linalg.py
saidctb: deleted in [0a922eb](https://github.com/pyccel/psydac/pull/248/commits/0a922ebe5711d1e83d5f158155d0b4ca99315bae)

psydac/api/ast/linalg.py
saidctb: see comment above

psydac/linalg/kernels.py
saidctb: I'm not sure, I have encountered small entries  many times

psydac/linalg/kernels.py
saidctb: done [7c15659](https://github.com/pyccel/psydac/pull/248/commits/7c15659b079742fedad0ff09b0450822e12cf815)

psydac/linalg/kernels.py
saidctb: done [7c15659](https://github.com/pyccel/psydac/pull/248/commits/7c15659b079742fedad0ff09b0450822e12cf815)

psydac/api/equation.py
saidctb: The previous implementation works only with homogeneous Neumann boundary conditions, so I changed it to add the case of non-homogeneous Neumann boundary conditions

psydac/api/discretization.py
saidctb: In some cases, the mappings dictionary has the logical domain interiors as a keyword and in some other cases the physical domain interiors.

psydac/api/ast/parser.py
saidctb: added in [a31334c](https://github.com/pyccel/psydac/pull/248/commits/a31334cb664a6ebc9abd1bd4411bdae7d7609ec2)

psydac/linalg/block.py
saidctb: This was a bug fix, we were not passing the right communicator in the multipath case

psydac/api/discretization.py
yguclu: This suggests that we have a fundamental inconsistency (i.e. a bug) somewhere in our code base. If you can describe when the two situations happen, please create an issue!

psydac/linalg/kernels.py
yguclu: If you really want to set the small entries to zero you should use a relative tolerance. If the maximum absolute value in the assembled matrix is small (e.g. `1e-8`), you will remove an entry that is not really small.

psydac/api/equation.py
yguclu: Was this happening for the biharmonic equation?

psydac/linalg/kernels.py
saidctb: I'm only taking out the values smaller than `1e-15` which is the machine error, so even in your case it wouldn't make sense to keep them.

yguclu: The machine error is a **relative** error, not an absolute one. The smallest double precision floating-point number (in absolute value) is `5.4e-79`.

saidctb: I see I will remove only the zero entries then

saidctb: done in [57d7134](https://github.com/pyccel/psydac/pull/248/commits/57d7134627a9f1118589701918f935b5b549619f)

psydac/api/discretization.py
saidctb: done https://github.com/pyccel/psydac/issues/264

yguclu: Thanks!

psydac/api/equation.py
yguclu: I see thanks

yguclu (approved):

PR #249, opened 2022-11-10
Fix bug in discretization of non-symmetric bilinear operators

vcarlier:


psydac/api/ast/fem.py
```diff
@@ -352,8 +352,8 @@ def __init__(self, expr, terminal_expr, spaces, mapping_space=None, tag=None, ma
             fields              = tuple(i for i in atoms if i not in tests+trials)
             is_broken           = spaces[1].symbolic_space.is_broken
             quad_order          = get_quad_order(spaces[1])
-            tests_degrees       = get_degrees(tests, spaces[0])
-            trials_degrees      = get_degrees(trials, spaces[1])
+            tests_degrees       = get_degrees(tests, spaces[1])
+            trials_degrees      = get_degrees(trials, spaces[0])
```
yguclu, line 356: @saidctb It am surprised that we have not encountered this bug before.

yguclu (approved):
Thanks @vcarlier! This looks good to me

saidctb (approved):

psydac/api/ast/fem.py
saidctb: yeah, I think because we where always using the same spaces for the test and trials

PR #251, opened 2022-11-15
Fix unit tests in feec/multipatch/tests

vcarlier:
After adding an empty init in directory feec/multipatch/tests, these run but they all fail.

- Blocked by #241
- Must first add the non-conforming projections by merging branch `devel-conga-non-conf`


FrederikSchnack:
Discussed in #320

PR #258, opened 2022-12-05
Fix span in assembly of FemField over boundary

e-moral-sanchez:


psydac/api/grid.py
```diff
@@ -202,10 +202,10 @@ def __init__( self, V, nderiv , trial=False, grid=None):
                 local_span = find_span(space.knots, space.degree, points[0, 0])
                 boundary_basis = basis_funs_all_ders(space.knots, space.degree,
                                                      points[0, 0], local_span, nderiv, space.basis)
-                if grid.ext is not None:
-                    self._basis[i][axis] = np.transpose(boundary_basis)[None, :, :, None]
-                    index = 0 if grid.ext == -1 else -1
```
e-moral-sanchez: Now if `grid.ext is None` then `index = -1`. Is this correct?

saidctb: `grid.ext` should not be `None` if `grid.axis` is not `None`, I don't remember why we got that error before

e-moral-sanchez: It was because we were using `ext` and not `grid.ext`, which at some point was set to `None`. 
I tested this with a Poisson equation prescribing a `FemField` on the right boundary and it works.

saidctb: perfect

saidctb (approved):

yguclu (approved):

PR #261, opened 2022-12-07
New Linear Operators

jowezarek:
Aim:

- Unify implementation of linear operators in Psydac
- Write container classes (SumLinearOperator, ComposedLinearOperator, ...)
- Reduce the amount of unnecessarily created temporary vectors

Main changes:

- basic.py:

  previously contained **VectorSpace**, **Vector**, **LinearOperator**, _LinearSolver_, _Matrix_

  now contains **VectorSpace**, **Vector**, **LinearOperator**,  _LinearSolver_,  **and additionally**
  ZeroOperator, IdentityOperator, ScaledLinearOperator, SumLinearOperator, ComposedLinearOperator, PowerLinearOperator, InverseLinearOperator

   _LinearSolver_ is yet to be removed. The additional value of ZeroOperator and IdentityOperator has to be evaluated.
   **Note:** Matrix has been removed.

   The remaining classes can be used in an intuitive way without having to address them explicitly: Let A, B, C, ... be linear operators of compatible domains and codomains. Then:

   `Z = A @ B` is a ComposedLinearOperator
   `Z = A @ B + C` is a SumLinearOperator
   `Z = 2 * (A @ B + C)` is a ScaledLinearOperator
   `Z = ( 2 * (A @ B + C) ) ** 3` is a PowerLinearOperator

   **Note:** Container classes are not **unique**. `2 * (A+B)` is a ScaledLinearOperator, `2*A + 2*B` is a SumLinearOperator.

   Useful exceptions have been implemented: 

   - Any LinearOperator subclass providing own algebra rules 

      The sum of two StencilMatrices is a StencilMatrix. 
      Multiplying a StencilMatrix by a scalar yields a StencilMatrix
   - Raising to the power of 1, multiplying by 1, adding a ZeroOperator, etc. does not change an object.
   - Taking the inverse of an inverse does not change an object.

- basic.InverseLinearOperator and solvers.py

   basic.InverseLinearOperator is an abstract class. Currently, five subclasses contained in solvers.py inherit from it:
   - ConjugateGradient, PConjugateGradient, BiConjugateGradient, MinRes, LSMR
   Objects of these classes can be used just like any other LinearOperator objects. When applying them to a vector, the respective iterative algorithm is employed.

   previously: 
   ```
   solver_settings = {"x0": ..., "tol": ..., "maxiter": ..., ...}
   x, info = cg( A, b, **solver_settings)
   ```
   now:
   ```
   solver_settings = {"x0": ..., "tol": ..., "maxiter": ..., ...}
   A_inv = inverse(A, 'cg', **solver_settings)
   x = A_inv @ b
   info = A_inv.get_info()
   ```
   The solver settings are stored in the ConjugateGradient object and can be changed if needed.

   This implementations yields several advantages:
   - Preallocating storage in A_inv allows to avoid the creation of any unnecessary temporary vectors - during each iteration of the iterative method and for each time step!
   - Inverse operators can now be included more naturally in the code. They additionally allow the use of `out = ...` to further decrease the amount of temporary vectors. Adapted from api/tests/test_api_feec_3d.py, compare
old implementation
        ```
        def step_ampere(b):
            y1 = M2.dot(b)
            y2 = CURL_T.dot(y1)
            return cg(M1, y2, tol=1e-12)[0]
        
        for  ts in range(niter):
            e_new = e + dt * step_ampere(b_new)
        ```
        and the new implementation
        ```
        step_ampere  = dt * ( inverse(M1, 'cg', tol=1e-12) @ CURL_T @ M2 )
        
        for  ts in range(niter):
            e += step_ampere.dot(b, out=de)
        ```


saidctb (approved):

psydac/linalg/basic.py
```diff
@@ -881,8 +881,8 @@ def transpose(self):
     def inverse(self, solver, **kwargs):
         return self._linop
 
-    @staticmethod
-    def jacobi(A, b):#, out=None):
+    #@staticmethod
+    def jacobi(A, b, out=None):
```
yguclu, line 885: If your remove the `@staticmethod` decorator you also need to pass `self` as the first function argument.

psydac/feec/multipatch/fem_linear_operators.py
```diff
@@ -59,6 +59,12 @@ def T(self):
     def dtype( self ):
         return self.domain.dtype
 
+    def toarray(self):
+        raise NotImplementedError('toarray() is not defined for FEMLinearOperators.')
+
+    def tosparse(self):
+        raise NotImplementedError('tosparse() is not defined for FEMLinearOperators.')
```
yguclu, line 66: Could we not implement these functions as follows?
```suggestion
    def toarray(self):
        self.matrix.toarray()

    def tosparse(self):
        self.matrix.tosparse()
```
CC: @campospinto

psydac/linalg/null.py
```diff
@@ -34,6 +34,12 @@ def codomain( self ):
     def dtype( self ):
         return self.domain.dtype
 
+    def toarray(self):
+        raise NotImplementedError('toarray() is not defined for this outdated class.')
+
+    def tosparse(self):
+        raise NotImplementedError('tosparse() is not defined for this outdated class.')
+
```
yguclu, line 42: What do you mean with "outdated class"? Is this not used anymore?

yguclu (changes requested):
I have not read all of your changes yet, but it looks really good so far! I've left a few comments

psydac/linalg/solvers.py
```diff
@@ -559,8 +571,10 @@ def solve(self, b, out=None):
             #-----------------------
             # MATRIX-VECTOR PRODUCTS
             #-----------------------
-            v  = A.dot(p , out=v) # overwriting v, then saving in v. Necessary?
-            vs = At.dot(ps, out=vs) # same story
```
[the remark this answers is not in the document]
jowezarek: I see, thank you!

psydac/linalg/identity.py
```diff
@@ -31,6 +31,12 @@ def codomain( self ):
     def dtype( self ):
         return self.domain.dtype
 
+    def toarray(self):
+        raise NotImplementedError('toarray() is not defined for this outdated class.')
+
+    def tosparse(self):
+        raise NotImplementedError('tosparse() is not defined for this outdated class.')
```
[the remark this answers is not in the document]
jowezarek: It is still in use, but eventually I'd like to replace it with my own IdentityOperator class written in /basic. That is why I didn't bother writing these functions. I added implementing these changes to my ToDo list.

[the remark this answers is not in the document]
yguclu: And where is this class used at the moment?

psydac/feec/multipatch/fem_linear_operators.py
jowezarek: True, that should work, will update before the next push!

psydac/linalg/identity.py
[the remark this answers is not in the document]
jowezarek: As the three classes defined in _identity.py_ appear only in _identity.py_ itself, the corresponding test file _test_null_identity.py_ and _derivatives.py_, I tried the naive approach of commenting out the first two files completely and replacing the **one** appearance of _IdentityStencilMatrix_ in _derivatives.py_ with my own _IdentityOperator_.
This unfortunately does not work, see

```
class DirectionalDerivativeOperator(Matrix):
   # ...
   def tokronstencil(self):
   
      # ...
      
      def make_id(i):
         # ...
         return IdentityStencilMatrix(StencilVectorSpace(cart))
         
      mats = [M if i == self._diffdir else make_id(i) for i in range(self._domain.ndim)]
      return KroneckerStencilMatrix(self._domain, self._codomain, *mats)
```

After replacing _IdentityStencilMatrix_ with _IdentityOperator_, the problem arises when using tostencil() on a _KroneckerStencilMatrix_ object created like this but also in various other parts of the code where differentiation matrices are being used.

I undo my changes for now until we come up with a workaround.

psydac/linalg/null.py
jowezarek: Unlike identity.py, null.py is not being used at all.

yguclu: If it's not used I suggest to remove it altogether. I can take care of it

yguclu: Actually, the files `null.py` and `identity.py` were created by @davschneller 1.5 years ago, so we should probably remove both at the same time.

yguclu: I suggest to leave this to a separate PR.

psydac/linalg/identity.py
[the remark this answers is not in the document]
yguclu: I agree. Let's postpone this to a future PR.

psydac/api/tests/test_api_feec_1d.py
```diff
@@ -221,21 +211,34 @@ class CollelaMapping1D(Mapping):
...
+        # Energy of numerical solution
+        def discrete_energies(self, e, b):
+            """ Compute electric & magnetic energies of numerical solution.
+            """
+            # Why does one need "self._tmp0 = "?
+            self._tmp0 = self._M0.dot(e, out=self._tmp0)
+            self._tmp1 = self._M1.dot(b, out=self._tmp1)
+            We = 0.5 * self._tmp0.dot(e)
+            Wb = 0.5 * self._tmp1.dot(b)
+            return (We, Wb)
```
yguclu, line 241: Q:
Why does one need "self._tmp0 = "?

A:
When an object of class `Diagnostics` is created, we have that `self._tmp0 = None`. Hence the first call to `discrete_energies` will allocate a vector by executing the line
```python
self._tmp0 = self._M0.dot(e, out=None)
```
where it is clearly fundamental to assign the result to `self._tmp0`. In the successive calls to `discrete_energies` the assignment is superfluous, but should not cause any noticeable overhead.

yguclu: Alternatively one could allocate the temporary arrays `self._tmp0` and `self._tmp1` in the `__init__` constructor.

psydac/linalg/tests/test_linalg.py
```diff
@@ -0,0 +1,959 @@
...
+
+def is_pos_def(x):
+    assert np.all(np.linalg.eigvals(x) > 0)
+
+def l2_norm_diff(x,y):
+    l = len(x)
+    assert l == len(y)
+    vec = [(x[i]-y[i])**2 for i in range(l)]
+    norm = np.sqrt(sum(vec))
+    return norm
```
yguclu, line 26: I think that you should use `numpy.linalg.norm(x - y, ord=2)`, which is surely much faster than this function.

psydac/linalg/tests/test_linalg.py
```diff
@@ -0,0 +1,959 @@
...
+import numpy as np
+
+from psydac.linalg.block import BlockLinearOperator, BlockVector, BlockVectorSpace
+from psydac.linalg.basic import ZeroOperator, IdentityOperator, ComposedLinearOperator, SumLinearOperator, PowerLinearOperator, InverseLinearOperator, ScaledLinearOperator
+from psydac.linalg.stencil import StencilVectorSpace, StencilVector, StencilMatrix
+from psydac.linalg.solvers import BiConjugateGradient, ConjugateGradient, PConjugateGradient, inverse
+from psydac.ddm.cart         import DomainDecomposition, CartDecomposition
+#===============================================================================
+
+# 11.12.22: Check for redundancy(?) in pytests 1,2 and 4, document 1,2
```
yguclu, line 11: Have you checked?

psydac/linalg/tests/test_linalg.py
```diff
@@ -0,0 +1,959 @@
...
+
+#===============================================================================
+#@pytest.mark.parametrize( 'n1', n1array)
+#@pytest.mark.parametrize( 'n2', n2array)
+#@pytest.mark.parametrize( 'p1', p1array)
+#@pytest.mark.parametrize( 'p2', p2array)
+
+# 'cg' inverse requires a positive definite matrix.
+# I did not yet come up with a way to create positive definite matrices for arbitrary n1, n2, p1, p2, P1, P2
+# Thus until changed: Only one test with a simple 4x4 positive definite matrix
```
yguclu, line 788: It is actually quite easy: a symmetric strictly diagonally dominant real matrix with positive diagonal entries is positive definite! So it has to be

- real-valued
- symmetric
- with positive entries in the (main) diagonal
- strictly diagonally dominant

yguclu (changes requested):

psydac/linalg/block.py
```diff
@@ -782,47 +767,53 @@ def __rmul__(self, a):
...
+            if len(mat._blocks) != len(self._blocks):
+                mat.set_backend(self._backend)
+            elif self._backend is not None:
+                mat._func = self._func
+                mat._args = self._args
+                mat._blocks_as_args = [mat._blocks[key]._data for key in self._blocks]
+                mat._backend = self._backend
+            return mat
+        else:
+            return LinearOperator.__add__(self, M)
```
yguclu, line 791: In order to minimize the number of lines changed in this file, I suggest to revert the body of this function back to its original content, and simply add two lines at the beginning of it:
```python
def __add__(self, M):
    if not isinstance(M, BlockLinearOperator):
        return LinearOperator.__add__(self, M)

    # Original code here
```
The same comment applies to the functions `__sub__`, `__iadd__` and `__isub__` below.

psydac/linalg/block.py
```diff
@@ -1441,4 +1423,4 @@ def __setitem__( self, key, value ):
         # Check domain of rhs
         assert value.space is self.space[key]
 
-        self._blocks[key] = value
+        self._blocks[key] = value
```
yguclu, line 1426: No newline at end of file

yguclu (changes requested):

psydac/feec/derivatives.py
```diff
@@ -365,7 +367,7 @@ def __sub__(self, m):
     # (except with an explicit neutral element, if there was one)
 
     def __imul__(self, a):
-        if isinstance(a, IdentityMatrix): return
+        if isinstance(a, IdentityOperator): return
```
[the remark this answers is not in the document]
jowezarek: No, this modification doesn't work, it's certainly not being tested. I forgot to undo the change. Will do with my next commit!

psydac/linalg/tests/test_linalg.py
```diff
@@ -0,0 +1,959 @@
...
+    H5 = ZV @ I
+    H = H1 @ ( H2 + H3 - H4 + H5 ).T
+    assert l2_norm_diff((H@v).toarray(), v.toarray()) < 10*tol
+
+#===============================================================================
+# SCRIPT FUNCTIONALITY
+#===============================================================================
+if __name__ == "__main__":
+    import sys
+    pytest.main( sys.argv )
```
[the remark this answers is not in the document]
jowezarek: Changed. Why is that required?

psydac/api/equation.py
```diff
@@ -254,12 +229,31 @@ def solve(self, **kwargs):
...
+        if inf == True:
+            X = M_inv @ rhs
             uh = FemField(self.trial_space, coeffs=X)
+            info = M_inv.get_info()
             return uh, info
-
         else:
-            X  = driver_solve(self.linear_system, **settings)
+            X = M_inv @ rhs
+            #X = pcg(M, rhs, pc='jacobi', tol=1e-8, verbose=True)
```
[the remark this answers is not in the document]
jowezarek: Yes indeed, will be removed with the next commit.

psydac/linalg/tests/basic_test.py
```diff
@@ -0,0 +1,319 @@
+# old tests here will be converted to pytests and added to the test_linalg module
```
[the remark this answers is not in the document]
jowezarek: Not at all, I removed it.

psydac/linalg/tests/test_linalg.py
jowezarek: :') this function was not meant to be seen by anybody, please forget it ever existed

psydac/linalg/tests/test_linalg.py
jowezarek: No, good catch. Did so now, will be changed with the next commit.

psydac/linalg/tests/test_linalg.py
jowezarek: That's neat, thank you! Will update this test in the near future.

psydac/linalg/basic.py
```diff
@@ -41,135 +48,976 @@ def zeros( self ):
...
+
+        return x
 
 #===============================================================================
-class Matrix( LinearOperator ):
+class Matrix(LinearOperator):
     """
     Linear operator whose coefficients can be viewed as a 2D matrix.
 
     """
```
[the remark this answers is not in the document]
jowezarek: That was the original plan. If I'm not mistaken, considering the three classes Matrix, BlockLinearOperator and BlockMatrix, at some point we decided on only removing the class BlockMatrix, making the BlockLinearOperator class a subclass only to LinearOperator.
Now that I had a look again at the almost empty Matrix class, I agree, we should remove it.
I will have to check the matrix subclasses DirectionalDerivativeOperator, StencilInterfaceMatrix, BlockLinearOperator, KroneckerStencilMatrix, StencilMatrix, DenseMatrix, LinearOperator_StencilToDense and LinearOperator_StencilToDense for eventual issues, but that should not take to long. I plan to have it removed before the next meeting.

psydac/linalg/block.py
```diff
@@ -479,6 +480,11 @@ def __init__( self, V1, V2, blocks=None ):
         self._args['n_cols'] = self._ncols
         self._func           = self._dot
         self._sync           = False
+        self._backend = None
+
+    ## 11.12.22: as several tests require BlockLO to have a tomatrix attribute
+    #def tomatrix(self):
+    #    return self
```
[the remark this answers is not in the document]
jowezarek: I think it was an old function of the old BlockLinearOperator class, transforming a BlockLinearOperator into a BlockMatrix. As this distinction does not exist anymore, the function became obsolete. I removed the three lines of code.

psydac/linalg/block.py
```diff
@@ -583,6 +632,25 @@ def remove_spurious_entries( self ):
...
+    #        blocks[i][j] = Mij.topetsc()
+
+    #    if self.n_block_cols == 1:
+    #        cart = self.domain.cart
+    #    else:
+    #        cart = self.domain.spaces[0].cart
+
+    #    petsccart = cart.topetsc()
+
+    #    return petsccart.petsc.Mat().createNest(blocks, comm=cart.comm)
```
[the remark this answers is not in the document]
jowezarek: It was part of the BlockMatrix class - See old commit "Remove BlockMatrix everywhere, delete linalg2, add out=None to .dot, … ", 01.12.2022, in block.py, starting at line 974.
I will remove these lines. The topetsc function in my branch in BlockLinearOperator and BlockVector is up to date with the devel topetsc functions.

psydac/linalg/block.py
```diff
@@ -645,101 +713,18 @@ def transform(self, operation):
...
-
-    #--------------------------------------
-    # Abstract interface
-    #--------------------------------------
-    def toarray( self, **kwargs ):
-        """ Convert to Numpy 2D array. """
-        return self.tosparse(**kwargs).toarray()
+    #def transpose(self):
+    #    blocks = {(j, i): b.transpose() for (i, j), b in self._blocks.items()}
+    #    return BlockLinearOperator(self.codomain, self.domain, blocks=blocks)
```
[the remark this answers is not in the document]
jowezarek: Yes!

psydac/api/tests/test_api_feec_1d.py
jowezarek: Thank you for clarifying!

psydac/linalg/solvers.py
```diff
@@ -0,0 +1,1369 @@
...
+    elif solver == 'lsmr':
+        obj = LSMR(A, **kwargs)
+    return obj
+
+#===============================================================================
+class ConjugateGradient(InverseLinearOperator):
+    """
+    
+
+    """
```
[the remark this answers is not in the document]
jowezarek: That was long overdue, finally did it.

psydac/linalg/tests/test_null_identity.py
```diff
@@ -102,4 +102,4 @@ def testnullself(op, run):
 
     testall(nullselfop, testnullself, run1)
     testall(nullselfop, testnullself, run2v)
-    testall(nullselfop, testnullself, run2i)
+    testall(nullselfop, testnullself, run2i)
```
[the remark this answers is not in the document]
yguclu: You have actually added a new line that contains some blank space characters, but this line does not end with a [newline character](https://en.wikipedia.org/wiki/Newline). So the problem is still there 😅

psydac/linalg/block.py
```diff
@@ -832,7 +795,10 @@ def __imul__(self, a):
 
     # ...
     def __iadd__(self, M):
-        assert isinstance(M, BlockMatrix)
+        if not isinstance(M, BlockLinearOperator):
+            return LinearOperator.__add__(self, M)
+
+        assert isinstance(M, BlockLinearOperator)
```
yguclu, line 801: I think this assert is redundant now, given the `if` statement just above.

psydac/linalg/block.py
```diff
@@ -852,7 +818,10 @@ def __iadd__(self, M):
 
     # ...
     def __isub__(self, M):
-        assert isinstance(M, BlockMatrix)
+        if not isinstance(M, BlockLinearOperator):
+            return LinearOperator.__sub__(self, M)
+
+        assert isinstance(M, BlockLinearOperator)
```
yguclu, line 824: I think this assert is redundant now, given the `if` statement just above.

yguclu (changes requested):

yguclu (approved):

PR #271, opened 2023-02-10
Free MPI channels when StencilVector is deleted

### LINKED ISSUE #263: Avoid MPI memory leaks on persistent communication
The channels for persistent MPI communications should be freed upon deletion of any object of class `StencilVector`. If we do not do so, the creation of many temporary vectors (e.g. in a CG solver) will cause a memory leak and eventually an out-of-memory error.

This can be obtained with the following magic function:

```python
def __del__(self):
    if self._requests:
        for request in self._requests:
            request.Free()
```

### LINKED ISSUE #270: Massive memory consumption with consecutive dot products
When I do multiple matrix-vector products one after the other (e.g. in a for loop) my machine eventually hangs up because of a massive memory accumulation. After running the code

```
for i in range(10000):
    x = A.dot(b)
```

for `A` being a diagonal `BlockMatrix` of shape `(1536, 1536)` and `b` some `BlockVector` I have a memory consumption of about 2.5 GB. Making the loop even longer eventually leads to a system hang up.

yguclu:
Fixes #263. Fixes #270.

### DIFF
--- a/psydac/linalg/stencil.py
+++ b/psydac/linalg/stencil.py
@@ -310,6 +310,12 @@ class StencilVector( Vector ):
         # TODO: distinguish between different directions
         self._sync  = False
 
+    def __del__(self):
+        # Release memory of persistent MPI communication channels
+        if self._requests:
+            for request in self._requests:
+                request.Free()
+
     #--------------------------------------
     # Abstract interface
     #--------------------------------------


saidctb:
Does this solve the issue https://github.com/pyccel/psydac/issues/270?

yguclu:
@saidctb It is not clear to me why the problem only happened with periodic boundary conditions.

saidctb (approved):

PR #278, opened 2023-02-22
Fix operator block stencil

vcarlier:
Fixes : linalg/block problem when the domain of a block linear operator is not a blockvectorspace in the creation of the arguments and the output for evaluation
linalg/basic : temporary vectors for the evaluation of composed operator were missing in case the last operator is also a composed operator 
api/fem : The returns of the assembly functions were not copy, creating errors when evaluating several time the same LinearForm or BilinearForm


psydac/api/fem.py
```diff
@@ -464,7 +464,10 @@ def assemble(self, *, reset=True, **kwargs):
             self._matrix.exchange_assembly_data()
 
         if self._matrix: self._matrix.ghost_regions_in_sync = False
-        return self._matrix
+        if self._matrix == None:
+            return None
+        else : 
+            return self._matrix.copy()
```
yguclu, line 470: I understand your reasoning here, but if we assemble the bilinear form only once (which is what we do all the time in our unit tests), we end up creating two matrices when we only need one. This means that the memory usage is twice what it should be.

CC: @saidctb

psydac/api/fem.py
```diff
@@ -994,7 +997,10 @@ def assemble(self, *, reset=True, **kwargs):
             self._vector.exchange_assembly_data()
 
         if self._vector: self._vector.ghost_regions_in_sync = False
-        return self._vector
+        if self._vector == None : 
+            return None
+        else : 
+            return self._vector.copy()
```
yguclu, line 1003: Same comment as above, but the extra memory usage is less important in the case of a vector.

psydac/api/tests/test_assembly.py
```diff
@@ -168,4 +181,4 @@ def test_non_symmetric_BilinearForm(backend):
     test_field_and_constant(None)
     test_multiple_fields(None)
     test_math_imports(None)
-    test_non_symmetric_BilinearForm(None)
+    test_non_symmetric_BilinearForm('python')
```
yguclu, line 184: This part of the file is only here as a convenience to developers, and it is not run by Pytest. Is it important to make this change?

psydac/linalg/tests/test_linalg.py
```diff
@@ -823,6 +833,63 @@ def test_operator_evaluation():#n1, n2, p1, p2, P1=False, P2=False):
...
+
+    assert len(Z1_1.tmp_vectors) == 2
+    assert len(Z1_2.tmp_vectors) == 2
+    assert len(Z1_3.tmp_vectors) == 2
+    assert len(Z2_1.tmp_vectors) == 3
+    assert len(Z2_2.tmp_vectors) == 3
+    assert len(Z2_3.tmp_vectors) == 3
+    assert len(Z2_4.tmp_vectors) == 3
+    assert np.all( y1_1.toarray() == y1_2.toarray() ) & np.all( y1_2.toarray() == y1_3.toarray() )
+    assert np.all( y2_1.toarray() == y2_2.toarray() ) & np.all( y2_2.toarray() == y2_3.toarray() ) & np.all( y2_3.toarray() == y2_4.toarray() )
```
yguclu, line 891: The standard way to compare Numpy arrays for exact equality is using the function [`array_equal`](https://numpy.org/doc/stable/reference/generated/numpy.array_equal.html). Here you should also avoid calling `toarray()` on the same object:
```suggestion
a1_1, a1_2, a1_3       = [v.toarray() for v in (y1_1, y1_2, y1_3)]
a2_1, a2_2, a2_3, a2_4 = [v.toarray() for v in (y2_1, y2_2, y2_3, y2_4)]
assert np.array_equal(a1_1, a1_2) and np.array_equal(a1_2, a1_3)
assert np.array_equal(a2_1, a2_2) and np.array_equal(a2_2, a2_3) and np.array_equal(a2_3, a2_4)
```

yguclu (changes requested):
Good job!! I have some comments

psydac/linalg/tests/test_linalg.py
jowezarek: Thanks for the feedback!

psydac/linalg/block.py
```diff
@@ -1255,13 +1255,13 @@ def set_backend(self, backend):
 
         if interface:
             def func(blocks, v, out, **args):
-                    vs   = [vi._interface_data[d_axis, d_ext] for vi in v.blocks] if isinstance(v, BlockVector) else v._data
-                    outs = [outi._data for outi in out.blocks] if isinstance(out, BlockVector) else out._data
+                    vs   = [vi._interface_data[d_axis, d_ext] for vi in v.blocks] if isinstance(v, BlockVector) else [v._data]
```
jowezarek, line 1258: @yguclu, @vcarlier, I have a hard time recreating the problem, could you help me out?
So far I've tried
```
# V is a StencilVectorSpace
# and S is a StencilMatrix from V -> V
U = BlockVectorSpace(V, V)

# 2 different block arrangements for both a BlockLinearOperator from U -> V as well as V -> U
dict_blocks1 = {(0,0):S, (0,1):S}
dict_blocks2 = {(0,0):S, (1,0):S}
B1 = BlockLinearOperator(U, V, blocks=dict_blocks1)
B2 = BlockLinearOperator(V, U, blocks=dict_blocks2)

# s is a StencilVector that belongs to the Space V
b = BlockVector(U, (s, s))

y1 = B1 @ b
y2 = B2 @ s
```
Both LinearOperators, although either from BlockVectorSpace/StencilVectorSpace to StencilVectorSpace/BlockVectorSpace, evaluate just the way they should.

yguclu:
Closed by PR #279

PR #281, opened 2023-03-10
Update ghost regions before assembling a form

vcarlier:
When assembling a discrete linear/bilinear form (or functional) which takes a `FemField` as parameter, make sure that the spline coefficients have up-to-date ghost regions. Failure to do so may yield wrong results.

Add a unit test showing how the error could show up. The test failed prior to the bug fix.

### DIFF
--- a/psydac/api/tests/test_assembly.py
+++ b/psydac/api/tests/test_assembly.py
@@ -2,12 +2,12 @@ import pytest
 from sympy import pi, sin, cos, tan, atan, atan2
 from sympy import exp, sinh, cosh, tanh, atanh, Tuple
 
+
 from sympde.topology import Line, Square
 from sympde.topology import ScalarFunctionSpace, VectorFunctionSpace
 from sympde.topology import element_of, Derham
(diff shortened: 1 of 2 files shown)


yguclu (approved):
@vcarlier Thanks for investigating the problem and fixing it! The PR looks good to me

psydac/api/tests/test_assembly.py
yguclu, line 1: Could you also test if norms (i.e. a `DiscreteFunctional`) have the same problem? I suspect they do.

yguclu (changes requested):

psydac/api/tests/test_assembly.py
vcarlier: The problem was there, I fixed it and add a small test where the bug was showing up

yguclu: Great, thanks!

yguclu (approved):

saidctb (approved):

PR #282, opened 2023-03-15
Add Hvec space to de Rham sequence, if requested

vcarlier:
The "Hvec" space is the (H^1)^d space taken as the product of d identical Vh^0 spaces.

List of modifications
---------------------
- Add the "Hvec" space to `DiscreteDerham` objects, if requested to the `discretize_derham` function through the optional argument `get_H1vec_space` (default is `False`).
- Add a geometric projection for this space, which is a component-wise interpolation combined with the pull_back for vector fields.
- Add documentation to the `discretize_derham` function and the `DiscreteDerham` class.
- Add tests for the subclasses of `GlobalProjector`.


psydac/api/feec.py
```diff
@@ -83,6 +92,11 @@ def V2(self):
     def V3(self):
         return self._spaces[3]
 
+    @property
+    def Vvec(self):
+        assert self.has_vec
```
spossann, line 97: docstring missing. What space is this?

psydac/feec/basis_projectors.py
```diff
@@ -0,0 +1,717 @@
...
+from psydac.fem.basic import FemField
+from psydac.utilities.utils import roll_edges
+
+from copy import deepcopy
+
+
+
+class BasisProjectionOperator(LinearOperator):
+    """
+    Class for "basis projection operators" PI_ijk(fun Lambda_mno) in the general form BP * P * DOF * EV^T * BV^T.
```
spossann, line 22: `BP * P * DOF * EV^T * BV^T` is not really applicable yet in Psydac, is it?

psydac/feec/global_projectors.py
```diff
@@ -673,6 +735,24 @@ def evaluate_dofs_2d_2form(
                     F[i1, i2] += quad_w1[i1, g1] * quad_w2[i2, g2] * \
                             f(quad_x1[i1, g1], quad_x2[i2, g2])
 
+#------------------------------------------------------------------------------    
+def evaluate_dofs_2d_vec(
```
spossann, line 739: Why not call the `dof_kernels` here?

spossann (approved):

psydac/api/discretization.py
```diff
@@ -81,18 +81,42 @@ def change_dtype(V, dtype):
...
+    Parameters
+    ----------
+
+    derham : sympde.topology.space.Derham
+        The symbolic Derham sequence
+
+    domain_h   : Geometry
+        Discrete domain where the spaces will be discretized
+        
+    get_vec : Bool
```
campospinto, line 98: to be consistent with the notation used elsewhere, I would call this argument `get_H1vec_space`

psydac/feec/pull_push.py
```diff
@@ -64,6 +65,36 @@ def f_logical(eta1):
 #==============================================================================
 # 2D PULL-BACKS
 #==============================================================================
+def pull_2d_v(f, F):
```
campospinto, line 68: I would prefer a more explicit name like `pull_2d_vec`

psydac/feec/pull_push.py
```diff
@@ -161,29 +192,47 @@ def f_logical(eta1, eta2):
 # TODO [YG 05.10.2022]:
 # Remove? But it makes sense to return a vector-valued function...
 
-def pull_3d_v(funcs_ini, mapping):
+def pull_3d_v(f, F):
```
campospinto, line 195: I would prefer a more explicit name like `pull_3d_vec`

psydac/feec/pull_push.py
```diff
@@ -64,6 +65,36 @@ def f_logical(eta1):
 #==============================================================================
 # 2D PULL-BACKS
 #==============================================================================
+def pull_2d_v(f, F):
+    #We should check if the metric terms are really the good ones!
```
campospinto, line 69: So, did you check ?

psydac/feec/global_projectors.py
```diff
@@ -147,6 +149,9 @@ def __init__(self, space, nquads = None):
                     if quad_x[j] is None:
                         u, w = uw[j]
                         global_quad_x, global_quad_w = quadrature_grid(V.histopolation_grid, u, w)
+                        #"roll" back points to the interval to ensure that the quadrature points are
+                        #in the domain. Probably only usefull on periodic cases
+                        roll_edges(V.domain, global_quad_x)
```
campospinto, line 154: is this related to the H1vec space, or fixing a pre-existing bug ?

psydac/feec/global_projectors.py
```diff
@@ -564,6 +569,63 @@ def __call__(self, fun):
...
+    """
+    Projector from H1^3 = H1 x H1 x H1 to a conforming finite element space, i.e.
+    a finite dimensional subspace of H1^3, constructed with tensor-product
+    B-splines in 2 or 3 dimensions.
+    This is a global projector constructed over a tensor-product grid in the
+    logical domain. The vertices of this grid are obtained as the tensor
+    product of the 1D splines' Greville points along each direction.
+    Parameters
+    ----------
+    H1vec : ProductFemSpace
```
campospinto, line 582: isn't the argument of the GlobalProjector constructor called `space` ?

psydac/api/feec.py
```diff
@@ -4,25 +4,47 @@
...
+    
+    Parameters
+    ----------
+
+    mapping : Mapping
+        The mapping from the logical space to the physical space of the discrete De Rham.
+
+    get_vec : Bool
+        True to also get the "Hvec" space discretizing (H1)^n vector fields
+
```
yguclu, line 28: I find it strange that this flag should be called `get_vec`, given that the H1-vector space is passed as an input to the constructor of this class.

yguclu (changes requested):
Looks good, but I have a few comments.

vcarlier (commented):
Am I reviewing when trying to reply?

psydac/api/ast/utilities.py
```diff
@@ -386,23 +386,23 @@ def compute_atoms_expr_mapping(atomic_exprs, indices_quad,
...
+        # ...
 
-        # ...
+        # ...
         val_name = SymbolicExpr(atom).name + '_values'
         val      = IndexedBase(val_name)[indices_quad]
         expr     = IndexedBase(element_name)[idxs] * test_fun
         updates += [AugAssign(val, '+', expr)]
-        # ...
+        # ...
```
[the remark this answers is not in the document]
vcarlier: This is probably a bug

psydac/api/ast/utilities.py
```diff
@@ -331,14 +331,14 @@ def compute_atoms_expr_field(atomic_exprs, indices_quad,
...
 
-        # ...
+        # ...
         args     = [IndexedBase(field_name)[idxs], test_fun]
         val_name = SymbolicExpr(atom).name + '_values'
         val      = IndexedBase(val_name)[indices_quad]
         updates += [AugAssign(val,'+',Mul(*args))]
-        # ...
+        # ...
```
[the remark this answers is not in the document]
vcarlier: This is probably a git bug

psydac/api/feec.py
```diff
@@ -4,25 +4,46 @@
...
     """ Represent the discrete De Rham sequence.
+    Should be initialized via discretize_derham function in api.discretization.py
+    
+    Parameters
+    ----------
+
+    mapping : Mapping
+        The mapping from the logical space to the physical space of the discrete De Rham.
+        
+    *spaces : list of
```
yguclu, line 28: List of `FemSpace` objects?

psydac/api/feec.py
yguclu: Yes, I think we should add docstrings for all of these properties. We should have done it before, could you please write one-line docstrings? Please let me know if you have any questions.

psydac/feec/global_projectors.py
```diff
@@ -791,3 +878,32 @@ def evaluate_dofs_3d_3form(
...
+#------------------------------------------------------------------------------
+def evaluate_dofs_3d_vec(
+        intp_x1, intp_x2, intp_x3, # interpolation points
+        F1, F2, F3,                # array of degrees of freedom (intent out)
+        f1, f2, f3,                # input scalar function (callable)
+        ):
+    
+    # evaluate input functions at interpolation points (make sure that points are in [0, 1])
+    
+
```
yguclu, line 891: Usually two consecutive empty lines are only used between separate functions or classes. These can be removed here completely.

yguclu (changes requested):

psydac/api/feec.py
yguclu: Checking the last space is an elegant solution. :+1:

psydac/feec/global_projectors.py
yguclu: Can you please further explain what is happening here? `roll_edges` is only relevant in the periodic case, why do you think it is necessary now?

psydac/feec/pull_push.py
vcarlier: h1vec might be the best choice cause maybe someday someone will need an l2vec space

psydac/feec/global_projectors.py
yguclu: @vcarlier Have you answered my question somewhere else?

yguclu (changes requested):

psydac/feec/global_projectors.py
vcarlier: Not yet I was doing tests to be sure that you need it and this is indeed the case : 
In the non periodic case roll edges does nothing so you don't care applying it all the time.
In the periodic case, if you don't use it you might have quadrature point outside of the domain, which might cause problem when you want to define your function by it's value on the domain but it's not periodic on R (for example `x*(x-1)` define a continuous function when restricted to (0,1) with periodic BC, but it's not a periodic function on R so if you don't roll back your going to evaluate it outside of your domain and have wrong values, see `test_derham_projector_2d_hdiv_2` in `test_global_projectors.py` )

psydac/feec/tests/test_global_projectors.py
```diff
@@ -83,6 +87,146 @@ def test_L2_projector_1d(domain, ncells, degree, periodic, nquads):
...
+    # Test if max-norm of error is <= TOL
+    maxnorm_error = abs(vals_u0 - vals_f).max()
+    print(ncells, maxnorm_error)
+    maxnorm_error = abs(vals_u1_1 - vals_f).max()
+    print(ncells, maxnorm_error)
+    maxnorm_error = abs(vals_u2 - vals_f).max()
+    print(ncells, maxnorm_error)
+    maxnorm_error = abs(vals_ux_1 - vals_f).max()
+    print(ncells, maxnorm_error)
+#    assert maxnorm_error <= 1e-14
```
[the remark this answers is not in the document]
vcarlier: I can just assert that the error is relatively small with a ad hoc defined threshold, if a future PR break the projections then this tests will fail.

yguclu (approved):
Good job!

PR #283, opened 2023-03-15
Push-forward function optimization

### LINKED ISSUE #178: Avoid useless operations in push-forward functions
Some of our push-forward functions recompute the same field multiple times. For example
```python
def push_2d_hdiv(a1, a2, xi1, xi2, mapping):


    mapping    = mapping.get_callable_mapping()
    J          = mapping._jacobian
    metric_det = mapping._metric_det


    J_value    = J(xi1, xi2)
    det_value  = metric_det(xi1, xi2)**0.5


    value1 = ( J_value[0,0]*a1(xi1, xi2) +
               J_value[0,1]*a2(xi1, xi2)) / det_value


    value2 = ( J_value[1,0]*a1(xi1, xi2) +
               J_value[1,1]*a2(xi1, xi2)) / det_value


    return value1, value2
```
Computes `a1(xi1, xi2)` and `a2(xi1, xi2)` twice. We should avoid this.

tomcaruso:
Fix #178 : Each push-forward functions now compute the field only once

### DIFF
--- a/psydac/feec/pull_push.py
+++ b/psydac/feec/pull_push.py
@@ -41,8 +41,7 @@ def pull_1d_h1(func_ini, mapping):
     def fun(xi1):
         x = f1(xi1)
 
-        value = func_ini(x)
-        return value
+        return func_ini(x)
 
     return fun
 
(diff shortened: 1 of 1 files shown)


yguclu (approved):
Well done!

PR #286, opened 2023-03-23
Do not decompose domain along user-specified dimensions

spossann:
Add new flag `mpi_dims_mask` to `discretize_domain`, which is successively passed to:
- `ddm.cart.DomainDecomposition`
- `ddm.partition.compute_dims`
- `ddm.partition.compute_dims_general`

Function `compute_dims_general` is the ones responsible to create the Cartesian grid of MPI processes. It receives `mpi_dims_mask` which is either `None` or a list of `bool`. In the latter case each element is set to `True` if the dimension is to be used in the domain decomposition (=default for each dimension). If `mpi_dim_mask[i] == False`, the i-th dimension is not decomposed.

Two unit tests for 2d and 3d have been added.

### DIFF
--- a/psydac/ddm/partition.py
+++ b/psydac/ddm/partition.py
@@ -139,9 +139,11 @@ def compute_dims( nnodes, gridsizes, min_blocksizes=None, mpi=None, try_uniform=
 def compute_dims_general( mpi_size, npts, dims_mask=None ):
 
     if dims_mask is None:
-        dims_mask = [True]*len( npts )
+        dims_mask = [True] * len(npts)
     else:
-        assert any(dims_mask), 'dims_mask must contain at least one True value.'
+        assert len(dims_mask) == len(npts), "dims_mask must have one entry for each dimension."
+        assert all(isinstance(m, bool) for m in dims_mask), "dims_mask must only contain True/False values."
+        assert any(dims_mask), "dims_mask must contain at least one True value."
 
     nprocs = [1]*len( npts )
     
--- a/psydac/ddm/partition.py
+++ b/psydac/ddm/partition.py
@@ -66,7 +66,7 @@ def partition_procs_per_patch(npts, size):
     return sizes, ranges
 
 #==============================================================================
-def compute_dims( nnodes, gridsizes, min_blocksizes=None, mpi=None, try_uniform=False, dims_mask=None ):
+def compute_dims( nnodes, gridsizes, min_blocksizes=None, mpi=None, try_uniform=False, mpi_dims_mask=None ):
     """
     With the aim of distributing a multi-dimensional array on a Cartesian
     topology, compute the number of processes along each dimension.
@@ -88,7 +88,7 @@ def compute_dims( nnodes, gridsizes, min_blocksizes=None, mpi=None, try_uniform=
     try_uniform: bool
         try to decompose the array uniformly.
         
-    dims_mask: list of bool
+    mpi_dims_mask: list of bool
         True if the dimension is to be used in the domain decomposition (=default for each dimension). 
         If dim_mask[i]=False, the domain decomposition yields blocksizes[i]=gridsizes[i] along the i-th dimension.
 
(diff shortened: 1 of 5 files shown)


psydac/ddm/partition.py
```diff
@@ -132,10 +136,21 @@ def compute_dims( nnodes, gridsizes, min_blocksizes=None, mpi=None, try_uniform=
     return dims, blocksizes
 
 #==============================================================================
-def compute_dims_general( mpi_size, npts ):
+def compute_dims_general( mpi_size, npts, dims_mask=None ):
+
+    if dims_mask is None:
+        dims_mask = [True]*len( npts )
+    else:
+        assert any(dims_mask), 'dims_mask must contain at least one True value.'
```
yguclu, line 144: I suggest a couple more checks here:
```suggestion
    if dims_mask is None:
        dims_mask = [True] * len(npts)
    else:
        assert len(dims_mask) == len(npts), "dims_mask must have one entry for each dimension."
        assert all(isinstance(m, bool) for m in dims_mask), "dims_mask must only contain True/False values."
        assert any(dims_mask), "dims_mask must contain at least one True value."
```

psydac/ddm/tests/test_partition.py
```diff
@@ -48,6 +48,30 @@ def test_partition_1d_general( mpi_size ):
...
+def test_partition_2d_dims_mask( mpi_size, npts, mask ):
+
+    print(f'\nmpi_size={mpi_size}:')    
+    # General partition: blocks are not all identical but closer to a cube
+    dims, blocksizes = compute_dims( mpi_size, npts, [3,3] )   
+    print(f'general:'.ljust(30), dims, blocksizes) 
+    
+    # Mask dimensions
+    dims, blocksizes = compute_dims( mpi_size, npts, [3,3], dims_mask=mask )   
+    print(f'dims_mask {mask}:'.ljust(30), dims, blocksizes)
```
yguclu, line 65: Although we still have print statements in some of our tests, Pytest is not happy about it. Here you should pass the flag `verbose = False` to the new functions, and only print to terminal if this flag is set to `True`. Then you can pass `verbose = True` when you run the script manually (bottom of the file).

Or you can simply remove the print statements, of course.

yguclu (changes requested):
Thanks @spossann, this looks good to me!

Although I wonder if we could find a better name for the `dims_mask` parameter in function `discretize_domain`. The current name may suggest the user that such a dimension is not discretized, while this is not the case. Maybe `mpi_dims_mask` is more telling?

@saidctb What do you think?

psydac/ddm/partition.py
spossann: Agreed. Should I do "commit suggestion"?

psydac/ddm/tests/test_partition.py
spossann: Ok I will add `verbose=False`. What do you mean by Pytest does not like it? Can there be problems?

yguclu: By default Pytest collects the standard output and only shows it if there is an error or warning. This means that if a test fails the error message will be cluttered with the output from all the other tests (which do not fail). This is far from ideal.

psydac/ddm/partition.py
yguclu: If you like my suggestion, that is the fastest way to include it.

yguclu:
The workflow with Python 3.7 fails because of a download problem, I am running it again.

Codacy raises an error that does not make sense to me...

yguclu (approved):

yguclu:
> Can we rerun the Codacy test? The error came with the commit [4c4385f](https://github.com/pyccel/psydac/commit/4c4385fe07b4944373a1e18a871fc8390b0ae491), which was however resolved by the later commits.

I have rerun them 10 times already, but Codacy only processes the files that have been changed in the last commit. So it keeps raising the same error. I deactivated the Codacy test temporarily, so we can merge this.

PR #288, opened 2023-04-13
Update PSYDAC examples

jowezarek:
I updated the files in the psydac/examples folder such that they are compatible with the latest changes to the library.

Status 02.06.:

- [x] Notebook: poisson_non_periodic [fixed periodicity and changed .yml to .yaml]
- [x] maxwell_2d_multi_patch
- [x] poisson_1d
- [x] poisson_2d_mapping [excluding case with `-t czarny`. See comment below]
- [x] ---> visualize_matrices
- [x] ---> test_matrices
- [x] poisson_2d_multi_patch
- [x] poisson_3d_multi_patch
- [x] poisson_3d_target_torus [excluding case with `-l`. See comment below]
- [x] sample_multi_patch_parallel
- [x] test_cg_perf

### -t czarny
   Only tried without `-s -c`. NaN all the way, expectedly took forever to run. Funny looking >1000 lines long output consisting of one huge function description(polynomials, exponentials, cosine, sine).

## sample_multi_patch_parallel & test_cg_perf
Comments explaining the expected output are needed.

## poisson_3d_target_torus
Does not run. Long error message consisiting mostly of "integer constant is too large for its type". The methods `transpose_3d_wrapper` and `type_check` are also being mentioned in the error message.
Note: Adding `-l`, the script appears to loop forever, being stuck at "Start creating logical expression". Without `-l`, it read "Start creating physical expression" and then after some time "Start discretization".


examples/poisson_2d_mapping.py
```diff
@@ -759,16 +765,17 @@ def main( *, test_case, ncells, degree, use_spline_mapping, c1_correction, distr
         if use_spline_mapping:
             geometry = Geometry(filename='geo.h5', comm=MPI.COMM_SELF)
             map_discrete = [*geometry.mappings.values()].pop()
-            V = map_discrete.space
+            Vnew = map_discrete.space
```
jowezarek, line 768: @yguclu, while I am quite certain that it is necessary to rename this space, maybe you want to check for yourself:

In line 640 of this file it reads `M, S = assemble_matrices(V, mapping, kernel)`.
In particular, M is an element of `StencilMatrix(domain=V.vector_space, codomain=V.vector_space)`.

Unless changed, `V` is being overwritten here.

Using ipython, I think that the intended order of usage was meant to be poisson_2d_mapping -> visualize_matrices -> test_mass_matrix.

In lines 21-24 of test_mass_matrix it reads 
```
rhs = assemble_rhs(V, mapping, f1)

M_inv = inverse(M, 'cg', tol=1e-10, maxiter=100, verbose=True
sol = M_inv @ rhs
```

As `V` would have been overwritten, the codomain of `M` and the space of `rhs` didn't coincide as requested by assertions. This is why I renamed `V` to `Vnew` in this file.
I am a little sceptical, as running poisson_2d_mapping.py -t target -n 10 20 -s -c, as suggested by the visualize_matrices docstring, yields a quite bad result.

examples/poisson_3d_target_torus.py
```diff
@@ -159,9 +159,9 @@ def run_model(ncells, degree, comm=None, is_logical=False):
 
     # Create discrete spline space
     if is_logical:
-        Vh = discretize(V, Omega_log_h, degree=degree, periodic=[False, True, True])
+        Vh = discretize(V, Omega_log_h, degree=degree)
```
jowezarek, line 162: @yguclu here I did not know where to put the periodicity information correctly, so I removed it in both cases. Unless already included somewhere else, this change will yield wrong results.

examples/test_mass_matrix.py
```diff
@@ -1,21 +1,37 @@
...
+
+In [1]: run poisson_2d_mapping.py -t target -n 10 20 -s -c
+
+In [2]: run -i test_mass_matrix.py
+
+"""
+
 from psydac.linalg.solvers      import inverse
 
+globals().update(namespace)
```
jowezarek, line 15: @yguclu If this file is to be called explicitely after visualize_matrices, then the line `globals().update(namespace)` is redundant.

examples/test_mass_matrix.py
```diff
@@ -1,21 +1,37 @@
+"""
+USAGE
+=====
+
+$ ipython
+
+In [1]: run poisson_2d_mapping.py -t target -n 10 20 -s -c
```
jowezarek, line 7: @yguclu Not certain about the intended usage of this file.

psydac/linalg/block.py
```diff
@@ -588,8 +588,8 @@ def tosparse(self, **kwargs):
         M.eliminate_zeros()
 
         # Sanity check
-        assert M.shape[0] == self.codomain.dimension
-        assert M.shape[1] == self.  domain.dimension
+        #assert M.shape[0] == self.codomain.dimension
```
jowezarek, line 591: @yguclu I stumbled across a case where the shape of a `polar.DenseMatrix`, a `polar.c1_linops.LinearOperator_StencilToDense` and a `polar.c1_linops.LinearOperator_DenseToStencil` was altered after calling `.tosparse()`. (Not the shape of the original object, but `M` and `M.tosparse()` would have a different shape)

This resulted in these sanity checks failing. Where's the mistake? Are these not proper sanity checks or should `.tosparse()` not change the shape of an object?

jowezarek:
Not yet, but I can do so soon.

examples/poisson_3d_target_torus.py
yguclu: The periodicity information should be provided to the discretized domain, i.e. when you create `Omega_log_h` (see function [discretize_domain](https://github.com/pyccel/psydac/blob/566c071fcf8f7c35afe542147d360f7acca313ff/psydac/api/discretization.py#L370-L386)).

examples/test_mass_matrix.py
yguclu: You are correct, this script was originally designed to run after  `visualize_matrices.py`. I think it's a good idea to have it independent of it, though.

examples/test_mass_matrix.py
yguclu: See next comment

psydac/linalg/block.py
yguclu: Maybe the call to `tosparse()` included the ghost regions for those classes? The dimension would be different in that case. More likely these are simply bugs in the `polar` module.

examples/poisson_2d_mapping.py
yguclu: Your reasoning seems correct to me. We should discuss about the "quite bad" results though :sweat

yguclu:
@jowezarek What is the status of this PR? There are several open conversations.

examples/poisson_1d.py
```diff
@@ -235,8 +236,11 @@ def assemble_rhs( V, f ):
     grid = np.linspace( *model.domain, num=ne+1 )
 
     # Create finite element space
-    V = SplineSpace( p, grid=grid, periodic=model.periodic )
-    V = TensorFemSpace( V )
+    space = SplineSpace(p, grid=grid, periodic=model.periodic)
+    ncells = [len(space.breaks)-1]
+    periods = [model.periodic]
+    domain_decomposition = DomainDecomposition(ncells=ncells, periods=periods)
+    V = TensorFemSpace(domain_decomposition, space)
```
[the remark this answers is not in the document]
jowezarek: 2cb3a490f5b9a18004e8d7480b890b46cd1ad537

psydac/api/discretization.py
```diff
@@ -216,7 +216,7 @@ def discretize_space(V, domain_h, *, degree=None, multiplicity=None, knots=None,
         The number of quadrature points in each direction.
 
     basis: str
-        The type of basis function can be 'b' for b-splines or 'M' for M-splines.
+        The type of basis function can be 'B' for b-splines or 'M' for M-splines.
```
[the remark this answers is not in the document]
jowezarek: 2cb3a490f5b9a18004e8d7480b890b46cd1ad537

saidctb:
In `poisson_2d_mapping` file, you need to call the method `exchange_assembly_data()` for both the matrix and the vector after finishing the assembly method, this method was added after the changes that were made in [ed19ecfcf8e6f7aba11abc2e216ff7ee1fbcd769`](https://github.com/pyccel/psydac/commit/`ed19ecfcf8e6f7aba11abc2e216ff7ee1fbcd769)

yguclu:
@jowezarek the example `poisson_3d_target_torus.py` was written by @anlavandier in order to test the output and post-processing capabilities described [here](https://github.com/pyccel/psydac/blob/devel/output.md).

If you pass the `-m` flag it actually computes a numerical solution to the Poisson equation in a toroidal domain, and writes two output files with extensions `.yml` and `.h5` respectively.

If you pass the `-p` flag the script runs a post-processing routine which reads the files above and generates a `.vtu` file which can be visualized with Paraview.

examples/poisson_2d_mapping.py
yguclu: Fixed in  4618079

examples/poisson_3d_target_torus.py
```diff
@@ -154,14 +154,16 @@ def run_model(ncells, degree, comm=None, is_logical=False):
 
     print("Start discretization", flush=True)
     # Create computational domain from topological domain
-    Omega_h = discretize(Omega, ncells=ncells, comm=comm)
-    Omega_log_h = discretize(Omega_logical, ncells=ncells, comm=comm)
+    periodic = [False, False, False]
+    periodic_log = [False, True, True]
```
yguclu, line 158: Using different periodicity condition in the two cases does not seem correct to me...

examples/notebooks/Poisson_non_periodic.ipynb
```diff
@@ -1,6 +1,7 @@
 {
  "cells": [
   {
+   "attachments": {},
```
yguclu, line 4: I don't see the value in updating this Jupiter Notebook file, with the exception of the lines that I point out below.

examples/notebooks/Poisson_non_periodic.ipynb
```diff
@@ -276,7 +282,7 @@
     "\n",
     "Pm = PostProcessManager(\n",
     "    domain=Omega,\n",
-    "    space_file=f'results_poisson/space_info_{Omega.name}.yml',\n",
+    "    space_file=f'results_poisson/space_info_{Omega.name}.yaml',\n",
```
yguclu, line 285: And maybe this line should be kept too. Was the output file name extension changed from `.yml` to `.yaml` at some point?

yguclu (changes requested):

examples/poisson_3d_target_torus.py
jowezarek: I agree. This however has already been the case before and I can't immediately tell why it has been like this.

examples/notebooks/Poisson_non_periodic.ipynb
jowezarek: Agreed. I can undo the change. These lines were part of the "hidden" output, I did not change them on purpose.

yguclu: I know. After modifying a Jupiter notebook one often has to stage the changes line by line using the `-p` flag:

```bash
git add -p filename.ipynb
```

yguclu (approved):

examples/poisson_3d_target_torus.py
yguclu: Eventually we will fix this file and make sure that the simulation works. But that it beyond the scope of this PR. Thanks for looking into this!

anlavandier: Hi, I'm a bit late to the party but I wanted to say that the reason why the periodicity is like this is that in the first list `periodic = [False, False, False]` corresponds to the periodicity in the cartesian `x,y,z` coordinates which are not periodic in anyway while the `periodic_log = [False, True, True]` coresponds to the periodicity in the parameter (or logical) space which for a torus is periodic in theta and phi.

PR #301, opened 2023-05-15
Add GMRES solver

e-moral-sanchez:
- Implement a GMRES iterative solver without the restart option
- In GMRES solver recycle work vectors between successive calls to solve()
- Use parametrized unit test of iterative solvers to reduce code duplication
- Check successive dot products in unit test to spot possible bugs
- Update documentation


psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,260 @@ def solve(self, b, out=None):
...
+        Q = []
+        beta = []
+        sn = []
+        cn = []
+
+        # First values
+
+        A.dot( x , out=r)
+
+        r += b - r
```
tomcaruso, line 1886: Did the previous dot product `A.dot( x , out=r)` is not useless since at the end we only have `b` store in `r` ?

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,260 @@ def solve(self, b, out=None):
...
+        h = self._H[:k+2, k]
+
+        p = self._tmps["p"]
+        self._A.dot( Q[k] , out=p) # Krylov vector
+
+        for i in range(k + 1): # Modified Gram-Schmidt, keeping Hessenberg matrix
+            h[i] = p.dot( Q[i] )
+            p -= h[i] * Q[i]
+        
+        h[k+1] = p.dot(p) ** 0.5
```
tomcaruso, line 1965: We need to do `(p.dot(p).real)**0,5` to avoid to raise warnings

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,260 @@ def solve(self, b, out=None):
...
+        for i in range(k):
+            temp = cn[i] * h[i] + sn[i] * h[i+1]
+            h[i+1] = - sn[i] * h[i] + cn[i] * h[i+1]
+            h[i] = temp
+        
+        mod = (h[k]**2 + h[k+1]**2)**0.5
+        cn.append( h[k] / mod )
+        sn.append( h[k+1] / mod )
+
+        h[k] = cn[k] * h[k] + sn[k] * h[k+1]
```
tomcaruso, line 1984: Should optimize the code with only internal operation:
`h[k] *= cn[k],
h[k+1] *= sn[k],
h[k] += h[k+1]
`

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,260 @@ def solve(self, b, out=None):
...
+        Q.append(p.copy())
+
+
+    def apply_givens_rotation(self, k, sn, cn):
+        # Apply Givens rotation to last column of H
+        h = self._H[:k+2, k]
+
+        for i in range(k):
+            temp = cn[i] * h[i] + sn[i] * h[i+1]
+            h[i+1] = - sn[i] * h[i] + cn[i] * h[i+1]
```
tomcaruso, line 1977: Should optimize the code with only internal operation:
h[i+1] *= cn[i], h[i] *= sn[i], h[i+1] -= h[i] 

The same should be done for temp

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,257 @@ def solve(self, b, out=None):
...
+
+    verbose : bool
+        If True, L2-norm of residual r is printed at each iteration.
+
+    References
+    ----------
+    [1] Y. Saad and M.H. Schultz, "GMRES: A generalized minimal residual algorithm for solving nonsymmetric linear systems", SIAM J. Sci. Stat. Comput., 7:856–869, 1986.
+
+    """
+    def __init__(self, A, *, x0=None, tol=1e-6, maxiter=1000, verbose=False):
```
yguclu, line 1765: Since we have no restart, the amount of memory grows with the number of iterations. To avoid filling up the memory we should use a low default value for the number of iterations.

psydac/linalg/solvers.py
tomcaruso: Also we should optimize the `r += b - r` line by only doing internal operation

e-moral-sanchez: This was an error. The first residual vector must be `r = b - A.dot(x)`. 
An extra temporary is necessary to save the content of `A.dot(x)` and it cannot be saved in `b`, because that would rewrite it.

tomcaruso (approved):

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,261 @@ def solve(self, b, out=None):
...
+        Results
+        -------
+        x : psydac.linalg.basic.Vector
+            Converged solution.
+
+        info : dict
+            Dictionary containing convergence information:
+            - 'niter'    = (int) number of iterations
+            - 'success'  = (boolean) whether convergence criteria have been met
+            - 'res_norm' = (float) 2-norm of residual vector r = A*x - b.
```
yguclu, line 1845: The `solve` function updates `self._info`, but does not return the variable itself. I think this error in the docstring appears throughout the file.

If we want this function to return `info` we should change the last lines to
```python
        # Convergence information
        info = {'niter': k+1, 'success': am < tol, 'res_norm': am }

        return x, info
```
and then update the `dot` accordingly:
```python
    def dot(self, b, out=None):
        x, info = self.solve(b, out=out)
        self._info = info
        return x
```

@e-moral-sanchez @jowezarek @tomcaruso What do you think? Is it worth making this change? Or should we just update the docstrings of the `solve` methods instead?

yguclu (changes requested):

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,261 @@ def solve(self, b, out=None):
...
+    def _check_options(self, **kwargs):
+        for key, value in kwargs.items():
+
+            if key == 'x0':
+                if value is not None:
+                    assert isinstance(value, Vector), "x0 must be a Vector or None"
+                    assert value.space == self._codomain, "x0 belongs to the wrong VectorSpace"
+            elif key == 'tol':
+                assert value is not None, "tol may not be None"
+                assert value*0 == 0, "tol must be a real number"
```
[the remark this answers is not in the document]
jowezarek: Is the last check `not isinstance(x, bool)` necessary?
Anyway, this change would have to be applied to all 6 `InverseLinearOperator` subclasses.

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,261 @@ def solve(self, b, out=None):
...
+            Right-hand-side vector of linear system Ax = b. Individual entries b[i] need
+            not be accessed, but b has 'shape' attribute and provides 'copy()' and
+            'dot(p)' functions (dot(p) is the vector inner product b*p ); moreover,
+            scalar multiplication and sum operations are available.
+
+        out : psydac.linalg.basic.Vector | NoneType
+            The output vector, or None (optional).
+
+        Results
+        -------
```
jowezarek, line 1837: Can you please change all instances of `Results` to `Returns`? I already did so in PR #303, but this will minimize the amount of potential merge conflicts I suppose.

psydac/linalg/solvers.py
jowezarek: I'd vote against making this change. Not only because we'd have to change all tests (`x = A_inv @ b` -> `x, info = A_inv @ b`), but also because I wouldn't want to store convergence information by default but only when I'm interested in it.

tomcaruso: I agree with Julian. The .solve method should only return the solution and not the info. I think that it is less confusing, if we only change the docstring.

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,261 @@ def solve(self, b, out=None):
...
+        cn = []
+
+        # First values
+        A.dot( x , out=r)
+        r -= b
+
+        am = r.dot(r).real ** 0.5
+        beta.append(am)
+        Q.append(- r / am)       
+
```
jowezarek, line 1890: It's just one temporary per `.dot` call, but I'd still replace
`Q.append(- r / am)`
with
```
r *= -1/am
Q.append(r)
```

psydac/linalg/solvers.py
```diff
@@ -1725,3 +1728,261 @@ def solve(self, b, out=None):
...
+            h[i] = p.dot(Q[i])
+            Q[i].copy(out=lv)
+            lv *= h[i]
+            p -= lv
+        
+        h[k+1] = p.dot(p).real ** 0.5
+        p /= h[k+1] # Normalize vector
+
+        Q.append(p.copy())
+
```
jowezarek, line 1964: This line creates lots of "temporaries", but from what I understood during our Psydac dev meeting last week there`s no way of avoiding this, right?

psydac/linalg/solvers.py
yguclu: All right. Let's change the docstrings then!

psydac/linalg/solvers.py
yguclu: There are two ways of avoiding temporaries (assuming the `solve` method is called multiple times):
1) Add the integer `restart` parameter and pre-allocate `restart` vectors (this would need a separate PR);
2) Reuse the vectors in `Q` between successive calls to `solve`:
 ```python
if len(Q) > k:
    p.copy(out = Q[k+1])
else:
    Q.append(p.copy())
```

File: psydac/linalg/solvers.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (e-moral-sanchez) [reply psydac/linalg/solvers.py:1801]: I thought about doing:
`assert value is not None, "tol cannot be None"`
`assert value.real == value, "tol must be a real number"`
What do you think it is best?

File: psydac/linalg/solvers.py (hunk 6 above)
Comment (e-moral-sanchez) [reply psydac/linalg/solvers.py:1845]: Then where should I explain the Object `._info`?

Comment (e-moral-sanchez) [reply psydac/linalg/solvers.py:1845]: OK, I did the changes but I am not convinced, because the iterative solvers of Scipy return a `Tuple` of solution and info.

File: psydac/linalg/solvers.py (hunk 10 above)
Comment (e-moral-sanchez) [reply psydac/linalg/solvers.py:1915]: For the moment I take the second option

File: psydac/linalg/solvers.py (hunk 7 above)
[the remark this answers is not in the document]
Comment (yguclu) [reply psydac/linalg/solvers.py]: I do not see a reason for checking the `None` case. `tol` must be a real number and if not we raise an error.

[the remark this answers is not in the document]
Comment (yguclu) [reply psydac/linalg/solvers.py]: Possible utility function:
```python
def is_real(x):
    from numbers import Number
    from numpy import isrealobj
    return isinstance(x, Number) and isrealobj(x) and not isinstance(x, bool)
```
which you can use as
```python
assert is_real(value), "tol must be a real number"
```

psydac/linalg/solvers.py
yguclu: > OK, I did the changes but I am not convinced, because the iterative solvers of Scipy return a `Tuple` of solution and info.

The iterative solvers of Scipy are simple functions, so they must return `info` when called. Our solvers are `InverseLinearOperator` objects and therefore they can be inspected at a later time. I do not see why we should mimic the Scipy interface now that we have proper classes instead of simple functions.

Personally I do not see the reason to expose the function `solve` to the user, as it is not part of the public interface of the `LinearOperator` objects. I would go as far as calling it `_solve` to make it a private function.

jowezarek:
Hey @e-moral-sanchez @yguclu, I just wanted to see whether the recent changes to `Q` work as intended and I've got good and bad news.
The good news is that for the one example that I checked, the number of vectors created after the first iteration is down to 0.
The bad news is that the solver has a bug.

I am referring to the example `api/tests/test_api_feec_1d.py`. To recreate the issue, and also to see the amount of vectors created, do
   - Add `print(f"The object {repr(self)} has been created.")` to the `__init__` of BlockVector and StencilVector.
   - In api/tests/test_api_feec_1d.py, change `M0_inv = inverse(M0, 'cg', **kwargs)` to `M0_inv = inverse(M0, 'gmres', **kwargs)` in line 331.
   - Go to api/tests and run `python3 test_api_feec_1d.py 20 3 -P -T 1`

psydac/linalg/solvers.py
```diff
@@ -365,13 +363,8 @@ def solve(self, b, out=None):
...
         x : psydac.linalg.basic.Vector
-            Converged solution.
-
-        info : dict
-            Dictionary containing convergence information:
-            - 'niter'    = (int) number of iterations
-            - 'success'  = (boolean) whether convergence criteria have been met
-            - 'res_norm' = (float) 2-norm of residual vector r = A*x - b.
+            Numerical solution of the linear system. To check the convergence of the solver,
+            inspect InverseLinearOperator._info.
```
yguclu, line 367: Actually we should not read a private attribute like `_info` unless strictly needed. The `InverseLinearOperator` class has a public method which I think is called `get_info()`: we should point the user to it.

psydac/linalg/solvers.py
```diff
@@ -1718,3 +1690,263 @@ def solve(self, b, out=None):
...
     def dot(self, b, out=None):
         return self.solve(b, out=out)
+
+#===============================================================================
+class GMRES(InverseLinearOperator):
+    """
+    A LinearOperator subclass. Objects of this class are meant to be created using :func:~`solvers.inverse`.
+
+    The .dot (and also the .solve) function are based on the 
+    Generalized minimum residual algorithm for solving linear system Ax=b.
```
yguclu, line 1700: There are two small errors in the name:

- minimum -> minimal
- "generalized" should not be capitalized

psydac/linalg/solvers.py
```diff
@@ -1718,3 +1690,263 @@ def solve(self, b, out=None):
...
+        self._check_options(**self._options) 
+        self._tmps = {key: domain.zeros() for key in ("r", "p", "v", "lv")}
+
+        # Initialize upper Hessenberg matrix
+        self._H = np.zeros((self._options["maxiter"] + 1, self._options["maxiter"]), dtype=A.dtype)
+        self._Q = []
+
+        self._info = None
+        
+
```
yguclu, line 1754: One empty line is enough.

psydac/linalg/solvers.py
```diff
@@ -1718,3 +1690,263 @@ def solve(self, b, out=None):
...
+        
+        h[k+1] = sqrt(p.dot(p).real)
+        p /= h[k+1] # Normalize vector
+
+        if len(self._Q) > k + 1:
+            p.copy(out=self._Q[k+1])
+        else:
+            self._Q.append(p.copy())
+
+
```
yguclu, line 1928: One empty line is enough.

yguclu (changes requested):
Good job! I have a couple of comments.

yguclu (approved):

PR #305, opened 2023-05-23
Accelerate kernel files with Pyccel at installation

### LINKED ISSUE #287: pyccelization of api/ast/linalg_kernels
The function inside `api/ast/linalg_kernels` are pyccelized on the fly by `api/ast/linalg` inside the `class TransposeOperator`.
This creates a bug if the kernels are pyccelized by hand (the code then tries to pyccelize a already pyccelized function.
The on the fly code writing and compilation are also slower than pyccelizing the code by hand once and for all (around 10s lost per execution).

tomcaruso:
Create a `setup.py` which accelerates all the kernels modules using Pyccel during the classical installation of Psydac. Rename these files using the unique pattern `[module]_kernels.py`.

Remove classes `TransposeOperator` and `InterfaceTransposeOperator` from `ast/api/linalg.py`, and prefer using a transpose function which is accelerated at installation. This fixes #287.

Also, add a Python script which accelerates all the files in the cloned repository after an editable install:
```bash 
python3 /path/to/psydac/psydac_accelerate [--language LANGUAGE] [--openmp]
```
If no language is given, the files will be pyccelised in Fortran. (For now, we cannot pyccelise our kernels in C due to some Numpy function which cannot be not pyccelised). This script is only necessary (and useful) in editable mode, because after a classical installation the kernels are already pyccelised.

Further changes
----------------
- Fix problem on macOS with OpenMP: export the environment variable `OMP_NUM_THREADS` before running the unit tests, and do not change `os.environ['OMP_NUM_THREADS']` in the test modules.
- Remove useless instruction from `psydac/core/bsplines_kernels.py`, which caused a warning during pyccelization.
- Increase relative and absolute tolerances for floating point comparisons in tests.
- Collect all `linalg` kernel files in the new folder `psydac/linalg/kernels`.


psydac/cmd/accelerate.py
```diff
@@ -0,0 +1,10 @@
+def main( language='fortran'):
+    import os
+    psydac_path=os.path.dirname(os.path.abspath(__file__))
+    print(psydac_path)
+    os.system('pyccel ' + psydac_path + '/../linalg/kernels.py --language '+ language)
+    os.system('pyccel ' + psydac_path + '/../api/ast/linalg_kernels.py --language '+ language)
+    os.system('pyccel ' + psydac_path + '/../core/kernels.py --language '+ language)
+    os.system('pyccel ' + psydac_path + '/../core/bsplines_pyccel.py --language '+ language)
+    return
```
yguclu, line 9: Are you sure that the `language` parameter is read from the terminal?

tomcaruso: @yguclu I change the method by creating an environment variable named PYCCEL_LANGUAGE. The method `pyccel-accelerate` will read the variable and checked if it is a supported language. If the variable doesn't exist, it will pyccelise the files with fortran

setup.py
```diff
@@ -0,0 +1,76 @@
...
+        level=distutils.log.INFO)
+    subprocess.run(command2, check=True, shell=True)
+    self.announce(
+        'Running commands: %s' % str(command3),
+        level=distutils.log.INFO)
+    subprocess.run(command3, check=True, shell=True)
+    self.announce(
+        'Running commands: %s' % str(command4),
+        level=distutils.log.INFO)
+    subprocess.run(command4, check=True, shell=True)
```
tomcaruso, line 59: @yguclu This part of the code don't pass the codacy test. It's complaining about the parameters `shell=True` that could be really problematic. The problem is that to use pyccel, it seems that we need this variable. Do you have any idea on how to solve this ?

psydac/cmd/accelerate.py
tomcaruso: This issue is outdated

psydac/linalg/stencil.py
```diff
@@ -14,7 +14,8 @@
 from psydac.linalg.basic   import VectorSpace, Vector, LinearOperator
 from psydac.ddm.cart       import find_mpi_type, CartDecomposition, InterfaceCartDecomposition
 from psydac.ddm.utilities  import get_data_exchanger
-from .kernels              import *
+from .stencil2coo_kernels              import *
+from .transpose_kernels                import *
```
yguclu, line 18: Maybe we can import the functions one by one here?

psydac/linalg/stencil.py
```diff
@@ -1917,9 +1853,8 @@ def __init__( self, V, W, s_d, s_c, d_axis, c_axis, d_ext, c_ext, *, flip=None,
         self._args         = args.copy()
         self._func         = self._dot
 
-        self._transpose_args_null = self._prepare_transpose_args()
-        self._transpose_args      = self._transpose_args_null.copy()
-        self._transpose_func      = self._transpose
+        self._transpose_args= self._prepare_transpose_args()
+        self._transpose_func      = eval(f'interface_transpose_{self._ndim}d')
```
yguclu, line 1857: Please align the equal signs and remove the useless empty spaces.

psydac/linalg/stencil.py
```diff
@@ -1629,77 +1601,41 @@ def _prepare_transpose_args(self):
...
-        args['sk']      = tuple(sk)
-        args['sl']      = tuple(sl)
-        return args
+                 + (-(p % mj) if mi > mj else 0)\
+                 + (-p + mj * (p//mi) if mi < mj else 0)\
+                 + (-p + mj * (p//mi) if mi == mj else 0)\
+                 for mi, mj, n, p in zip(cm, dm, ndiagsT, pp)]
+
+        args = [*nrows, *ncols, *gpads, *pp, *dm, *cm, *ndiags, *ndiagsT, *si, *sk, *sl]
+        return np.int64(args)
```
yguclu, line 1628: Originally `args` was a dictionary, so one could pass it as `**args` without worrying about ordering the arguments. In the current solution `args` is a list, so the order of the arguments becomes very important. Are you sure that this is a robust choice?

psydac/linalg/tests/test_stencil_matrix.py
```diff
@@ -2001,193 +2001,6 @@ def test_stencil_matrix_2d_serial_backend_dot_3(dtype, n1, n2, p1, p2, s1, s2, P
...
-    else:
-        M1[0:n1, :] = np.random.random((n1, 2 * p1 - 1))
-        M2[0:n1 - 1, :] = np.random.random((n1 - 1, 2 * p1 - 1))
-    # If any dimension is not periodic, set corresponding periodic corners to zero
-    M1.remove_spurious_entries()
-    M2.remove_spurious_entries()
-
-    # Convert stencil objects to Numpy arrays
-    M1a = M1.toarray()
-    M2a = M2.toarray()
```
yguclu, line 2046: Does the `tosparse` method use an accelerated kernel?

psydac_accelerate.py
```diff
@@ -0,0 +1,52 @@
+
+import argparse
+import os
+from subprocess import run as sub_run
+import shutil
+
+# Add Argument --language at the command psydac-accelerate
+parser = argparse.ArgumentParser(
+    formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    description="Get language for the pyccelisation."
```
yguclu, line 10: This is the description of the command `psydac-accelerate`...

yguclu (changes requested):
Great job @tomcaruso! I have a few comments.

psydac/linalg/tests/test_stencil_matrix.py
```diff
@@ -2001,193 +2001,6 @@ def test_stencil_matrix_2d_serial_backend_dot_3(dtype, n1, n2, p1, p2, s1, s2, P
     assert np.allclose(y1a, y1a_exact, rtol=1e-13, atol=1e-13)
     assert np.allclose(y2a, y2a_exact, rtol=1e-13, atol=1e-13)
 
-# ===============================================================================
```
saidctb, line 2004: Why are these test cases deleted?

tomcaruso: Those tests try to do the transpose with different backend. But now the transpose function are stored in a kernel that should be pyccelised at the installation of psydac. So the `backend` passed to the `StencilMatrix`, will not change the transpose function. It is only defined during the initialization of the `StencilMatrix`. So those tests are useless

psydac_accelerate.py
tomcaruso: I have changed the comment + I add a small text to explain the goal of this file.
https://github.com/pyccel/psydac/pull/305/commits/dae1195ba40b3dc9757c340658811457ea7d5064

psydac_accelerate.py
```diff
@@ -0,0 +1,52 @@
...
+                    action='store',
+                    dest='language',
+                    help='Language used to pyccelise all the _kernels files'
+                    )
+
+parser.add_argument('--openmp',
+                    default=False,
+                    action='store_true',
+                    dest='openmp',
+                    help='Do we read the OpenMP instrucions'
```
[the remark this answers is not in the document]
tomcaruso: [bb927a5](https://github.com/pyccel/psydac/pull/305/commits/bb927a59eae6b9774bbe1a3c3cf2a01b5e55ddbb)

psydac/linalg/stencil.py
tomcaruso: [d81b27f](https://github.com/pyccel/psydac/pull/305/commits/d81b27f827b9ed72705d739d76ff1fa374119e2b)

psydac/linalg/tests/test_stencil_matrix.py
tomcaruso: Yes the `toarray` and the `tosparse` method both used the kernels `stencil2coo_{dim}d_{order}` EXCEPT in the case were we want to get the pads in parallel mode

saidctb (approved):

psydac_accelerate.py
yguclu: What I meant is that the string passed to the argument `description` is what gets printed when you ask for help on the command line. It should describe the script itself, something like `"Accelerate all computational kernels in Psydac using Pyccel (editable install only)"`.

yguclu (approved):

PR #310, opened 2023-06-26
Add kernels for various linear algebra operations

vcarlier:
Add kernels for:
- inner product of two `StencilVector` objects belonging to the same space (still called `dot` for now);
- `axpy` operation `y = a * x + y` of `StencilVectorSpace`, where `(x, y)` are `StencilVector` objects and `a` is scalar;
- Matrix-vector product of `StencilMatrix` (called `dot`).

Further details
---------------
Added 3 kernels for the different possible inner product (1/2/3 d)
Updated the call in stencil according to this changes
Beware, the complex conjugate is on the first variable as this was the case in the previous implementation and swapping that would break a lot of other things. This will need to be updated when everything is changed to use conjugate on the first variable.

Added 3 kernels for the Matrix vector dot product and updated the call in stencil matrix according to this changes. 
Some parameters were added in the `self._arg` for the dot product in the init but the unnecessary arguments are removed in `set_backend`.


psydac/linalg/kernels.py
```diff
@@ -189,3 +189,134 @@ def stencil2coo_3d_F(A:'T', data:'data', rows:'int64[:]', cols:'int64[:]',
...
+        pads0, shift0, shape0 : int
+            pads, shift and shape of the space of the vectors in the 0 direction.
+            
+    Returns
+    -------
+        result : float containing the results'''
+    result = float(0)
+    for i0 in range(pads0*shift0,shape0-pads0*shift0):
+        result += v1[i0]*v2[i0]
+    return result
```
tomcaruso, line 213: @yguclu For all the implemented kernels, we should initialize the `result` with the good type. Thus, @vcarlier decide to separate complex and real cases to initialize `result` as a scalar. But we could unify the complex and real cases by initializing `result` as a numpy array of size (1) with the line :
`result = np.zeros_like(v1, shape=(1,)`

Thus we could reduce the amount of code but it may be slower. What do you think ?

Also, we could also pass the result as an argument of the kernel and thus we should be able to create only one kernel for both case with another pyccel template for this new argument.

yguclu: @tomcaruso I think it is a very good idea to use a Numpy array of size 1 for reducing code duplication. The resulting code may be slower if we are not careful, because Fortran will use an `allocatable` array which is stored in the heap. To avoid this we should use the [`stack_array`](https://github.com/pyccel/pyccel/blob/devel/tutorial/decorators.md#stack-array) decorator:

```python
import numpy as np

from pyccel.decorators import stack_array


@template('T', ['float[:]', 'complex[:]'])
@stack_array('zeros')
def dot_product_1d(
      v1: 'T',
      v2: 'T',
      pads0 : 'int64',
      shift0: 'int64',
      shape0: 'int64'):

    zeros  = np.zeros_like(v1, shape=(1,))
    result = zeros[0]

    for i0 in range(pads0 * shift0, shape0 - pads0 * shift0):
        result += v1[i0] * v2[i0]

    return result
```

Hopefully the compiler will understand that the `zeros` array is useless and it will optimize it out.

yguclu: Actually there is a much simpler solution as suggested by @EmilyBourne:

```python
from pyccel.decorators import template

@template('T', ['float[:]', 'complex[:]'])
def dot_product_1d(
      v1: 'T',
      v2: 'T',
      pads0 : 'int64',
      shift0: 'int64',
      shape0: 'int64'):

    result = v1[0] - v1[0]   # = 0

    for i0 in range(pads0 * shift0, shape0 - pads0 * shift0):
        result += v1[i0] * v2[i0]

    return result
```

yguclu:
> We need to add OpenMP pragmas to handle the multithreading case otherwise, it will be slow when we use hybrid parallelization

Good point! I suppose that we can start by adding the following instruction before our nested loops:
```
#$ omp parallel for collapse(N) reduction(+ : result)
```
where `N` is the number of dimensions. (In the 1D case the clause `collapse(1)` is redundant.)

saidctb:
> > We need to add OpenMP pragmas to handle the multithreading case otherwise, it will be slow when we use hybrid parallelization
> 
> Good point! I suppose that we can start by adding the following instruction before our nested loops:
> 
> ```
> #$ omp parallel for collapse(N) reduction(+ : result)
> ```
> 
> where `N` is the number of dimensions. (In the 1D case the clause `collapse(1)` is redundant.)

We also need to specify private and shared variables to get better performance.

yguclu:
> > > We need to add OpenMP pragmas to handle the multithreading case otherwise, it will be slow when we use hybrid parallelization
> > 
> > 
> > Good point! I suppose that we can start by adding the following instruction before our nested loops:
> > ```
> > #$ omp parallel for collapse(N) reduction(+ : result)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > where `N` is the number of dimensions. (In the 1D case the clause `collapse(1)` is redundant.)
> 
> We also need to specify private and shared variables to get better performance.

Then, in 3D:
```
#$ omp parallel private(i0, i1, i2) shared(result)
#$ omp for collapse(3) reduction(+ : result)
```

saidctb:
> > > > We need to add OpenMP pragmas to handle the multithreading case otherwise, it will be slow when we use hybrid parallelization
> > > 
> > > 
> > > Good point! I suppose that we can start by adding the following instruction before our nested loops:
> > > ```
> > > #$ omp parallel for collapse(N) reduction(+ : result)
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > where `N` is the number of dimensions. (In the 1D case the clause `collapse(1)` is redundant.)
> > 
> > 
> > We also need to specify private and shared variables to get better performance.
> 
> Then, in 3D:
> 
> ```
> #$ omp parallel private(i0, i1, i2) shared(result)
> #$ omp for collapse(3) reduction(+ : result)
> ```

It would be like this in 3D
```
 #$ omp parallel default(private) firstprivate( pads0, pads1, pads2, shift0, shift1 shift2, shape0, shape1, shape2) shared(result, v1, v2)
 #$ omp for collapse(3) reduction(+ : result)
```

psydac/linalg/kernels.py
tomcaruso: Ok in fact we could not fuze both function because the conjugate always return a complex data even if the entry data is real. Thus, the file could not be pyccelise, because the data type is not consistent and managing complex even in real case will slow the code

tomcaruso:
> It would be like this in 3D
> 
> ```
>  #$ omp parallel default(private) firstprivate( pads0, pads1, pads2, shift0, shift1 shift2, shape0, shape1, shape2) shared(result, v1, v2)
>  #$ omp for collapse(3) reduction(+ : result)
> ```

Pyccel is complaining about result that don't have a specified type : 
``` bash 
2019 |     !$omp &shared(result, v1, v2)
     |                        
Error: Symbol ‘result’ at (1) has no IMPLICIT type; did you mean ‘result_0001’?
/home/cato/Code/psydac/psydac/linalg/__pyccel__/kernels.f90:1970:72:


```
Also the fact that the type of result is not specified create an issue on the reduction : 

```
 1869 |     !$omp do collapse(1) reduction(+: result)
      |                                     
Error: !$OMP DECLARE REDUCTION + not found for type UNKNOWN at (1)

```

yguclu:
> Also the fact that the type of result is not specified create an issue on the reduction :
> 
> ```
>  1869 |     !$omp do collapse(1) reduction(+: result)
>       |                                     
> Error: !$OMP DECLARE REDUCTION + not found for type UNKNOWN at (1)
> ```

This error is caused by the first one, you do not need to worry about it.

EmilyBourne:
> > It would be like this in 3D
> > ```
> >  #$ omp parallel default(private) firstprivate( pads0, pads1, pads2, shift0, shift1 shift2, shape0, shape1, shape2) shared(result, v1, v2)
> >  #$ omp for collapse(3) reduction(+ : result)
> > ```
> 
> When I try to add the openMP instruction and the flag `--openmp` in the pyccel command, pyccel is complaining about result that don't have a specified type :
> 
> ```shell
> 2019 |     !$omp &shared(result, v1, v2)
>      |                        
> Error: Symbol ‘result’ at (1) has no IMPLICIT type; did you mean ‘result_0001’?
> /home/cato/Code/psydac/psydac/linalg/__pyccel__/kernels.f90:1970:72:
> ```
> 
> Also the fact that the type of result is not specified create an issue on the reduction :
> 
> ```
>  1869 |     !$omp do collapse(1) reduction(+: result)
>       |                                     
> Error: !$OMP DECLARE REDUCTION + not found for type UNKNOWN at (1)
> ```

This is related to this issue: https://github.com/pyccel/pyccel/issues/590

The contents of the clause are not parsed through pyccel so any variables that are renamed are not renamed in the clause. `result` is a keyword in Fortran so it gets renamed before printing. This is then causing the error you see here. The code will compile if you use a different variable name

psydac/linalg/block.py
```diff
@@ -292,6 +292,14 @@ def __setitem__(self, key, value):
         assert isinstance(value, Vector)
         self._blocks[key] = value
 
+    #...
+    def axpy(self, v2, a):
+        assert isinstance(v2, BlockVector)
+        assert v2._space is self._space
+        for i in range(len(self.blocks)):
+            self.blocks[i].axpy(v2.blocks[i], a)
+        self._sync = self._sync and v2._sync
```
yguclu, line 301: `axpy` is a standard BLAS function which has signature`axpy(a, x, y)` (excluding those C arguments that are useless in Python).

Therefore, if we want to use the same name for this function we should also try to keep the same signature. For this reason it would be natural to make `axpy` a method of the `VectorSpace` base class, to be subclassed by `StencilVectorSpace` and `BlockVectorSpace`:

```python
class VectorSpace:
    ...
    @abstractmethod
    def axpy(self, a, x, y):
        """ Docstring. """
```

If we want to add a method to the `Vector` base class instead, then we should change its name to something like `mul_iadd`:

```python
class Vector:
    ...
    @abstractmethod
    def mul_iadd(self, a, x):
        """ Docstring. """
```

psydac/linalg/basic.py
```diff
@@ -52,6 +52,20 @@ def dot(self, a, b):
 
         """
 
+    @abstractmethod
+    def axpy(self, a, x, y):
+        """
+        This method compute the operation x+=a*y.
```
yguclu, line 58: This is not correct, it should be `y += a * x`. See the BLAS documentation [here](https://spec.oneapi.io/versions/latest/elements/oneMKL/source/domains/blas/axpy.html).

tomcaruso: Ah I thought the name was strange for the purpose of the function

psydac/linalg/block.py
tomcaruso: This PR create the two function
[9080230](https://github.com/pyccel/psydac/pull/310/commits/9080230dd760c0c6884c163acd3e2e1d1d3aa46c)

psydac/linalg/basic.py
tomcaruso: This PR correct it, Fortunately, we just have to change the order of the arguement in `mul_iadd` method to make all the previous call work again:
[4e98e4c](https://github.com/pyccel/psydac/pull/310/commits/4e98e4c52fd97dac04a948466a0dc7180cdc4239)

psydac/linalg/block.py
jowezarek: I guess I prefer mul_iadd.

Using axpy, we have to include the VectorSpace. But in certain situations, we create vectors without explicitly using a VectorSpace, for example when projecting a continuous function into a discrete space:
```
domain  = logical_domain
derham  = Derham(domain)

domain_h = discretize(domain, ncells=ncells, periodic=periodic, comm=MPI.COMM_WORLD)
derham_h = discretize(derham, domain_h, degree=degree)

P0, P1, P2, P3  = derham_h.projectors(nquads=[5,5,5])

a_1 = lambda x, y, z : ...
a_2 = lambda x, y, z : ...
a_3 = lambda x, y, z : ...
a   = (a_1, a_2, a_3)

a_coeff = P1(a).coeffs 
``` 
If I wanted to add a multiple of a different coefficient vector to `a_coeff`, I'd first have to explicitly mention `V1.vector_space`, which was not necessary previously.

psydac/linalg/kernels.py
```diff
@@ -443,7 +443,7 @@ def axpy_3d(alpha: 'T', x: "Tarray", y: "Tarray",shape0:'int64',shape1:'int64',s
     for i1 in range(0, shape0, 1):
         for i2 in range(0, shape1, 1):
             for i3 in range(0, shape2, 1):
```
yguclu: You can just write
```suggestion
    for i1 in range(shape0):
        for i2 in range(shape1):
            for i3 in range(shape2):
```
But why do we need to pass  `shape0`, `shape1` and `shape2` to this function? One can just write
```python
    n1, n2, n3 = x.shape

    for i1 in range(n1):
        for i2 in range(n2):
            for i3 in range(n3):
```

psydac/linalg/basic.py
tomcaruso: This PR correct a typo I didn't see 
 [6133446](https://github.com/pyccel/psydac/pull/310/commits/613344625a0e5b965475f651a3491f68ec5896ef)

psydac/linalg/kernels.py
tomcaruso: This commit correct it 81d63c011b84ef5f43e5a8e9acf5e1b6b133c22a

psydac/linalg/block.py
yguclu: Tom has provided both functions `VectorSpace.axpy()` and `Vector.mul_iadd()`.

yguclu (approved):

PR #316, opened 2023-07-09
Bug Fix

saidctb:
Do not create a `TensorFemSpace` twice when the patches are conforming.


yguclu:
Hey @saidctb thanks for this bug fix! What kind of error did you get before the fix?

tomcaruso:
> Hey @saidctb thanks for this bug fix! What kind of error did you get before the fix?

We get an AssertionError when we tried to call the mesh `multipatch/square_repeated_knots.h5` in parallel

self = <psydac.ddm.cart.CartDecomposition object at 0x7fb74fd60f70>
starts = (8, 0), ends = (3, 17), parent_starts = (None, None)
parent_ends = (None, None)

    def change_starts_ends( self, starts, ends, parent_starts,  parent_ends):
        """ Create a slice of the cart based on the new starts and ends.
        WARNING! this function should be used carefully,
        as it might generate errors if it was not used properly in the communication process.
        """
        cart = CartDecomposition(self._domain_decomposition, self._npts, self._global_starts, self._global_ends, self._pads, self._shifts)
    
>       assert self.comm is None or self.comm.size == 1
E       AssertionError

tomcaruso (approved):

PR #322, opened 2023-07-18
Invert quadrature loops

saidctb:
Invert the quadrature loop in the assembly methods of linear and bilinear forms, which means that we loop over the quadrature points before looping over the basis functions. This loop inversion is not automatic, we do it only in the case where we have a spline or NURBs mapping, or where the analytical mapping is not affine.

To deactivate the loop inversion we just set the flag here to `False`: https://github.com/pyccel/psydac/blob/1b4787f1eab756b976cf6e8627e75a0c29767f12/psydac/api/ast/fem.py#L473

- Add `lhs` property to `ComputeLogicalBasis` and delete `prefix` property from:
    - `BasisAtom`
    - `IteratorBase`
    - `LogicalBasisValue`
    - `Loop`
- Delete initialisation code from `_visit_EvalField` and `_visit_EvalMapping` and move it to the constructor of the classes `EvalField` and `EvalMapping`

When looping over the quadrature points first, the vector assembly algorithm in 1d looks like this
```python
def assemble_vector(test_basis, spans, global_x1, p, ne, nq, g_vec):

    p     :'degree of the basis functions'
    ne    : 'number of elements'
    nq    : 'number of quadrature points'
    g_vec : 'global vector'
    lvec  : 'element vector'
    
    local_x1 = zeros_like(global_x1[0,:])
    l_vec    = zeros((p+1,), dtype='float64')
    
   # element loop
    for ie in range(ne):
        local_x1[:] = global_x1[ie,:]
        span        = spans[ie]
        l_vec_v[:]  = 0.0
        
        # quadrature loop
        for iq in range(0, nq, 1):
            x1 = local_x1[iq]
           
            # basis function loop
            for ib in range(0, 1 + p, 1):
                v = test_basis[ie,ib,0,iq]
                l_vec[ib] += -4*pi**2*v*sin(2*pi*x1)

        g_vec_v[span - p:1 + p + span] += l_vec[:]
```


yguclu (commented):
Thanks for this PR, @saidctb!

I see that you have made many changes to the Psydac AST, which has the useful side effect of simplifying the parser considerably.

In order to properly review this PR we need a bit more information. First you should provide some pseudo-code which shows how the loops are arranged in the two cases. Next you should add a list of changes to the PR description. I look forward to this!

saidctb:
> Thanks for this PR, @saidctb!
> 
> I see that you have made many changes to the Psydac AST, which has the useful side effect of simplifying the parser considerably.
> 
> In order to properly review this PR we need a bit more information. First you should provide some pseudo-code which shows how the loops are arranged in the two cases. Next you should add a list of changes to the PR description. I look forward to this!

done

yguclu:
> > Thanks for this PR, @saidctb!
> > I see that you have made many changes to the Psydac AST, which has the useful side effect of simplifying the parser considerably.
> > In order to properly review this PR we need a bit more information. First you should provide some pseudo-code which shows how the loops are arranged in the two cases. Next you should add a list of changes to the PR description. I look forward to this!
> 
> done

Thanks!

I think that we can use stack arrays for the variables `local_x1` and `l_vec`. This should be beneficial when the local problem size is small (e.g. for strong-scaling).

saidctb:
> > > Thanks for this PR, @saidctb!
> > > I see that you have made many changes to the Psydac AST, which has the useful side effect of simplifying the parser considerably.
> > > In order to properly review this PR we need a bit more information. First you should provide some pseudo-code which shows how the loops are arranged in the two cases. Next you should add a list of changes to the PR description. I look forward to this!
> > 
> > 
> > done
> 
> Thanks!
> 
> I think that we can use stack arrays for the variables `local_x1` and `l_vec`. This should be beneficial when the local problem size is small (e.g. for strong-scaling).

I tried that before but I didn't get any performance from using them

yguclu:
By reading the PR description it sounds like the loop inversion is always applied. But this is not what I understand from the code changes. Can you please explain when the loop in version is actually used (and why) in the PR description?

saidctb:
> By reading the PR description it sounds like the loop inversion is always applied. But this is not what I understand from the code changes. Can you please explain when the loop inversion is actually used (and why) in the PR description?

Yes, the loop inversion is not automatic, we do it only in the case where we have a Spline or Nurbs mapping or when the analytical mapping is not Affine, these are the cases where the loop inversion gives better performance.
It would be interesting to do a complexity analysis using Sympy and depending on the different cases we determine if we should invert the loops or not.

yguclu (approved):

PR #321, opened 2023-07-17
Upgrade SymPy

tomcaruso:
Update Psydac to support SymPy up to version 1.9 after the corresponding changes in SymPDE.

**Main changes**

- import cache from `sympy.core`
- import `with_metaclass` from `sympde.old_sympy_utilities` because SymPy does not provide it anymore
- Store the dictionaries `_known_functions_{numpy|math|mpmath}` and `_known_constants_{numpy|math|mpmath}` in `psydac.api.ast.utilities` for the same reason.

PR #325, opened 2023-07-27
Use SemiNorm

saidctb:
Use H^1 and H^2 `SemiNorm` from SymPDE. Previously the class `Norm`
was used for all seminorms, which was not mathematically correct.


yguclu:
This PR appears to be blocked by #321, which should take precedence.

psydac/api/tests/test_2d_biharmonic.py
```diff
@@ -74,9 +74,9 @@ def run_biharmonic_2d_dir(solution, f, dir_zero_boundary, ncells, degree, backen
 
     # Error norms
     error  = u - solution
-    l2norm = Norm(error, domain, kind='l2')
-    h1norm = Norm(error, domain, kind='h1')
-    h2norm = Norm(error, domain, kind='h2')
+    l2norm = SemiNorm(error, domain, kind='l2')
+    h1norm = SemiNorm(error, domain, kind='h1')
+    h2norm = SemiNorm(error, domain, kind='h2')
```
yguclu, line 79: I think it is confusing to write `SemiNorm(..., kind='l2')` in the code, given that it is equivalent to `Norm(..., kind='l2')`. In this case we should prefer the `Norm` class wherever possible.

@campospinto  @saidctb Do you agree?

psydac/api/fem.py
```diff
@@ -1555,7 +1555,7 @@ def assemble(self, **kwargs):
                 args += (v, )
 
         v = self._func(*args)
-        if isinstance(self.expr, sym_Norm):
+        if isinstance(self.expr, sym_SemiNorm):
```
yguclu, line 1558: This will not work for a `Norm`, unless we make `Norm` subclass `SemiNorm`.

psydac/api/tests/test_2d_biharmonic.py
yguclu: I have replaced L2 seminorm with L2 norm in 975404e

psydac/api/fem.py
yguclu: Fixed in 976e1f7.

psydac/api/tests/test_2d_biharmonic.py
saidctb: okay, it is fine with me

Issue #188, opened 2022-02-21
Improve the ways to evaluate `FemField`s


anlavandier:
Currently, there are only two ways of evaluating `FemField`s, point-wise, and one the grid defined in `TensorFemSpace.preprocess`. 

Ideally the interface should allow for those 4 options:
1. scalar coordinates
2. 1D arrays of coordinates -> grid is tensor-product, but refinement is not the same in each cell
3. 2D arrays of coordinates -> grid is tensor-product, and each cell has the same number of cells
4. N-D arrays of coordinates -> unstructured grid

PR #335, opened 2023-09-26
Add single 3d field evaluation kernel

jowezarek:
Adds the kernel `eval_field_3d_once` to `psydac.core.field_evaluation_kernels.py` and makes use of it in `psydac.fem.tensor.py`'s method `eval_field`.

The purpose of this kernel is to accelerate the evaluation of a single 3d field at a single point. This is needed repeatedly, for example, when creating a Poincarè plot tracing a single fieldline.

### DIFF
--- a/psydac/fem/tensor.py
+++ b/psydac/fem/tensor.py
@@ -38,7 +38,8 @@ from psydac.core.field_evaluation_kernels import (eval_fields_2d_no_weights,
                                                   eval_fields_3d_no_weights,
                                                   eval_fields_3d_irregular_no_weights,
                                                   eval_fields_3d_weighted,
-                                                  eval_fields_3d_irregular_weighted)
+                                                  eval_fields_3d_irregular_weighted,
+                                                  eval_field_3d_once)
 
 __all__ = ('TensorFemSpace',)
 
@@ -230,9 +231,7 @@ class TensorFemSpace( FemSpace ):
         #   - Pros: small number of Python iterations = ldim
         #   - Cons: we create ldim-1 temporary objects of decreasing size
         #
-        res = coeffs
-        for basis in bases[::-1]:
-            res = np.dot( res, basis )
+        res = eval_field_3d_once(coeffs, bases[0], bases[1], bases[2])
 
 #        # Option 2: cycle over each element of 'coeffs' (touched only once)
 #        #   - Pros: no temporary objects are created
--- a/psydac/core/field_evaluation_kernels.py
+++ b/psydac/core/field_evaluation_kernels.py
@@ -5,6 +5,32 @@ from pyccel.decorators import template
 # Field evaluation functions
 # =============================================================================
 # -----------------------------------------------------------------------------
+# 0: Evaluation of single 3d field at single point
+# -----------------------------------------------------------------------------
+@template(name='T1', types=['float[:,:,:]', 'complex[:,:,:]'])
+@template(name='T2', types=['float[:]', 'complex[:]'])
+def eval_field_3d_once(local_coeffs: 'T1', 
+                       local_bases_0: 'T2', local_bases_1: 'T2', local_bases_2: 'T2'):
+    """
+    Parameters
+    ----------
+    local_coeffs: ndarray of floats
+        Active (local) coefficients of the fields in all directions
+
+    local_bases: list of ndarrays
+        Active (local) 1D-basis functions values at the point of evaluation. 
+    """
+    res = local_bases_0[0] - local_bases_0[0]
+
+    for i, c1 in enumerate(local_coeffs):
+        for j, c2 in enumerate(c1):
+            for k, c3 in enumerate(c2):
+                res += c3 * local_bases_0[i] * local_bases_1[j] * local_bases_2[k]
+
+    return res
+
+
+# -----------------------------------------------------------------------------
 # 1: Regular tensor grid without weight
 # -----------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


jowezarek:
The current use of `template` does not work @yguclu:
```
@template(name='T1', types=['float[:,:,:]', 'complex[:,:,:]'])
@template(name='T2', types=['float[:]', 'complex[:]'])
```
I get the error message 
```
ERROR at annotation (semantic) stage
pyccel:
 |error [semantic]: field_evaluation_kernels.py [28,16]| Variable has already been defined with type 'float64'. Type 'complex128' in assignment is incompatible (res=c3 * local_bases_0[i] * local_bases_1[j] * local_bases_2[k])
```
Locally I will switch back to the `float`-only case for now. Do you know how to properly make sure that dtype `complex` can also be handled?

yguclu:
Related to issue #188

yguclu:
> The current use of `template` does not work @yguclu:
> 
> ```
> @template(name='T1', types=['float[:,:,:]', 'complex[:,:,:]'])
> @template(name='T2', types=['float[:]', 'complex[:]'])
> ```
> 
> I get the error message
> 
> ```
> ERROR at annotation (semantic) stage
> pyccel:
>  |error [semantic]: field_evaluation_kernels.py [28,16]| Variable has already been defined with type 'float64'. Type 'complex128' in assignment is incompatible (res=c3 * local_bases_0[i] * local_bases_1[j] * local_bases_2[k])
> ```
> 
> Locally I will switch back to the `float`-only case for now. Do you know how to properly make sure that dtype `complex` can also be handled?

The templates should work now, provided that you upgrade Pyccel to version >= 1.11.2

vcarlier:
@yguclu 
I cannot reproduce the error in the tests on my laptop (even when running pytest in parallel), do you have any idea why?

yguclu:
> @yguclu I cannot reproduce the error in the tests on my laptop (even when running pytest in parallel), do you have any idea why?

You did not forget to run the script `psydac_accelerate.py`, did you?

yguclu:
> > @yguclu I cannot reproduce the error in the tests on my laptop (even when running pytest in parallel), do you have any idea why?
> 
> You did not forget to run the script `psydac_accelerate.py`, did you?

I see the problem now. This branch has not been updated with `devel`, so it is still using the old version of Pyccel.

yguclu:
@jowezarek Is there a reason for allowing the basis functions to be complex valued?

Instead of this code
```python
@template(name='T1', types=['float[:,:,:]', 'complex[:,:,:]'])
@template(name='T2', types=['float[:]', 'complex[:]'])
def eval_field_3d_once(local_coeffs: 'T1', 
                       local_bases_0: 'T2', local_bases_1: 'T2', local_bases_2: 'T2'):
```
we can just write
```python
@template(name='T', types=[float, complex])
def eval_field_3d_once(local_coeffs: 'T[:,:,:]', 
                       local_bases_0: 'float[:]', local_bases_1: 'float[:]', local_bases_2: 'float[:]'):
```

jowezarek:
> @jowezarek Is there a reason for allowing the basis functions to be complex valued?
> 
> Instead of this code
> 
> ```python
> @template(name='T1', types=['float[:,:,:]', 'complex[:,:,:]'])
> @template(name='T2', types=['float[:]', 'complex[:]'])
> def eval_field_3d_once(local_coeffs: 'T1', 
>                        local_bases_0: 'T2', local_bases_1: 'T2', local_bases_2: 'T2'):
> ```
> 
> we can just write
> 
> ```python
> @template(name='T', types=[float, complex])
> def eval_field_3d_once(local_coeffs: 'T[:,:,:]', 
>                        local_bases_0: 'float[:]', local_bases_1: 'float[:]', local_bases_2: 'float[:]'):
> ```

@yguclu, is it possible that your proposed solution was not possible prior to your `template`-update PR? 
Either way, that is exactly what I wanted in the first place. I never intended to define a template for complex coefficients and real basis function values / real coefficients and complex basis function values.

psydac/fem/tensor.py
```diff
@@ -183,45 +191,62 @@ def eval_field( self, field, *eta, weights=None):
...
+            pad = vs.pads[0]
+            shift = vs.shifts[0]
+            index.append( slice( loc_span-degree-start+shift*pad, loc_span+1-start+shift*pad ) )
+            if weights:
+                w_index.append(slice(loc_span-degree, loc_span+1))
         # Get contiguous copy of the spline coefficients required for evaluation
+        bases = self._tmp_bf
         index  = tuple( index )
-        coeffs = field.coeffs[index].copy()
+        coeffs[:] = field.coeffs._data[index]
```
yguclu, line 246: Accessing `coeffs._data` instead of just `coeffs` does not appear to work in all cases (see failing tests).

vcarlier:
@yguclu everything seems to be fixed now

psydac/fem/tensor.py
yguclu: This appears to work now.

yguclu:
> The `Documentation / build_docs` job is failing with the error
> 
> > Extension error (sphinx.ext.autosummary): Handler <function process_generate_options ...> for event 'builder-inited' threw an exception (exception: no module named psydac.api)
> 
> I am really surprised given that we have made no change to the documentation infrastructure. I have asked Copilot and it says that this happens because Psydac cannot be imported. So it suggests to either install Psydac, or add its source directory to the PYTHONPATH. I am not convinced, but I will try the latter...
> 
> @jowezarek: What do you think?

I have decided to install Psydac as this only takes 2 minutes. Things appear to work now.

PR #339, opened 2023-09-29
De Rham sequences w/ knot multiplicity > 1

vcarlier:
Enable the use of the de Rham sequence with arbitrary knots multiplicity. With this we mean that the interior knots (all knots in the periodic case) can have a constant multiplicity higher than 1.

List of modifications
---------------------
* Changes in `core/bsplines.py` and `core/bsplines_kernels.py` : 
    - Modify the `make_knots` function to create an appropriate knot sequence in all cases. In the non periodic case the interior knots are repeated `multiplicity` times and the boundary knots `p+1` times. In the periodic case, interior knots are repeated `multiplicity` times and we add `p+1` knots at each side by periodicity. In all cases the number of knots is `(n_cells-1) * multiplicity + 2 * (p+1)`.
    - Modify functions `collocation_matrix` and `histopolation_matrix` to take in account that now in the periodic case the number of basis function is `len(knots) - 2 * (p+1) + multiplicity`. 
    - Modify `greville` function for the maximum multiplicity case to avoid having the discontinuity points of the spline.

* Changes in modules `psydac.feec.derivatives` and `psydac.feec.global_projectors` to take in account that the ghost regions have size `shifts * pads` (previously was `pads`)

* Changes in `psydac.fem.splines`: repercussion of the changes in `psydac.core.bsplines`, adding `multiplicity` as argument to several functions. Correct the computation of the `multiplicity` from the knot sequence.

* Add tests in `feec/tests/test_differentiation_matrices.py` and `feec/tests/test_global_projectors.py` with higher multiplicity.

* Add test in `api/tests/tests_api_feec_1d.py, api/tests/tests_api_feec_2d.py, api/tests/tests_api_feec_3d.py` with higher multiplicity.


psydac/api/grid.py
```diff
@@ -230,11 +230,12 @@ def space(self):
...
-    p    = space.degree     # spline degree
-    n    = space.nbasis     # total number of control points
-    grid = space.breaks     # breakpoints
-    nc   = space.ncells     # number of cells in domain (nc=len(grid)-1)
+    T     = space.knots      # knots sequence
+    p     = space.degree     # spline degree
+    n     = space.nbasis     # total number of control points
+    grid  = space.breaks     # breakpoints
+    nc    = space.ncells     # number of cells in domain (nc=len(grid)-1)
+    perio = space.perio
```
campospinto, line 238: `perio` ? Several FemSpace subclasses have a `periodic` attribute, but I could not find any `perio`.

Also, a docstring would be nice :)

psydac/core/bsplines.py
```diff
@@ -57,7 +57,7 @@
 
 
 #==============================================================================
-def find_span(knots, degree, x):
+def find_span(knots, degree, x, periodic, multiplicity = 1):
```
campospinto, line 60: The docstring should be a bit more precise:

here I guess that the `knots` passed as argument are assumed to have all the same `multiplicity`, except for boundary knots when `periodic = False`, which always have multiplicity `degree+1` . Is that correct ?

and the actual knot sequence which will be used 

here I assume that there is no repea

you mean: multiplicity of the interior knots (since boundary knots always have multiplicity = degree+1 in the non-periodic case)

psydac/api/grid.py
vcarlier: You're right but it looks like this function is actually never called and I don't know what it's suppose to do.

psydac/core/bsplines.py
vcarlier: Yes, I'll add it to the doc

campospinto: Thanks. Also in the periodic case, it would help to specify (remind ?) what is actually in the knot vector -- namely if some knots are "repeated" outside the domain to have an efficient handling of the periodicity condition.

psydac/api/grid.py
campospinto: that's already something to write in the docstring 😅

campospinto: that's already something to write in the docstring 😅

psydac/core/bsplines.py
vcarlier: Maybe I should just say that we assume that the knots sequence was created by `make_knots` and then in `make_knots` specify what it looks like?

campospinto: yes, that's a good idea -- I was actually missing this pointer to `make_knots`

and by doing this you can improve the docstring in `make_knots`: for instance the sentence 
```
Otherwise, knot sequence is clamped (i.e. endpoints are repeated p times).
```
is ambiguous (what does it mean to be repeated "one time" ?), it would be better to write:
```
Otherwise, knot sequence is clamped (i.e. endpoints have p+1 multiplicity).
```

yguclu: I do not understand why we should pass `multiplicity` to this function. The knot sequence already contains all the required information, and the computed `span` is the index of the last non-zero basis function, which has the same meaning regardless of the multiplicity.

campospinto: I was also a bit confused at the start -- thinking first that passing some `multiplicity` > 1 would further repeat the given knots, which is not the case

yguclu: I see that the interior `multiplicity` is used to compute the `high` index in the periodic case. Do we get the wrong `x` coordinate of the right boundary otherwise?

vcarlier: In the periodic case `high` would not be right, however `low` remains untouched this is due to how I compute the knots in the periodic case, for example if our space is `[a,b]` with `n` cells, my knots will be `a+i/n*(b-a)` for `i `in `[1,n]`
Then I will add `p+1` knots on the left by translating the rightmost knots, meaning that `low` is still `p`, 
But I'll add `p` more knots to the right and this is why we need to change `high` in this case

yguclu: Could we pass the dimension of the spline space instead of `multiplicity`?

vcarlier: Well since the dimension of the space, the number of knots the multiplicity and the degree are all linked we could use that yes but is it really better? I don't see why passing the multiplicity is any bad here?

yguclu: I find that using the multiplicity is not satisfactory because I would like this function to work for any knot sequence, where the multiplicity can vary arbitrarily from one knot to another. This is something that we want to support in the future.

psydac/core/bsplines_kernels.py
```diff
@@ -35,8 +43,14 @@ def find_span_p(knots: 'float[:]', degree: int, x: float):
         Springer-Verlag Berlin Heidelberg GmbH, 1997.
     """
     # Knot index at left/right boundary
+    #In the periodic case the p+1 knots is the last knot equal to the left 
+    #boundary, while the first knots being the right boundary is multiplicity+p
+    #before the last knot of the sequence (see make knots)
```
campospinto, line 48: I guess this sentence means 
```
in the periodic case, the last knot located on the left boundary is the one with index i = p
and the first knot located on the right boundary is the one with index i = -m-p
where p = degree and m = multiplicity
```
by the way, the first statement holds also in the non-periodic case.

vcarlier: Exactly

psydac/core/bsplines.py
yguclu: Conceptually I think that the knot sequence should be extended to the left always in the same way. As a result, the basis function of index 0 is the left-most B-spline which is non-zero on the left boundary.

psydac/core/bsplines_kernels.py
campospinto: As for the second statement, I'm not sure to follow the logic: I don't see why the multiplicity matters to the index of the first knot on the right boundary.

campospinto: Consider for example the periodic case with n_cells = 3, degree =2, multiplicity = 2
Then the dimension of the spline space is 6, and the knots needed to evaluate the splines on the domain are:
(-1/3, 0, 0, 1/3, 1/3, 2/3, 2/3, 1, 1, 4/3)
the first knot on the right boundary has index -3 = -degree-1

psydac/core/bsplines.py
campospinto: That is also what I would think -- unless I am missing something ?

psydac/core/bsplines_kernels.py
campospinto: ok, but why adding the second knot 4/3 in the sequence ?
Given that the B-spline defined by the knots (1, 1, 4/3, 4/3) vanishes on the domain [0,1], I don't see the purpose of this

psydac/core/bsplines.py
vcarlier: If I am not mistaken this is the case with how things are build at the moment?

vcarlier: What we could do is : 
-Build the knot sequence by repeating all the break mult times (including the boundary ones blindly).
-Add (p-m+1) knots at each side by periodicity.

Then we would have the same number of knots as in the None periodic case and enough knots to evaluate everything inside the domain.

psydac/core/bsplines_kernels.py
vcarlier: I think I remember, we are adding those basis function for compatibiliy with the parallel case where you have p function on the ghost regions. As in the non periodic case you use the ghost regions to enforce the periodicity I think we need those more basis function. 
Is that right @yguclu

yguclu: Yes, and this was designed for the case of `multiplicity == 1`. When the multiplicity is higher we need to repeat fewer extra basis functions on both sides.

In fact one could argue that a smaller ghost region is needed when `multiplicity > 1`. The problem is that the ghost regions are mostly designed with the assembled FEM matrices in mind, where the required ghost regions are larger not smaller.

yguclu: I think that we can forget about the ghost regions for the moment. We should be able to build the smallest knot sequence that is needed for the evaluation of the field, knowing that we store the correct amount of spline coefficients (equal to the dimension of the space).

psydac/core/bsplines.py
vcarlier: Yes this is the same. This would not cause problem with the ghost regions then?

campospinto: Yes. Also, note that in this way we have the same number of knots in the periodic and non-periodic case. The only difference is that in the non-periodic case the "extended" knots are all on the boundaries (p+1 on each side)

campospinto: and if I'm not mistaken you no longer need to pass any multiplicity argument to the `find_span` function

psydac/core/bsplines_kernels.py
vcarlier: I have implemented the previous solution but it's not working really well, I have really high errors in maxwell 2d (10^14 ...). 
I will continue looking to see if I can understand why it's not working.

psydac/feec/tests/test_differentiation_matrices.py
```diff
@@ -746,17 +758,19 @@ def eval_curl(fx, fy, fz, *eta):
 @pytest.mark.parametrize('degree', [(3, 2), (4, 5)])                 # 2 cases
 @pytest.mark.parametrize('periodic', [(True, False), (False, True)]) # 2 cases
 @pytest.mark.parametrize('seed', [1,3])
+@pytest.mark.parametrize('multiplicity', [[1,1],[1,2],[2,2]])
```
yguclu, line 761: More readable:
```suggestion
@pytest.mark.parametrize('multiplicity', [(1, 1), (1, 2), (2, 2)])
```

psydac/feec/tests/test_differentiation_matrices.py
```diff
@@ -820,17 +834,20 @@ def eval_div(fx, fy, *eta):
                                       (False,  True, False),
                                       (False, False,  True)])
 @pytest.mark.parametrize('seed', [1,3])
+@pytest.mark.parametrize('multiplicity', [[1,1,1],[1,2,2],[2,2,2]])
```
yguclu, line 837: More readable:
```suggestion
@pytest.mark.parametrize('multiplicity', [(1, 1, 1), (1, 2, 2), (2, 2, 2)])
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -17,14 +17,17 @@
...
+    #change mulitplicity if higher than degree to avoid problems (case p<m doesn't work)
+    multiplicity = min(multiplicity,degree)
     breaks = np.linspace(*domain, num=ncells+1)
-    knots  = make_knots(breaks, degree, periodic)
-
+    knots  = make_knots(breaks, degree, periodic, multiplicity=multiplicity)
+    
     domain_decomposition = DomainDecomposition([ncells], [periodic])
-
+
```
yguclu, line 30: Please remove leading empty spaces.

psydac/feec/tests/test_global_projectors.py
```diff
@@ -17,14 +17,17 @@
 @pytest.mark.parametrize('ncells', [500])
 @pytest.mark.parametrize('degree', [1, 2, 3, 4, 5, 6, 7])
 @pytest.mark.parametrize('periodic', [False, True])
+@pytest.mark.parametrize('multiplicity', [1,2])
```
yguclu, line 20: More readable:
```suggestion
@pytest.mark.parametrize('multiplicity', [1, 2])
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -52,16 +55,22 @@ def test_H1_projector_1d(domain, ncells, degree, periodic):
 @pytest.mark.parametrize('domain', [(0, 2*np.pi)])
 @pytest.mark.parametrize('ncells', [100, 200, 300])
 @pytest.mark.parametrize('degree', [2])
-@pytest.mark.parametrize('periodic', [False])
+@pytest.mark.parametrize('periodic', [True,False])
 @pytest.mark.parametrize('nquads', [100, 120, 140, 160])
+@pytest.mark.parametrize('multiplicity', [1,2])
```
yguclu, line 60: More readable:
```suggestion
@pytest.mark.parametrize('periodic', [True, False])
@pytest.mark.parametrize('nquads', [100, 120, 140, 160])
@pytest.mark.parametrize('multiplicity', [1, 2])
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -52,16 +55,22 @@ def test_H1_projector_1d(domain, ncells, degree, periodic):
...
+    multiplicity = min(multiplicity,degree)
     breaks = np.linspace(*domain, num=ncells+1)
-    knots  = make_knots(breaks, degree, periodic)
+    knots  = make_knots(breaks, degree, periodic, multiplicity=multiplicity)
 
     domain_decomposition = DomainDecomposition([ncells], [periodic])
 
+    # H1 space (0-forms)
+    #change mulitplicity if higher than degree to avoid problems (case p<m doesn't work)
+    multiplicity = min(multiplicity,degree)
```
yguclu, line 73: Typo: mulitplicity -> multiplicity

More readable:
```suggestion
    #change multiplicity if higher than degree to avoid problems (case p<m doesn't work)
    multiplicity = min(multiplicity, degree)
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -92,15 +101,17 @@ def test_L2_projector_1d(domain, ncells, degree, periodic, nquads):
 @pytest.mark.parametrize('ncells', [[200,200]])
 @pytest.mark.parametrize('degree', [[2,2], [2,3], [3,3]])
 @pytest.mark.parametrize('periodic', [[False, False], [True, True]])
+@pytest.mark.parametrize('multiplicity', [[1,1],[2,2]])
```
yguclu, line 104: More readable:
```suggestion
@pytest.mark.parametrize('multiplicity', [(1, 1), (2, 2)])
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -92,15 +101,17 @@ def test_L2_projector_1d(domain, ncells, degree, periodic, nquads):
...
 
-def test_derham_projector_2d_hdiv(ncells, degree, periodic):
+def test_derham_projector_2d_hdiv(ncells, degree, periodic, multiplicity):
 
     domain = Square('Omega', bounds1 = (0,2*np.pi), bounds2 = (0,2*np.pi))
     domain_h = discretize(domain, ncells=ncells, periodic=periodic)
-    
+    #change mulitplicity if higher than degree to avoid problems (case p<m doesn't work)
+    multiplicity = [min(m,p) for p, m in zip (degree, multiplicity)]
```
yguclu, line 111: Typo: mulitplicity -> multiplicity

More readable:
```suggestion
    #change multiplicity if higher than degree to avoid problems (case p<m doesn't work)
    multiplicity = [min(m, p) for p, m in zip(degree, multiplicity)]

```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -140,14 +151,16 @@ def test_derham_projector_2d_hdiv(ncells, degree, periodic):
 @pytest.mark.parametrize('ncells', [[200,200]])
 @pytest.mark.parametrize('degree', [[2,2], [2,3], [3,3]])
 @pytest.mark.parametrize('periodic', [[False, False], [True, True]])
+@pytest.mark.parametrize('multiplicity', [[1,1],[2,2]])
```
yguclu, line 154: More readable:
```suggestion
@pytest.mark.parametrize('multiplicity', [(1, 1), (2, 2)])
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -140,14 +151,16 @@ def test_derham_projector_2d_hdiv(ncells, degree, periodic):
...
 
-def test_derham_projector_2d_hdiv_2(ncells, degree, periodic):
+def test_derham_projector_2d_hdiv_2(ncells, degree, periodic, multiplicity):
 
     domain = Square('Omega', bounds1 = (0,1), bounds2 = (0,1))
     domain_h = discretize(domain, ncells=ncells, periodic=periodic)
     
+    multiplicity = [min(m,p) for p, m in zip (degree, multiplicity)]
```
yguclu, line 161: More readable:
```suggestion
    multiplicity = [min(m, p) for p, m in zip(degree, multiplicity)]
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -237,18 +250,19 @@ def test_derham_projector_2d_hcurl(ncells, degree, periodic):
 @pytest.mark.parametrize('ncells', [[30,30,30]])
 @pytest.mark.parametrize('degree', [[2,2,2], [2,3,2], [3,3,3]])
 @pytest.mark.parametrize('periodic', [[False, False, False], [True, True, True]])
+@pytest.mark.parametrize('multiplicity', [[1,1,1],[1,2,2],[2,2,2]])
```
yguclu, line 253: More readable:
```suggestion
@pytest.mark.parametrize('multiplicity', [[1,1,1], [1,2,2], [2,2,2]])
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -237,18 +250,19 @@ def test_derham_projector_2d_hcurl(ncells, degree, periodic):
...
-def test_derham_projector_3d(ncells, degree, periodic):
+def test_derham_projector_3d(ncells, degree, periodic, multiplicity):
 
     domain = Cube('Omega', bounds1 = (0,2*np.pi), bounds2 = (0,2*np.pi), bounds3 = (0,2*np.pi))
     domain_h = discretize(domain, ncells=ncells, periodic=periodic)
     
     derham   = Derham(domain)
-    derham_h   = discretize(derham, domain_h, degree=degree, get_H1vec_space = True)
+    #change mulitplicity if higher than degree to avoid problems (case p<m doesn't work)
+    multiplicity = [min(m,p) for p, m in zip (degree, multiplicity)]
```
yguclu, line 262:
```suggestion
    #change multiplicity if higher than degree to avoid problems (case p<m doesn't work)
    multiplicity = [min(m, p) for p, m in zip(degree, multiplicity)]

```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -273,19 +287,19 @@ def test_derham_projector_3d(ncells, degree, periodic):
...
-    assert maxnorm_error <= 2e-2
+    assert maxnorm_error <= 3e-2
     maxnorm_error = abs(vals_u3 - vals_f).max()
     print(ncells, maxnorm_error)
-    assert maxnorm_error <= 2e-2
+    assert maxnorm_error <= 3e-2
     maxnorm_error = abs(vals_ux_1 - vals_f).max()
     print(ncells, maxnorm_error)
-    assert maxnorm_error <= 2e-2
+    assert maxnorm_error <= 3e-2
```
yguclu, line 302: Why did you have the increase the error tolerance so much?

psydac/feec/tests/test_global_projectors.py
```diff
@@ -294,17 +308,20 @@ def test_derham_projector_3d(ncells, degree, periodic):
     degree   = 3
     periodic = True
     ncells   = [10, 20, 40, 80, 160, 320, 640]
-
+    
+    for nc in ncells:
+        test_derham_projector_2d_hdiv([nc, nc], [degree, degree], [periodic, periodic],[2,2])
```
yguclu, line 313: More readable:
```suggestion
        test_derham_projector_2d_hdiv([nc, nc], [degree, degree], [periodic, periodic], [2, 2])
```

psydac/feec/tests/test_global_projectors.py
```diff
@@ -294,17 +308,20 @@ def test_derham_projector_3d(ncells, degree, periodic):
...
+        test_H1_projector_1d(domain, nc, degree, periodic, multiplicity = 2)
 
     nquads = degree
     for nc in ncells:
         test_L2_projector_1d(domain, nc, degree, periodic, nquads)
         
     for nc in ncells:
         test_derham_projector_2d_hdiv_2([nc, nc], [degree, degree], [periodic, periodic])
-        test_derham_projector_2d_hdiv([nc, nc], [degree, degree], [periodic, periodic])
+        test_derham_projector_2d_hdiv([nc, nc], [degree, degree], [periodic, periodic],2)
```
yguclu, line 324: More readable:
```suggestion
        test_derham_projector_2d_hdiv([nc, nc], [degree, degree], [periodic, periodic], 2)
```

psydac/fem/splines.py
```diff
@@ -178,9 +177,10 @@ def init_interpolation( self, dtype=float ):
             degree   = self.degree,
             periodic = self.periodic,
             normalization = self.basis,
-            xgrid    = self.greville
+            xgrid    = self.greville,
+            multiplicity = self.multiplicity
         )
-
+
```
yguclu, line 183: Please revert this change

psydac/feec/tests/test_global_projectors.py
vcarlier: I don't feel like increasing from 2e-2 to 3e-2 is that big of an increase, but for the case with higher multiplicity the error was something like 2.1e-2 so I increased the error to get those tests passing.

yguclu (approved):

yguclu:
@campospinto Do you have any further comments or questions on this PR?

psydac/core/bsplines_kernels.py
campospinto: Since we discussed the precise meaning of this:
```
    # Knot index at left/right boundary
    low  = degree
    high = len(knots)-1-degree
```
I suggest to specify the comment and write
```
    # last knot on the left boundary
    low  = degree
    # first knot on the right boundary
    high = len(knots)-1-degree
```

campospinto:
Just one minor suggestion ! Otherwise it looks good, thanks !

psydac/core/bsplines_kernels.py
campospinto: thanks !

Issue #332, opened 2023-09-20
Allow different quadrature points for the same `TensorFemSpace`


yguclu:
- Prepared in Julian’s PR #289
- `nquads` should not be provided to `TensorFemSpace` and `discretize_space()`, but only to `discretize_(bi)linearform()`
- Related to #307

Example API
------------

```python
domain = Square('Omega')
V = ScalarFunctionSpace('V', domain)
u, v = elements_of(V, names='u, v')

a = BilinearForm((u,v),  integral(domain, dot(grad(u), grad(v))))
l = LinearForm(v, integral(domain, f * v))  # f is some analytical expression

domain_h = discretize(domain, ncells=nc)
V_h = discretize(V, domain_h, degree=p)                # do not create `FemAssemblyGrid`s here
a_h = discretize(a, domain_h, (V_h, V_h), nquads=p+1)  # create `FemAssemblyGrid`s with (p+1) quad pts
l_h = discretize(l, domain_h, V_h, nquads=p+2)         # create finer `FemAssemblyGrid`s
```

Implementation breakdown 
------------

Modify class `TensorFemSpace` in module `psydac.fem.tensor`:

- [x] Remove `nquads` from [argument list](https://github.com/pyccel/psydac/blob/cf4f4b6cc496457258aa444a92cb20d95af6c523/psydac/fem/tensor.py#L63-L66) of constructor, and do not store `_nquads` private attribute
- [x] Initialize [`self._quad_grids`](https://github.com/pyccel/psydac/blob/cf4f4b6cc496457258aa444a92cb20d95af6c523/psydac/fem/tensor.py#L111-L113) in constructor as an empty dictionary
- [x] Remove property [`nquads`](https://github.com/pyccel/psydac/blob/cf4f4b6cc496457258aa444a92cb20d95af6c523/psydac/fem/tensor.py#L760-L762) and method [`quad_grids`](https://github.com/pyccel/psydac/blob/cf4f4b6cc496457258aa444a92cb20d95af6c523/psydac/fem/tensor.py#L764-L780)
[...]

PR #340, opened 2023-10-02
Fix `DirectionalDerivativeOperator.tosparse` w/ 1 cell & periodic

### LINKED ISSUE #337: cannot convert 3D curl matrix to sparse when using periodic BC and 1 cell along the same direction
The `tosparse()` method of the 3D curl matrix fails when we use periodic BC along a direction with only one cell.

Minimum example:
```
domain = Cube('domain', bounds1=(0, 1), bounds2=(0,1), bounds3=(0,1))
derham = Derham(domain)
domain_h = discretize(domain, ncells=[2, 2, 1], periodic=(False,False,True))
derham_h = discretize(derham, domain_h, degree=[1, 1, 1])
D0, D1, D2 = derham_h.derivatives_as_matrices
D1_m = D1.tosparse() # Fails
```
It does not fail if there is more than one cell or if no periodic BC are imposed.

e-moral-sanchez:
Ensure the directional derivative is correctly converted to sparse matrix when we use one cell and periodic BC along a certain axis. In this case the spline space has dimension 1 and contains only constant functions, and the application of the directional derivative operator should always return zero.

Fixes #337.

### DIFF
--- a/psydac/feec/derivatives.py
+++ b/psydac/feec/derivatives.py
@@ -320,7 +320,7 @@ class DirectionalDerivativeOperator(LinearOperator):
             if self._diffdir == d:
                 # if we are at the differentiation direction, construct differentiation matrix
                 if codomain_local == 1 and domain_local == 1 and self.domain.periods[d]:
-                    # case of one cell and periodic BC should be treated as a constant
+                    # case of one cell and periodic BC should be treated as a constant, thus zero matrix
                     addmatrix = spa.coo_array((codomain_local, domain_local))
 
                 else:
@@ -413,9 +413,8 @@ class Derivative_1D(DiffOperator):
         assert H1.periodic[0] == L2.periodic[0]
         assert H1.degree[0] == L2.degree[0] + 1
 
-        self._domain   = H1
-        self._codomain = L2
-        self._matrix   = DirectionalDerivativeOperator(H1.vector_space, L2.vector_space, 0)
+        # Store data in object   
+        super().__init__(H1, L2, DirectionalDerivativeOperator(H1.vector_space, L2.vector_space, 0))
 
 #====================================================================================================
 class Gradient_2D(DiffOperator):
@@ -453,10 +452,9 @@ class Gradient_2D(DiffOperator):
                   [DirectionalDerivativeOperator(B_B, B_M, 1)]]
         matrix = BlockLinearOperator(H1.vector_space, Hcurl.vector_space, blocks=blocks)
 
-        # Store data in object
-        self._domain   = H1
-        self._codomain = Hcurl
-        self._matrix   = matrix
+        # Store data in object   
+        super().__init__(H1, Hcurl, matrix)
+
 
 #====================================================================================================
 class Gradient_3D(DiffOperator):
@@ -497,10 +495,8 @@ class Gradient_3D(DiffOperator):
                   [DirectionalDerivativeOperator(B_B_B, B_B_M, 2)]]
         matrix = BlockLinearOperator(H1.vector_space, Hcurl.vector_space, blocks=blocks)
 
-        # Store data in object
-        self._domain   = H1
-        self._codomain = Hcurl
-        self._matrix   = matrix
+        # Store data in object   
+        super().__init__(H1, Hcurl, matrix)
 
 #====================================================================================================
 class ScalarCurl_2D(DiffOperator):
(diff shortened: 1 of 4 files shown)


psydac/feec/tests/test_commuting_projections_dual.py
```diff
@@ -0,0 +1,149 @@
...
+
+    v2 = element_of(derham.V2, name='v2')
+    v3 = element_of(derham.V3, name='v3')
+    div = Divergence_3D(derham_h.V2, derham_h.V3)
+
+    f2 = LinearForm(v2, integral(domain, D1fun1(*domain.coordinates)*v2[0] + D2fun1(*domain.coordinates)*v2[1] + D3fun1(*domain.coordinates)*v2[2]))
+    f3 = LinearForm(v3, integral(domain, fun1(*domain.coordinates) * v3))
+
+    u2 = discretize(f2, domain_h, derham_h.V2, backend=PSYDAC_BACKENDS['pyccel-gcc']).assemble()
+    u3 = discretize(f3, domain_h, derham_h.V3, backend=PSYDAC_BACKENDS['pyccel-gcc']).assemble()
```
yguclu, line 46: Please pass the argument `nquads = (N, N, N)` with `N=10` (for example) and check if the error decreases.

If I remember correctly, the same argument should be passed to the discretization of the de Rham sequence. See also #332.

psydac/feec/tests/test_commuting_projections_dual.py
```diff
@@ -0,0 +1,149 @@
...
+
+import numpy as np
+from sympy import sin, cos
+import pytest
+
+
+@pytest.mark.parametrize('Nel', [8, 12])
+@pytest.mark.parametrize('p', [2, 3])
+@pytest.mark.parametrize('bc', [True, False])
+def test_3d_commuting_pro_dual_1(Nel, p, bc):
```
yguclu, line 18: Maybe we can call this function `test_transpose_div_3d`?

psydac/feec/tests/test_differentiation_matrices.py
yguclu, line 1: I am not sure why you test `ncells = (1, 1, 1)` as this is almost never a situation of interest. I would rather test `ncells = (1, 8, 8)`, `ncells = (8, 1, 8)` and `ncells = (8, 8, 1)`, for example.

yguclu (changes requested):

psydac/feec/derivatives.py
```diff
@@ -356,6 +361,16 @@ def copy(self):
 
 #====================================================================================================
 class DiffOperator:
```
campospinto, line 363: Btw, why is DiffOperator not subclassing LinearOperator ?

psydac/feec/derivatives.py
```diff
@@ -319,21 +319,26 @@ def tosparse(self, **kwargs):
...
-
-                # handle special case with not self.domain.parallel and not with_pads and periodic
-                if self.domain.periods[d] and not self.domain.parallel and not with_pads:
-                    # then: add element to other side of the array
-                    adddiagcirc = np.array([sign])
-                    offsets = (-codomain_local+1, 0, 1)
-                    diags = (adddiagcirc, maindiag, adddiag)
+                if codomain_local == 1 and domain_local == 1 and self.domain.periods[d]:
+                    # case of one cell and periodic BC should be treated as a constant
+                    addmatrix = spa.coo_array((codomain_local, domain_local))
```
campospinto, line 324: why call this `addmatrix` since it will be a term in a Kronecker _product_ ?

campospinto: (also for clarity, you can add a comment specifying that this is a zero matrix)

e-moral-sanchez: This is the name that was used for this variable before. I can change it though. Any suggestion?

campospinto: why not `directional_matrix` ?

psydac/feec/derivatives.py
e-moral-sanchez: `DiffOperator` was implemented before Julian's work on `LinearOperator`s. There are things that we would need to fix in order to subclass it, for instance, now `DiffOperator` works with `FemField`s and not `StencilVector`s. However this would be another PR.

campospinto: ok thanks !

psydac/feec/tests/test_differentiation_matrices.py
yguclu: Indeed this is something that we have never used. Setting one of `ncells` to 1 is needed for running a 2D simulation using 3D fields. Setting all of them to 1 is not what Psydac is designed for, especially with periodic BCs.

yguclu: @e-moral-sanchez please write a comment like the following:

Fixed with cc7240f.

And let the reviewer resolve the conversation.

yguclu (approved):
Good job!

PR #349, opened 2023-10-25
Add documentation to DiscreteDerham

vcarlier:
This PR just adds some documentation to the methods and properties of DiscreteDerham.
This was discussed on PR #282 but then was merged before the whole documentation was added


psydac/api/feec.py
```diff
@@ -86,51 +86,88 @@ def __init__(self, mapping, *spaces):
     #--------------------------------------------------------------------------
     @property
     def dim(self):
+        """dimension of the physical and logical space"""
```
campospinto, line 89: I guess you mean: dimension of the physical and logical _domain_ ?

campospinto: and you can specify: (assuming both are of same dimension) since one could have dim(logical_Omega) < dim(physical_Omega)

psydac/api/feec.py
```diff
@@ -86,51 +86,88 @@ def __init__(self, mapping, *spaces):
...
         return self._spaces[2]
 
     @property
     def V3(self):
+        """Fourth space of the De Rham sequence : L2 space in 3d"""
         return self._spaces[3]
 
     @property
     def H1vec(self):
+        """Vectorial H1 space built has cartesian product of V0"""
```
campospinto, line 119: typo: build _as_

campospinto: also please specify: `... cartesian product of V0 n times, with n = dimension of (logical) domain`

psydac/api/feec.py
```diff
@@ -86,51 +86,88 @@ def __init__(self, mapping, *spaces):
...
         return self._H1vec
 
     @property
     def spaces(self):
+        """Spaces of the proper de Rham sequence (excluding Hvec)"""
         return self._spaces
 
     @property
     def mapping(self):
+        """The mapping from the logical space to the physical space of the discrete De Rham."""
```
campospinto, line 130: you can specify `(assuming a single patch parametrization)`

psydac/api/feec.py
```diff
@@ -86,51 +86,88 @@ def __init__(self, mapping, *spaces):
...
         return self._mapping
 
     @property
     def callable_mapping(self):
+        """The mapping as a callable"""
         return self._callable_mapping
 
     @property
     def derivatives_as_matrices(self):
+        """Differential operators of the De Rham sequence as BlockLinearOperator"""
```
campospinto, line 140: `as BlockLinearOperator objects`, and you can specify on which type of objects these operators act

@yguclu is it important that they are of type `BlockLinearOperator`, or just specifying `as LinearOperator` is good enough?

also probably the name `derivatives_as_matrices` is no longer appropriate ?

psydac/api/feec.py
```diff
@@ -86,51 +86,88 @@ def __init__(self, mapping, *spaces):
...
         return tuple(V.diff.matrix for V in self.spaces[:-1])
 
     @property
     def derivatives_as_operators(self):
+        """Differential operators of the De Rham sequence as DiffOperator"""
```
campospinto, line 145: `as DiffOperator objects`, and you can specify on which type of objects these operators act

psydac/api/feec.py
```diff
@@ -86,51 +86,88 @@ def __init__(self, mapping, *spaces):
...
+            Number of quadrature points along each direction, to be used in Gauss
+            quadrature rule for computing the (approximated) degrees of freedom.
+            
+        Returns
+        -------
+        P0, ..., Pn : callables
+            Projectors that can be called on any callable function that maps 
+            from the physical space to R (scalar case) or R^d (vector case) and
+            returns a FemField belonging to the i-th space of the De Rham sequence
+            """
```
campospinto, line 170: side question: can these operators be also called on "pushed-forward explicit functions" ? There was a discussion about that feature some time ago but I don't remember

campospinto (changes requested):
Thanks Valentin this is useful ! I added a few comments and suggestions

psydac/api/feec.py
yguclu: Good point! These objects are not always of type `BlockLinearOperator`; in the 1D case you get a `DirectionalDerivativeOperator` instance. Hence we should write `as LinearOperator objects` as you suggested.

The name is not perfect, but these are the differential operators appearing in the de Rham sequence of the coefficient spaces. We usually refer to them as "matrices" in our publications, and I do not have a better name... 😅

campospinto: Ok ! Just asking :)

psydac/api/feec.py
yguclu: These projectors appear in the primal de Rham sequence on the physical domain. We commonly call them "geometric projection operators".

psydac/api/feec.py
yguclu: @vcarlier These operators act on `FemField` objects, i.e. they take as input a `FemField` belonging to their `domain` `FemSpace`, and return a `FemField` belonging to their `codomain` `FemSpace`.

psydac/api/feec.py
```diff
@@ -86,51 +86,88 @@ def __init__(self, mapping, *spaces):
...
+        
+        nquads : list(int) | tuple(int)
+            Number of quadrature points along each direction, to be used in Gauss
+            quadrature rule for computing the (approximated) degrees of freedom.
+            
+        Returns
+        -------
+        P0, ..., Pn : callables
+            Projectors that can be called on any callable function that maps 
+            from the physical space to R (scalar case) or R^d (vector case) and
```
yguclu, line 168: The underlying field does not have to be necessarily real. It can be complex.

psydac/api/feec.py
yguclu: In other parts of Psydac we use the properties `ldim` for the logical domain and `pdim` for the physical domain. The FEEC part of Psydac always assumes that `pdim == ldim`.

psydac/api/feec.py
yguclu: I would also replace "Vectorial" with "Vector", which is much more common in English.

psydac/api/feec.py
yguclu: I think this can be a multipatch domain too. Is that correct @saidctb?

psydac/api/feec.py
campospinto: (regarding the objects on which they act, I guess we do not need to be very specific -- reminding that they act on `Vectors` is probably good enough)

psydac/api/feec.py
campospinto: that's also good to specify indeed

psydac/api/feec.py
vcarlier: Is it really the place to put all this information? Maybe we should redirect to the documentation of DiffOperator (and create it)?

psydac/api/feec.py
yguclu: Fixed in 8e2067f

psydac/api/feec.py
yguclu: Actually no, class `DiscreteDerham` is only or single patch domains. Its multipatch counterpart is `MultipatchDiscreteDerham` is in `psydac.feec.multipatch.api`.

@vcarlier  please add a comment about it.

psydac/api/feec.py
campospinto: so I suggest: 
```
"""Differential operators of the De Rham sequence as LinearOperator (acting on coefficient vectors)"""
```

psydac/api/feec.py
yguclu: Let's change the name of this function to `derivatives_as_fem_operators`. There is no need to mention class `DiffOperator` as this feels like an internal implementation detail to me.

psydac/api/feec.py
```diff
@@ -86,51 +86,90 @@ def __init__(self, mapping, *spaces):
...
         return self._spaces[2]
 
     @property
     def V3(self):
+        """Fourth space of the De Rham sequence : L2 space in 3d"""
         return self._spaces[3]
 
     @property
     def H1vec(self):
+        """Vector H1 space built as cartesian product of V0 n times with n = dimension of (logical) domain"""
```
campospinto, line 119: actually I would specify `Vector-valued H1 space ... `

psydac/api/feec.py
yguclu: Actually maybe we could just call it `derivatives`, which is consistent with the `projectors` method.

psydac/api/feec.py
yguclu: > side question: can these operators be also called on "pushed-forward explicit functions" ? There was a discussion about that feature some time ago but I don't remember

Not yet. But we could modify this after SymPDE's PR [#126](https://github.com/pyccel/sympde/pull/126)

yguclu: The brief description could be
"Projectors mapping callable functions of the physical coordinates to a corresponding `FemField` object in the De Rham sequence."

psydac/api/feec.py
vcarlier: solved in [49ed53e](https://github.com/pyccel/psydac/pull/349/commits/49ed53eec0a2ac1e5117de0e818720027b2a6e36)

psydac/api/feec.py
vcarlier: Done in [49ed53e](https://github.com/pyccel/psydac/pull/349/commits/49ed53eec0a2ac1e5117de0e818720027b2a6e36)

psydac/api/feec.py
vcarlier: Called derivatives in [49ed53e](https://github.com/pyccel/psydac/pull/349/commits/49ed53eec0a2ac1e5117de0e818720027b2a6e36) 
 Looks like this method is anyway never used in the code at this point.

psydac/api/feec.py
vcarlier: Done in [49ed53e](https://github.com/pyccel/psydac/pull/349/commits/49ed53eec0a2ac1e5117de0e818720027b2a6e36)

psydac/api/feec.py
vcarlier: Done in [49ed53e](https://github.com/pyccel/psydac/pull/349/commits/49ed53eec0a2ac1e5117de0e818720027b2a6e36)

psydac/api/feec.py
campospinto: you mean: in not a single branch 😅 ?

vcarlier: I only scanned devel.
But the multipatch counterpart seems to be used in some multipatch tests.

campospinto (approved):
Ok

campospinto:
Thanks Yaman for these commits 👍

yguclu:
> Thanks Yaman for these commits 👍

If only I had not introduced a bug! 😅

Issue #353, opened 2023-10-27
Avoid deprecation warnings


yguclu:
NumPy warning:
> DeprecationWarning: `product` is deprecated as of NumPy 1.25.0, and will be removed in NumPy 2.0. Please use `prod` instead.

SymPy warning:
> SymPyDeprecationWarning: 
>  
>  non-Expr objects in a Matrix has been deprecated since SymPy 1.9. Use
>  list of lists, TableForm or some other data structure instead. See
>  https://github.com/sympy/sympy/issues/21497 for more info.

kvrigor:
With #355 merged, should we close this issue?

yguclu:
> With #355 merged, should we close this issue?

No, we cannot, because #355 did not address the most serious warning, which is the `ComplexWarning` raised by NumPy.

PR #355, opened 2023-11-02
Avoid runtime warnings

yguclu:
Avoid runtime warnings from third-party libraries:

- Use Numpy `prod()` instead of deprecated `product()`
- Do not put non-`Expr` objects inside SymPy `Matrix` -- this is obtained by using 3D `MutableDenseNDimArray` objects containing scalars, instead of 2D `Matrix` objects containing tuples

TO DO
-------
Do not cast complex values to real during post-processing (which will fix issue #353)


campospinto (approved):
Looks good, no particular comments from my side !

PR #357, opened 2023-11-28
Add optional `out` argument for `transpose` and `copy` in `StencilMatrix` and `BlockLinearOperator`

vcarlier:
This PR implements the possibility to pass an `out` argument to the methods `copy` and `transpose` of classes `StencilMatrix` and `BlockLinearOperator`.

This is specially usefull to avoid creating temporaries when one copy/transpose frequently Operator for example in the case of iterative algorithm that update a matrix at every step.

Some tests were added to check the implementation.

### DIFF
--- a/psydac/api/ast/parser.py
+++ b/psydac/api/ast/parser.py
@@ -541,19 +541,12 @@ class Parser(object):
         body = tuple(inits) + body
         name = expr.name
 
-        # TODO : when Pyccel will work on the cmath library, we should import the math function from cmath and not from numpy
-        # If we are with complex object, we should import the mathematical function from numpy and not math to handle complex value.
-        if expr.domain_dtype=='complex':
-            numpy_imports = ('array', 'zeros', 'zeros_like', 'floor', *self._math_functions)
-            imports       = [Import('numpy', numpy_imports)] + \
-                            [*expr.imports]
-        # Else we import them from math
-        else:
-            math_imports  = (*self._math_functions,)
-            numpy_imports = ('array', 'zeros', 'zeros_like', 'floor')
-            imports       = [Import('numpy', numpy_imports)] + \
-                            ([Import('math', math_imports)] if math_imports else []) + \
-                            [*expr.imports]
+        math_library  = 'cmath' if expr.domain_dtype=='complex' else 'math' # Function names are the same
+        math_imports  = (*self._math_functions,)
+        numpy_imports = ('array', 'zeros', 'zeros_like', 'floor')
+        imports       = [Import('numpy', numpy_imports)] + \
+                        ([Import(math_library, math_imports)] if math_imports else []) + \
+                        [*expr.imports]
 
         results = [self._visit(a) for a in expr.results]
 
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -43,27 +43,37 @@ from psydac.mapping.discrete import NurbsMapping
 from psydac.linalg.stencil   import StencilVectorSpace
 from psydac.linalg.block     import BlockVectorSpace
 
-__all__ = ('discretize', 'discretize_derham', 'reduce_space_degrees', 'discretize_space', 'discretize_domain')
-
+__all__ = (
+    'discretize',
+    'discretize_derham',
+    'reduce_space_degrees',
+    'discretize_space',
+    'discretize_domain'
+)
 
 #==============================================================================
 def change_dtype(V, dtype):
     """
-    This function take a FemSpace V and create a new vector_space for it with the data type required.
+    Given a FemSpace V, change its underlying vector_space (i.e. the space of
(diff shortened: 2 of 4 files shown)


yguclu:
The `transpose` method is an `abstractmethod` of the base class `LinearOperator` (see [here](https://github.com/pyccel/psydac/blob/535717c6f5ea328aacbbbbcc2d582a92b31c9377/psydac/linalg/basic.py#L237-L244)).

A question therefore arises: should we add the optional `out` argument to `LinearOperator`?

CC: @jowezarek

vcarlier:
I think it makes sense to add an `out` option to the `copy` and `transpose` method for all `LinearOperator`.
It is something I implemented in my branch because without this possibility, in the case of iterative algorithm I was not even able to run the simulation because of `out of memory error`.

yguclu:
Currently `copy` is **not** part of the abstract interface of the `LinearOperator` base class. Should we add it?

yguclu:
We can modify the base class `LinearOperator` in a separate PR.

For now please check that the new code works in parallel by adding the `out` argument to `linalg.tests.test_stencil_matrix_2d_parallel_transpose`

vcarlier:
> We can modify the base class `LinearOperator` in a separate PR.
> 
> For now please check that the new code works in parallel by adding the `out` argument to `linalg.tests.test_stencil_matrix_2d_parallel_transpose`.
> 
> Please also create a parallel test in `linalg.tests.test_block`.

Test created in [715eb4a](https://github.com/pyccel/psydac/pull/357/commits/715eb4a8e53df83e7f12e247021a1fd4e36ea40f)

psydac/linalg/tests/test_stencil_matrix.py
```diff
@@ -2569,12 +2569,17 @@ def test_stencil_matrix_2d_parallel_transpose(dtype, n1, n2, p1, p2, sh1, sh2, P
...
+    To = StencilMatrix(V, V)
+    M.transpose(out=To)
+    Tos = To.tosparse()
 
     # Exact result: convert to Scipy sparse format, then transpose
     Ts_exact = M.tosparse().transpose()
 
     # Exact result: convert to Scipy sparse format including padding, then
     # transpose, hence remove entries that do not belong to current process.
+    # V.C 29/11/2023 : Is there a reason to compute twice Ts_exact?
```
yguclu, line 2582: Not that I am aware of. Maybe the line above was copied from another test.

psydac/linalg/tests/test_block.py
```diff
@@ -989,6 +989,34 @@ def test_block_linear_operator_parallel_dot( dtype, n1, n2, p1, p2, P1, P2 ):
     # Check data in 1D array
     assert np.allclose( Y.blocks[0].toarray(), y1.toarray(), rtol=1e-14, atol=1e-14 )
     assert np.allclose( Y.blocks[1].toarray(), y2.toarray(), rtol=1e-14, atol=1e-14 )
+
+    # Test copy with an out 
+    K = BlockLinearOperator( W, W )
+    L.copy(out=K)
```
yguclu, line 995: I am not convinced that this is the test we want. The `K` object is created completely empty, with no blocks in its dictionary. Hence when we copy `L` into `K` we are creating all new `StencilMatrix` blocks instead of writing into preexisting blocks.

psydac/linalg/tests/test_block.py
```diff
@@ -989,6 +989,34 @@ def test_block_linear_operator_parallel_dot( dtype, n1, n2, p1, p2, P1, P2 ):
...
+    # Compute Block-vector product
+    K.dot(X, out= Y)
+
+    # Check data in 1D array
+    assert np.allclose( Y.blocks[0].toarray(), y1.toarray(), rtol=1e-14, atol=1e-14 )
+    assert np.allclose( Y.blocks[1].toarray(), y2.toarray(), rtol=1e-14, atol=1e-14 )
+
+    # Test transpose with an out
+    N = BlockLinearOperator( W, W )
+    L.transpose(out = N)
```
yguclu, line 1006: Same comment. Maybe we need to write into `N` (and `K`) at least twice in order to properly test the `out` argument?

vcarlier: Is what I did in [71f2d07](https://github.com/pyccel/psydac/pull/357/commits/71f2d07fad81f77597e52e417c8707c1b93661a7) what you expected?

yguclu: Yes! The test is much stronger now!

yguclu:
Apparently the parallel implementation needs a bit more work! 😅

vcarlier:
> Apparently the parallel implementation needs a bit more work! 😅

No I just forgot to add padding argument in one test 😅.
Everything is working now, I think we are ready for merge

yguclu (approved):

PR #376, opened 2024-02-07
Fix GFortran compiler flags on ARM processors

yguclu:
- Give the flag `-mavx` to the GFortran compiler only if the processor's architecture is `x86_64`.
- Clean up the backends settings in `psydac.api.settings`.
- Update `.gitignore`.

### DIFF
--- a/psydac/api/settings.py
+++ b/psydac/api/settings.py
@@ -13,7 +13,7 @@ PSYDAC_BACKEND_PYTHON = {'name': 'python', 'tag':'python', 'openmp':False}
 
 PSYDAC_BACKEND_GPYCCEL  = {'name': 'pyccel',
                        'compiler': 'GNU',
-                       'flags'   : '-O3 -march=native -mtune=native -ffast-math -ffree-line-length-none',
+                       'flags'   : '-O3 -march=native -mtune=native -ffast-math',
                        'folder'  : '__gpyccel__',
                        'tag'     : 'gpyccel',
                        'openmp'  : False}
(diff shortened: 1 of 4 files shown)


psydac/api/settings.py
```diff
@@ -1,39 +1,38 @@
...
+
+PSYDAC_BACKEND_IPYCCEL  = {'name': 'pyccel',
+                       'compiler': 'intel',
+                       'flags'   : '-O3',
+                       'folder'  : '__ipyccel__',
+                       'tag'     :'ipyccel',
+                       'openmp'  : False}
+
+PSYDAC_BACKEND_PGPYCCEL = {'name': 'pyccel',
+                       'compiler': 'PGI',
```
EmilyBourne, line 25: PGI is a legacy compiler whose last community release was in 2019. As PGI compilers and tools have evolved into the NVIDIA HPC SDK, it might be a good idea to add a nvidia option.

yguclu: Good idea! Should we remove the option from Pyccel too? See [here](https://github.com/pyccel/pyccel/blob/b03c874705c0b45a3037fd7536a6e2e283094fd7/pyccel/compilers/default_compilers.py#L65-L80).

EmilyBourne: I think we could consider it. As we move to version 2.0 it is a sensible time to make breaking changes like this. That said it is not difficult to support both as the only difference is the name. The reason I mentioned it here was that you are only supporting the legacy version not the supported version

yguclu: I've added the Nvidia compiler in e2df18f.

PR #381, opened 2024-02-27
Conversion to PETSc with complex

e-moral-sanchez:
Fix the functions to cast Psydac format to PETSc format and back:

- Identify the correct installation procedure for PETSc and `petsc4py`;
- Make functions work on complex-valued vectors and matrices;
- Make functions work in the serial case (where `comm=None`);
- Add more extensive unit testing;
- Update unit test workflow;
- Update README file.

This allows the use of high-performance PETSc solvers:

- PETSc must be installed with the complex configuration (see the instructions in the README).
- `StencilVector`, `BlockVector`, `StencilMatrix` and `BlockLinearOperator` have a `topetsc()` method.
- The function `petsc_to_psydac` in `psydac.linalg.utilities.py` allows us to convert a `PETSc.Vec` object (for instance, the vector solution after solving a system with PETSc) back to Psydac `StencilVector` or `BlockVector` format.

How to use it: 
Let `A` be a `StencilMatrix` or `BlockLinearOperator` and b a `StencilVector`. We want to solve `Ax=b` with `PETSc` CG:

```
from psydac.linalg.utilities import petsc_to_psydac
from petsc4py import PETSC

Ap = A.topetsc() #converts operator to PETSc.Mat format

bp = b.topetsc() #converts vector to PETSc.Vec format

x = bp.duplicate() #x will be both the initial guess and where the solution is stored.
# If we have a particular initial guess in Psydac format, just convert it to PETSc.Vec format x = x.topetsc()

ksp = PETSc.KSP() # create Krylov solver type object
ksp.create(comm=Ap.comm) # pass communicator from PETSc.Mat object
opts = PETSc.Options() # create object to pass options to solver
opts["ksp_type"] = "cg" # set type of solver ("gmres" is default)
opts["pc_type"] = "none" # set preconditioner
opts["ksp_rtol"] = 1e-13
opts["ksp_atol"] = 1e-13
ksp.setFromOptions() # pass options to solver ksp
ksp.setOperators(Ap) # set system matrix. Preconditioner can also be passed here

ksp.solve(bp, x) # solve system

sol = petsc_to_psydac(x, V.vector_space) # convert solution to Psydac StencilVector format
```
Note: With this PR, the conversion from PETSC format to StencilVector format is not efficient, see [issue #389](https://github.com/pyccel/psydac/issues/389).

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -105,7 +105,7 @@ jobs:
       - name: Install petsc4py
         working-directory: ./petsc
         run: |
-          python -m pip install wheel
+          python -m pip install wheel numpy
           PETSC_DIR=$(pwd) pip install --no-deps src/binding/petsc4py
 
       - name: Install Python dependencies
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -106,8 +106,8 @@ jobs:
       - name: Alternative install of PETSc and petsc4py
         run: |
           export PETSC_CONFIGURE_OPTIONS="--with-scalar-type=complex --with-fortran-bindings=0 --have-numpy=1 --with-petsc4py=1 --with-64-bit-indices=1"
-          python -m pip install wheel
-          python -m pip install petsc
+          python -m pip install wheel numpy
+          python -m pip install petsc --verbose
 
       - name: Install Python dependencies
         run: |
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -107,7 +107,7 @@ jobs:
         run: |
           export PETSC_CONFIGURE_OPTIONS="--with-scalar-type=complex --with-fortran-bindings=0 --have-numpy=1 --with-petsc4py=1 --with-64-bit-indices=1"
           python -m pip install build numpy
-          python -m pip install petsc --verbose
+          python -m pip install petsc --verbose --no-binary=petsc
           python -m pip list
 
       - name: Install Python dependencies
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -106,7 +106,7 @@ jobs:
       - name: Install petsc4py
         working-directory: ./petsc
         run: |
-          PETSC_ARCH=linux-gnu-complex-64 python -m pip install /src/binding/petsc4py/setup.py
+          PETSC_ARCH=linux-gnu-complex-64 PETSC_DIR=~/petsc-main python -m pip install src/binding/petsc4py/setup.py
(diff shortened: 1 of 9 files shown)


saidctb (approved):

.github/workflows/continuous-integration.yml
```diff
@@ -105,6 +105,7 @@ jobs:
       - name: Install petsc4py
         working-directory: ./petsc
         run: |
+          python -m pip install Cython
           PETSC_ARCH=linux-gnu-complex-64 python -m pip install src/binding/petsc4py
```
e-moral-sanchez, line 109: I think we need to add here `PETSC_DIR=~/petsc-main`

yguclu: We are working in the directory `./petsc` here. I will try anyway

yguclu: The installation works after 0d589a6

.github/workflows/continuous-integration.yml
```diff
@@ -88,6 +88,26 @@ jobs:
...
+          PETSC_VERSION="v3.20.5"
+          wget https://gitlab.com/petsc/petsc/-/archive/$PETSC_VERSION/petsc-$PETSC_VERSION.zip
+          unzip petsc-$PETSC_VERSION.zip
+          rm petsc-$PETSC_VERSION.zip
+          mv petsc-$PETSC_VERSION petsc
+
+      - name: Install PETSc with complex support, and test it
+        working-directory: ./petsc
+        run: |
+          ./configure --with-scalar-type=complex --with-fortran-bindings=0
```
e-moral-sanchez, line 102: Just as a remark, external packages like MUMPS use Fortran (https://petsc.org/release/install/install_tutorial/), which requires an installation without the flag `--with-fortran-bindings=0`.

.github/workflows/continuous-integration.yml
```diff
@@ -88,6 +88,21 @@ jobs:
...
+          unzip petsc-$PETSC_VERSION.zip
+          rm petsc-$PETSC_VERSION.zip
+          mv petsc-$PETSC_VERSION petsc
+
+      - name: Install PETSc and petsc4py with complex support, and test it
+        working-directory: ./petsc
+        run: |
+          ./configure --with-scalar-type=complex --with-fortran-bindings=0 --have-numpy=1 --with-petsc4py=1 --with-64-bit-indices=1
+          make all check
+          python3 -c "import sys, os; sys.path.append(os.getcwd() + '/arch-linux-c-debug/lib')"
```
yguclu, line 104: `arch-linux` will not work on macOS

yguclu: Modifying `sys.path` here is local to the current Python process, which ends after the call. It will have no effect on future calls to Python.

yguclu (approved):
Good job!

PR #388, opened 2024-03-13
Fix bug in `copy` method of `BlockLinearOperator`

jowezarek:
Add the missing statement `return out` to the method `copy` of the class `BlockLinearOperator`. Also handle the corner case of `out=self`.

### DIFF
--- a/psydac/linalg/block.py
+++ b/psydac/linalg/block.py
@@ -938,6 +938,11 @@ class BlockLinearOperator(LinearOperator):
         ----------
         out : BlockLinearOperator(optional)
             The existing BlockLinearOperator in which we want to copy self.
+
+        Returns
+        -------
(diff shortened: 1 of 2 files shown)


psydac/linalg/block.py
```diff
@@ -953,6 +953,8 @@ def copy(self, out=None):
                 Lij.copy(out=out[i,j])
 
         out.set_backend(self._backend)
+
+        return out
```
yguclu, line 957: Please add the `Returns` section to this method's docstring.

yguclu (changes requested):
Good catch!

psydac/linalg/block.py
jowezarek: 5f47f54a59842a945a2c31611a6700e20e8aa920

PR #392, opened 2024-03-20
Fix several errors in `cell_index_p`

campospinto:
In `cell_index_p` we look for cells containing points. If a point is outside the domain exactly at `tol` distance from the bounds, then the search loop will be stuck in infinite recursion. One way to fix this is to lower the tolerance when checking if the points are within the domain (before actually entering the loop).

**Edit (21.03.2024):**

in this PR I also fix the `status` value returned by the cell search in `cell_index_p`, so that it is:
- 0 if all points are in domain
- -1 if some points are outside domain as expected by calling `cell_index` (was 1 before)
- -2 if search can't decide

in the third case the previous code was actually caught in an infinite loop. To avoid this behaviour I added a max number of iterations in `cell_index_p`

(and a unit test)

### DIFF
--- a/psydac/core/bsplines_kernels.py
+++ b/psydac/core/bsplines_kernels.py
@@ -1097,7 +1097,6 @@ def cell_index_p(breaks: 'float[:]', i_grid: 'float[:]', tol: float, out: 'int[:
             i_cell = (low + high)//2
             it += 1
         if it >= max_it:
-            print("Warning: binary search loop is too long in cell_index_p -- aborted")
             return -2
         
         # Check were we landed with the binary search
--- a/psydac/core/bsplines_kernels.py
+++ b/psydac/core/bsplines_kernels.py
@@ -1077,8 +1077,8 @@ def cell_index_p(breaks: 'float[:]', i_grid: 'float[:]', tol: float, out: 'int[:
     nbk = len(breaks)
 
     # Check if there are points outside the domain
-    if np.min(i_grid) < breaks[0] - tol/2: return 1
-    if np.max(i_grid) > breaks[nbk - 1] + tol/2: return 1
+    if np.min(i_grid) < breaks[0] - tol/2: return -1
+    if np.max(i_grid) > breaks[nbk - 1] + tol/2: return -1
 
     current_index = 0
     while current_index < nx:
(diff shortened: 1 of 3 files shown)


yguclu (commented):
Looks good!

Even better with a unit test... 😄

campospinto:
I have a "test" that triggers the infinite loop. Question: how do I write a unit test which fails (quickly) when the infinite loop is triggered ?

campospinto:
maybe a better solution is to fix a max nb of iterations in the search. This would also have saved me some time when I got this bug...

campospinto:
btw there is another bug: in `bspline.cell_index` we have 
```
    status = cell_index_p(breaks, i_grid, tol, out)
    if status == -1:
        raise ValueError("Encountered a point that was outside of the domain")
```
but `cell_index_p` returns `1` when a point is seen outside the domain...

psydac/core/bsplines_kernels.py
yguclu, line 1: Since we are looking into the function `cell_index_p`, we might as well check its algorithm overall...

Are we using the fact that the array `i_grid` is sorted? It does not look so, because for each point in it we set `low = 0`.

By the way, the goal of this function is essentially equivalent to `numpy.searchsorted`. If I am correct, we could use that as a reference in our tests.

campospinto: the algorithm seemed ok to me, but yes it's better to check that it's correct...

campospinto: I added a comparison with numpy.searchsorted

campospinto: (note: I used a random grid for the unit test. Thinking again this does not sound like a great idea...)

yguclu: Indeed, we cannot be sure that our algorithm will behave exactly like `numpy.searchsorted` in those corner cases... Better to create a grid manually.

campospinto: Done (I was also worried about non-reproducibility)

psydac/core/bsplines_kernels.py
```diff
@@ -1087,13 +1087,19 @@ def cell_index_p(breaks: 'float[:]', i_grid: 'float[:]', tol: float, out: 'int[:
...
+        while it < max_it and (x < breaks[i_cell] - tol or x >= breaks[i_cell + 1] + tol):
             if x < breaks[i_cell]:
                 high = i_cell
             else:
                 low = i_cell
             i_cell = (low + high)//2
-
+            it += 1
+        if it >= max_it:
+            print("Warning: binary search loop is too long in cell_index_p -- aborted")
```
yguclu, line 1100: We already raise a `ValueError` in the calling function, hence I see limited value in this print statement. I suggest to remove it.

campospinto: Indeed, I wrote it before changing the return value, now it is useless.

yguclu (approved):

PR #393, opened 2024-03-21
Fix `None` value in `Pushforward`

campospinto:
there was a small bug in the Pushforward function, when no `grid_local` was passed for a mapping of `Mapping` type

I also added a unit test

### DIFF
--- a/psydac/feec/pushforward.py
+++ b/psydac/feec/pushforward.py
@@ -100,7 +100,10 @@ class Pushforward:
         self.npts_per_cell = npts_per_cell
         self.cell_indexes = cell_indexes
         self.grid_type=grid_type
-        grid_local=grid_local
+        if grid_local is None:
+            grid_local=grid
+        else:
+            grid_local=grid_local
 
         if isinstance(mapping, Mapping):
             self._mesh_grids = np.meshgrid(*grid_local, indexing='ij', sparse=True)
--- a/psydac/feec/pushforward.py
+++ b/psydac/feec/pushforward.py
@@ -102,8 +102,6 @@ class Pushforward:
         self.grid_type=grid_type
         if grid_local is None:
             grid_local=grid
-        else:
-            grid_local=grid_local
 
         if isinstance(mapping, Mapping):
             self._mesh_grids = np.meshgrid(*grid_local, indexing='ij', sparse=True)
--- a/psydac/feec/tests/test_pushforward.py
+++ b/psydac/feec/tests/test_pushforward.py
@@ -1,0 +1,40 @@
+import numpy as np
+import pytest
+
+from psydac.api.discretization  import discretize
+from sympde.topology            import ScalarFunctionSpace
+from sympde.topology            import Square
+from sympde.topology            import Mapping
+from psydac.mapping.discrete    import SplineMapping
+from psydac.feec.pushforward    import Pushforward
+
+def test_basic_call():
+
+    logical_domain = Square('Omega')
+
+    ncells = [2**4, 2**4]
+    degree = [2, 2]
+
+    # Mapping and physical domain
+    class CollelaMapping2D(Mapping):
+
+        _ldim = 2
+        _pdim = 2
+        _expressions = {'x': 'a * (x1 + eps / (2*pi) * sin(2*pi*x1) * sin(2*pi*x2))',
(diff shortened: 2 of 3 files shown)


psydac/feec/pushforward.py
```diff
@@ -100,7 +100,10 @@ def __init__(
         self.npts_per_cell = npts_per_cell
         self.cell_indexes = cell_indexes
         self.grid_type=grid_type
-        grid_local=grid_local
+        if grid_local is None:
+            grid_local=grid
+        else:
+            grid_local=grid_local
```
yguclu, line 106: These two lines do not seem necessary at all.

The original `grid_local = grid_local` was also useless of course 😉 
```suggestion
```

psydac/feec/tests/test_pushforward.py
```diff
@@ -0,0 +1,40 @@
...
+    hat_V0_h = discretize(hat_V0, domain_h, degree = degree)
+    
+    grid_x1 = hat_V0_h.breaks[0]
+    grid_x2 = hat_V0_h.breaks[1]
+
+    F = SplineMapping.from_mapping(hat_V0_h, mapping.get_callable_mapping())    
+    Pushforward(grid=(grid_x1, grid_x2), mapping=F, grid_type=0)
+
+    F = mapping   
+    Pushforward(grid=(grid_x1, grid_x2), mapping=F, grid_type=0)
```
yguclu, line 40: Unit tests should have an `assert` statement. What are you testing here exactly?

Maybe it is time to merge the modules `feec.pushforward` (which does not have any dedicated unit tests) and `feec.pull_push` (which has lots of tests). We should probably create a dedicated issue about that.

yguclu (changes requested):
Good catch!

AFAIK the `Pushforward` class is only [used](https://github.com/pyccel/psydac/blob/cf4f4b6cc496457258aa444a92cb20d95af6c523/psydac/api/postprocessing.py#L2113-L2122) by the `PostProcessManager` class (which always passes `grid_local` to it) and nowhere else in Psydac. So it is only tested indirectly in file [test_postprocessing.py](https://github.com/pyccel/psydac/blob/devel/psydac/api/tests/test_postprocessing.py). So how did you find this bug?

psydac/feec/pushforward.py
campospinto: ahah indeed

psydac/feec/tests/test_pushforward.py
campospinto: the original code would raise a type error somewhere, so no assert was needed to make it fail...

campospinto:
> Good catch!
> 
> AFAIK the `Pushforward` class is only [used](https://github.com/pyccel/psydac/blob/cf4f4b6cc496457258aa444a92cb20d95af6c523/psydac/api/postprocessing.py#L2113-L2122) by the `PostProcessManager` class (which always passes `grid_local` to it) and nowhere else in Psydac. So it is only tested indirectly in file [test_postprocessing.py](https://github.com/pyccel/psydac/blob/devel/psydac/api/tests/test_postprocessing.py). So how did you find this bug?

It is used in the code for conga polar splines

psydac/feec/tests/test_pushforward.py
campospinto: what kind of assert would you use, if we are just checking that the code runs ?

yguclu: In general I would check that the `Pushforward` class does what it is intended to do, and have some cases where the code was crashing before this PR.

yguclu: The real issue is that we have no proper unit tests for this class, as `test_postprocessing.py` is not extensive enough. The same appears to be true also for `pull_push.py`. I am creating an issue for this.

yguclu: See issue #395

campospinto: you are absolutely right, but here I just wanted to fix one bug :)

yguclu (approved):

yguclu:
@campospinto I have increased the tolerance in the failing unit test. Hopefully our workflows will pass now.

yguclu (approved):

PR #394, opened 2024-03-21
Add 1D field evaluation kernels

wbarham:
The ``eval_fields`` method of class `TensorFemSpace` raises a ``NotImplementedError`` when called to evaluate fields in 1D. To fix this, the missing 1D kernels have been added, and they are now called when needed.

List of changes
--------------

- Add kernel functions ``eval_fields_1d_no_weights``, ``eval_fields_1d_irregular_no_weights``, ``eval_fields_1d_weighted``, and ``eval_fields_1d_irregular_weighted`` to ``psydac/core/field_evaluation_kernels.py``. 
- Modify functions ``eval_fields_irregular_tensor_grid`` and ``eval_fields_regular_tensor_grid`` in ``psydac/fem/tensor.py`` to call these kernel functions in the case of 1D fields.
- Add several 1D tests to ``test_regular_evaluations`` in ``psydac/core/tests/test_field_evaluation_kernel.py``.

Note
-------
An evaluation kernel for the L2 pushforward in 1D was not added. I am not sure if this is needed as mappings are not likely to be needed in 1D problems. Therefore, I refrained from adding this functionality in this PR.

### DIFF
--- a/psydac/core/field_evaluation_kernels.py
+++ b/psydac/core/field_evaluation_kernels.py
@@ -316,6 +316,51 @@ def eval_fields_2d_irregular_no_weights(np1: int, np2: int, f_p1: int, f_p2: int
                     out_fields[i_p_1, i_p_2, :] += spline * coeff_fields
 
 
+@template(name='T', types=['float[:,:]', 'complex[:,:]'])
+def eval_fields_1d_irregular_no_weights(np1: int, f_p1: int,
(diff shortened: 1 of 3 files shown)


yguclu (changes requested):
Good job! I have a couple of comments

yguclu (approved):

Issue #219, opened 2022-05-10
Missing requirement `h5py`?


spossann:
In a pristine virtual environment I tried installing psydac dependencies via
```
export CC="mpicc"
export HDF5_MPI="ON"
export HDF5_DIR=/path/to/hdf5/openmpi
python3 -m pip install -r requirements.txt
python3 -m pip install -r requirements_extra.txt --no-build-isolation
```
At the stage `python3 -m pip install -r requirements.txt` I get the error:
```
ERROR: Command errored out with exit status 1:
   command: /home/spossann/git_repos/psydac/env_psy/bin/python3 /tmp/tmpxl1tl5bp build_wheel /tmp/tmpvflrk0au                                                                                
       cwd: /tmp/pip-install-cnmon8nq/h5py                                                                                                                                                   
  Complete output (71 lines):                                                                                                                                                                
  running bdist_wheel                                                                                                                                                                        
  running build                                                                                                                                                                              
  running build_py                                                                                                                                                                           
  creating build                                                                                                                                                                             
  creating build/lib.linux-x86_64-cpython-38                                                                                                                                                 
  creating build/lib.linux-x86_64-cpython-38/h5py                                                                                                                                            
  copying h5py/ipy_completer.py -> build/lib.linux-x86_64-cpython-38/h5py                                                                                                                    
  copying h5py/version.py -> build/lib.linux-x86_64-cpython-38/h5py                                                                                                                          
  copying h5py/__init__.py -> build/lib.linux-x86_64-cpython-38/h5py                                                                                                                         
  copying h5py/h5py_warnings.py -> build/lib.linux-x86_64-cpython-38/h5py                                                                                                                    
  creating build/lib.linux-x86_64-cpython-38/h5py/_hl                                                                                                                                        
  copying h5py/_hl/vds.py -> build/lib.linux-x86_64-cpython-38/h5py/_hl                                                                                                                      
  copying h5py/_hl/filters.py -> build/lib.linux-x86_64-cpython-38/h5py/_hl                                                                                                                  
  copying h5py/_hl/datatype.py -> build/lib.linux-x86_64-cpython-38/h5py/_hl                                                                                                                 
  copying h5py/_hl/dims.py -> build/lib.linux-x86_64-cpython-38/h5py/_hl                                                                                                                     
  copying h5py/_hl/base.py -> build/lib.linux-x86_64-cpython-38/h5py/_hl                                                                                                                     
[...]

PR #399, opened 2024-04-03
Provide alternative shell command for discovering HDF5_DIR

kvrigor:
On Ubuntu 22.04, `h5pcc -showconfig` gives the wrong installation point required by `h5py`:

```sh

            SUMMARY OF THE HDF5 CONFIGURATION
            =================================

General Information:
-------------------
                   HDF5 Version: 1.10.7
                  Configured on: Wed, 08 Dec 2021 23:33:27 +0000
                  Configured by: Debian
                    Host system: x86_64-pc-linux-gnu
              Uname information: Debian
                       Byte sex: little-endian
             Installation point: /usr
                    Flavor name: openmpi
```

`pip3 install h5py` will look for `libhdf5.so` which is at `/usr/lib/x86_64-linux-gnu/hdf5/openmpi`, not `/usr`. On Ubuntu/Debian one could directly extract `HDF5_DIR` with the command

```sh
export HDF5_DIR=$(dpkg -L libhdf5-openmpi-dev | grep libhdf5.so | xargs dirname)
```

I wonder if we should still suggest `h5pcc -showconfig` which doesn't really work and may trip users (e.g. #219). The command above may look "hacky", but I believe its robust and would still work for future Ubuntu/Debian releases unless `libhdf5-openmpi-dev` has been renamed—in such case the README has to be updated anyway to change the parallel HDF5 library name for the `apt install` step.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -82,7 +82,7 @@ jobs:
       - name: Determine directory of parallel HDF5 library
         run: |
           if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
-            HDF5_DIR=/usr/lib/x86_64-linux-gnu/hdf5/openmpi
+            HDF5_DIR=$(dpkg -L libhdf5-openmpi-dev | grep libhdf5.so | xargs dirname)
           elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
             HDF5_DIR=$((h5pcc -showconfig -echo || true) | grep "Installation point:" | cut -d ":" -f2 | tr -d " ")
           fi
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -84,7 +84,7 @@ jobs:
           if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
             HDF5_DIR=$(dpkg -L libhdf5-openmpi-dev | grep libhdf5.so | xargs dirname)
           elif [[ "${{ matrix.os }}" == "macos-latest" ]]; then
-            HDF5_DIR=$((h5pcc -showconfig -echo || true) | grep "Installation point:" | cut -d ":" -f2 | tr -d " ")
+            HDF5_DIR=$(brew list hdf5-mpi | grep "libhdf5.dylib" | xargs dirname)
           fi
           echo $HDF5_DIR
           echo "HDF5_DIR=$HDF5_DIR" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


yguclu (changes requested):
The changes look good to me! Thanks @kvrigor!

If the `h5pcc` command does not provide the correct information, I suggest that we remove it from the README.

yguclu:
@pyccel/psydac-dev If you use a Ubuntu/Debian OS, can you please verify that the new command works for you?

kvrigor:
> If the h5pcc command does not provide the correct information, I suggest that we remove it from the README.

Done 👍

Also I've applied the `HDF5_DIR` command to [continuous-integration.yml](https://github.com/pyccel/psydac/pull/399/files#diff-c471496bcb3ed47397b73d2e3e3aa41a8a679c86accaed9e05676fd9d5987434R85). So far it could parse the path correctly.

yguclu (approved):

yguclu:
> > If the h5pcc command does not provide the correct information, I suggest that we remove it from the README.
> 
> Done 👍
> 
> Also I've applied the `HDF5_DIR` command to [continuous-integration.yml](https://github.com/pyccel/psydac/pull/399/files#diff-c471496bcb3ed47397b73d2e3e3aa41a8a679c86accaed9e05676fd9d5987434R85). So far it could parse the path correctly.

I have just realized that in the continuous integration we do not use `brew info hdf5-mpi` on macOS, but rather the command
```sh
HDF5_DIR=$((h5pcc -showconfig -echo || true) | grep "Installation point:" | cut -d ":" -f2 | tr -d " ")
```
which was given by @nhamidn in [commit c6e4cf1](https://github.com/pyccel/psydac/pull/200/commits/c6e4cf1713925eab7cbcab1d42f9ae8236c317b2). Maybe a similar command can be made to work with `brew`?

yguclu:
> > Maybe a similar command can be made to work with brew?
> 
> I've no macOS to test on, but perhaps `brew list hdf5-mpi | grep libhdf5.so | xargs dirname` may work?

This one seems to work:
```sh
brew list hdf5-mpi | grep "libhdf5.dylib" | xargs dirname | xargs dirname
```

kvrigor:
`HDF5_DIR=$(brew list hdf5-mpi | grep "libhdf5.dylib" | xargs dirname)` seems sufficient for `h5py`; the macOS test [passed](https://github.com/pyccel/psydac/actions/runs/8664454949/job/23760901726?pr=399) with this change 👍

I've modified the `README` once more—removing unnecessary details about the `HDF5_DIR`. I hope this makes the [Installing the library](https://github.com/pyccel/psydac/pull/399/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R86) step flow better.

yguclu (approved):
Excellent! Thanks!

Issue #406, opened 2024-05-07
Remove dependency on `distutils`


yguclu:
In order to support the most recent Python versions, we should remove any dependency on the `distutils` module, as this is not part of `python 3.12`. To this end we must:

- [x] Remove the `distutils` import from our [setup.py](https://github.com/pyccel/psydac/blob/devel/setup.py)
- [x] Make sure that SymPDE uses `sympy >= 1.12` (hence we must fix the SymPDE issue [#106](https://github.com/pyccel/sympde/issues/106))
- [x] Either drop the Igakit dependency, or update Igakit so that it does not use `distutils` to build the Fortran extension modules (see issue #53)

kvrigor:
Some updates:

* Remove the distutils import from our [setup.py](https://github.com/pyccel/psydac/blob/devel/setup.py) 
  - Addressed in #407
* Either drop the Igakit dependency, or update Igakit so that it does not use distutils to build the Fortran extension modules
  - Migration to Meson in progress:  [`pyccel/igakit#meson-build`](https://github.com/pyccel/igakit/tree/meson-build)

kvrigor:
With [igakit Py3.12](https://github.com/dalcinl/igakit/pull/5) already supported, we can close this issue.

PR #407, opened 2024-05-08
Remove deprecated distutils (Python >=3.12)

kvrigor:
Solves first TODO item in #406.  Summary of changes:

* Removed `finalize_options()` since this method does not differ from its [base class implementation](https://github.com/pypa/setuptools/blob/b3fc00fe1b242714029f46f03cafaaf324ae380d/setuptools/command/build_py.py#L38-L44), therefore making the override unnecessary.
* For `run()`, it's sufficient to call the [base class `run()`](https://github.com/pypa/setuptools/blob/b3fc00fe1b242714029f46f03cafaaf324ae380d/setuptools/command/build_py.py#L57) and then add the custom `pyccel` commands.
* `sub_run()` should throw an error instead of failing silently (`check=True`).
* Formatted code to conform with `Pylint` and `PEP8` checks

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -50,8 +50,9 @@ jobs:
           brew install open-mpi
           brew install hdf5-mpi
           brew install libomp
-          if [[ ! -f "/usr/local/bin/gfortran" ]]; then
-            ln -s $(which gfortran) /usr/local/bin/gfortran
+          if ! command -v gfortran; then
+            echo "gfortran could not be found"
+            exit 1
           fi
           echo "MPI_OPTS=--oversubscribe" >> $GITHUB_ENV
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,6 +47,7 @@ jobs:
       - name: Install non-Python dependencies on macOS
         if: matrix.os == 'macos-latest'
         run: |
+          brew install gcc
           brew install open-mpi
           brew install hdf5-mpi
           brew install libomp
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,13 +47,11 @@ jobs:
       - name: Install non-Python dependencies on macOS
         if: matrix.os == 'macos-latest'
         run: |
-          brew install gfortran
           brew install open-mpi
           brew install hdf5-mpi
           brew install libomp
           if [[ ! -f "/usr/local/bin/gfortran" ]]; then
-            gfort=$(ls /usr/local/bin/gfortran-* | tail -n 1)
-            ln -s ${gfort} /usr/local/bin/gfortran
+            ln -s $(which gfortran) /usr/local/bin/gfortran
           fi
           echo "MPI_OPTS=--oversubscribe" >> $GITHUB_ENV
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,13 +47,17 @@ jobs:
       - name: Install non-Python dependencies on macOS
         if: matrix.os == 'macos-latest'
         run: |
-          brew reinstall gcc
           brew install open-mpi
           brew install hdf5-mpi
           brew install libomp
-          if ! command -v gfortran; then
-            echo "gfortran could not be found"
-            exit 1
+          GFORTRAN_HOME=$(which gfortran || true)
+          echo "GFORTRAN_HOME : $GFORTRAN_HOME"
+          if [[ ! -f "$GFORTRAN_HOME" ]]; then
+            gfort=$(find ${PATH//:/\/ } -name 'gfortran-*' -exec basename {} \; | sort | tail -n 1 || true)
+            echo "Found $gfort"
+            gfort_path=$(which ${gfort})
+            folder=$(dirname ${gfort_path})
+            ln -s ${gfort_path} ${folder}/gfortran
           fi
           echo "MPI_OPTS=--oversubscribe" >> $GITHUB_ENV
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,6 +47,7 @@ jobs:
       - name: Install non-Python dependencies on macOS
         if: matrix.os == 'macos-latest'
         run: |
+          brew install gfortran
           brew install open-mpi
           brew install hdf5-mpi
           brew install libomp
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Install non-Python dependencies on macOS
         if: matrix.os == 'macos-latest'
         run: |
-          brew install gcc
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -1,57 +1,35 @@
...
-            sub_run([shutil.which('pyccel'), outfile, '--language', 'fortran', '--openmp'], shell=False)
-
-        # This part is copy from setuptools.command.build_py.build_module
-        if copied:
-            self.__updated_files.append(outfile)
+            self.announce(f"\nPyccelise module: {module}", level=logging.INFO)
+            sub_run([shutil.which('pyccel'), outfile,
+                     '--language', 'fortran',
+                     '--openmp'],
+                    shell=False, check=True) # nosec B603
```
kvrigor, line 24: This change reveals **a hidden bug in macOS CI test**. It fails `pyccel`-izing kernel files since `gfortran` is missing ([logfile](https://github.com/pyccel/psydac/actions/runs/9000566960/job/24724996688)):

```
Building wheels for collected packages: psydac, igakit
  Building wheel for psydac (pyproject.toml): started
  Building wheel for psydac (pyproject.toml): finished with status 'error'
  error: subprocess-exited-with-error
  
  × Building wheel for psydac (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [84 lines of output]
      running bdist_wheel
      running build
      8< --- LOG TRIMMED --- 8< 
      copying psydac/core/bsplines_kernels.py -> build/lib/psydac/core
      
      Pyccelise module: bsplines_kernels
      
      ERROR at Fortran compilation stage
      pyccel:
       |fatal [codegen]: bsplines_kernels.py| Could not find compiler (gfortran)
```

This is fixed by `brew reinstall gcc` (9015ccc5), which ensures that  `gfortran` is properly installed.

yguclu: Good catch!

yguclu: Do you know if it possible to add some print statements during the installation process, in order to show that the kernel files are pyccelized? Currently these are only shown with `pip install . --verbose`, which prints lots of additional information.

yguclu (commented):
After running `pip install . --verbose` I see the following warning messages:

1.
```shell
  Running command Getting requirements to build wheel
  <string>:23: DeprecationWarning:

    `numpy.distutils` is deprecated since NumPy 1.23.0, as a result
    of the deprecation of `distutils` itself. It will be removed for
    Python >= 3.12. For older Python versions it will remain present.
    It is recommended to use `setuptools < 60.0` for those Python versions.
    For more details, see:
      https://numpy.org/devdocs/reference/distutils_status_migration.html
```
2.
```shell
/private/var/folders/2g/2bkf6vld6yg482tr2n241zy00000gp/T/pip-build-env-a0j99xuc/overlay/lib/python3.11/site-packages/setuptools/_distutils/cmd.py:66: SetuptoolsDeprecationWarning: setup.py install is deprecated.
  !!

          ********************************************************************************
          Please avoid running ``setup.py`` directly.
          Instead, use pypa/build, pypa/installer or other
          standards-based tools.

          See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
          ********************************************************************************

  !!
    self.initialize_options()
```
3.
```shell
/private/var/folders/2g/2bkf6vld6yg482tr2n241zy00000gp/T/pip-build-env-a0j99xuc/overlay/lib/python3.11/site-packages/setuptools/command/sdist.py:115: SetuptoolsDeprecationWarning: `build_py` command does not inherit from setuptools' `build_py`.
  !!

          ********************************************************************************
          Custom 'build_py' does not implement 'get_data_files_without_manifest'.
          Please extend command classes from setuptools instead of distutils.

          See https://peps.python.org/pep-0632/ for details.
          ********************************************************************************

  !!
    self._add_data_files(self._safe_data_files(build_py))
```

Can we do something about warning no. 3 at least?

.github/workflows/continuous-integration.yml
```diff
@@ -47,12 +47,13 @@ jobs:
...
+          brew reinstall gcc
           brew install open-mpi
           brew install hdf5-mpi
           brew install libomp
-          if [[ ! -f "/usr/local/bin/gfortran" ]]; then
-            gfort=$(ls /usr/local/bin/gfortran-* | tail -n 1)
-            ln -s ${gfort} /usr/local/bin/gfortran
+          if ! command -v gfortran; then
+            echo "gfortran could not be found"
+            exit 1
```
yguclu, line 56: The macOS runners provided by GitHub already provide multiple version of the GFortran compiler, but none of them is actually called `gfortran` (see [here](https://github.com/actions/runner-images/blob/7781af869bec02d4d491cba6d01de412e96d865d/images/macos/macos-14-Readme.md?plain=1#L16-L18)).

Instead of reinstalling GCC, I believe that we can replicate [the strategy adopted in Pyccel](https://github.com/pyccel/pyccel/blob/1e97f1b6a21d226fc5863853252e8f720178bac6/.github/actions/macos_install/action.yml#L10-L18), where we find the newest version of GFortran and create a symbolic link to it called `gfortran`:
```yaml
        GFORTRAN_HOME=$(which gfortran || true)
        echo "GFORTRAN_HOME : $GFORTRAN_HOME"
        if [[ ! -f "$GFORTRAN_HOME" ]]; then
          gfort=$(find ${PATH//:/\/ } -name 'gfortran-*' -exec basename {} \; | sort | tail -n 1 || true)
          echo "Found $gfort"
          gfort_path=$(which ${gfort})
          folder=$(dirname ${gfort_path})
          ln -s ${gfort_path} ${folder}/gfortran
        fi
```

yguclu:
> Can we do something about warning no. 3 at least?

Upon closer inspection it appers that this warning is not caused by Psydac, but rather by one of its dependencies. So please ignore this comment.

setup.py
kvrigor: `subprocess.sub_run()` can be replaced by `subprocess.Popen()` to capture output of `pyccel` in realtime. I can modify this on the next commit.

kvrigor: > `subprocess.sub_run()` can be replaced by `subprocess.Popen()` to capture output of `pyccel` in realtime. I can modify this on the next commit.

Unfortunately this doesn't work and I'm not able to limit `pip` verbosity to `pyccel`.  The most likely reason is that every module logs at [logging.INFO](https://github.com/numpy/numpy/blob/67a04b100b9de75c72814701d0437f48f6651cb4/numpy/distutils/log.py#L67-L77) verbosity, and we're not able to silence those output.

Should we do  `pip install psydac --verbose` in CI? It seems be quite useful in detecting odd behavior that couldn't be tested easily. However combing through the CI logs would become more annoying 😅

.github/workflows/continuous-integration.yml
kvrigor: Implemented in [1f8f8cb](https://github.com/pyccel/psydac/pull/407/commits/1f8f8cb4d828c87fc9295701c8b9974430821f79)

setup.py
yguclu: I think we can leave things as they are for now. If we run into an installation issue we can always use the `--verbose` flag to get more information.

yguclu (approved):

PR #408, opened 2024-05-22
Improve conversion to PETSc

e-moral-sanchez:
Implement a new conversion to PETSc which reduces the communication in the assembly to almost none. The converted `PETSc.Vec` and `PETSc.Mat` objects locally contain the same coefficients as the local `StencilVector` or `StencilMatrix`/`BlockLinearOperator`. This is done with an index conversion function which maps the Psydac natural multi-index to the PETSc global index.

A pyccelizable kernel for computing the I,J,V format of PETSc has been added.

The conversion is not implemented for:
- `BlockVector` with at least a block that is not a `StencilVector`.
- `BlockLinearOperator` with at least a block that is not a `StencilMatrix`.
- Matrix-free `LinearOperator`s.


kvrigor:
The macOS runner now uses macOS 14 which uses the new Apple M1 silicon: https://github.com/orgs/community/discussions/69211

As a workaround, I'd suggest to change `macos-latest` to `macos-12` in this file:

https://github.com/pyccel/psydac/blob/904b7d7ba06cfdbe00f0ce05af7d544bbb30030c/.github/workflows/continuous-integration.yml#L25-L27

I think Psydac support for macOS 14 is a new feature and could be addressed in a separate PR.

e-moral-sanchez:
Thanks for the hint @kvrigor ! I tried it but the tests fail when printing the information about the MPI library. Do you know how to fix this?

kvrigor:
My bad; I checked again and discovered that we've been using `macOS-14` all along, but not exactly the same version. The last working macOS run was using **macOS-14.4.1**

image

And the failing one was using `macOS-14.5`:

image

This means we can't downgrade and thus we have to implement a workaround for the error introduced in `macOS-14.5`. I will experiment on this and share the bugfix once I find one.

pyproject.toml
```diff
@@ -50,7 +50,10 @@ dependencies = [
     'tblib',
 
     # IGAKIT - not on PyPI
-    'igakit @ https://github.com/dalcinl/igakit/archive/refs/heads/master.zip'
+
+    # !! WARNING !! Path to igakit below is from fork pyccel/igakit. This was done to
+    # quickly fix the numpy 2.0 issue. See https://github.com/dalcinl/igakit/pull/4
+    'igakit @ https://github.com/pyccel/igakit/archive/refs/heads/bugfix-numpy2.0.zip'
```
yguclu, line 56: @kvrigor @campospinto Do we still need this change?

yguclu:
@kvrigor Could you please add a couple of lines to the PR description and explain the changes to file `psydac/api/settings.py`?

pyproject.toml
campospinto: Maybe, if we allow igakit imports with python < 3.12?

yguclu: I am asking because the changes in the `pyccel/igakit` fork have also been merged in the `master` branch of Igakit: see https://github.com/dalcinl/igakit/pull/4

campospinto: ah sorry I have missed that

yguclu: I have removed it.

yguclu:
> @kvrigor Could you please add a couple of lines to the PR description and explain the changes to file `psydac/api/settings.py`?

Not necessary anymore, as those changes were added to `devel` in a separate PR, #421.

PR #411, opened 2024-06-12
Cache Psydac dependencies in CI

kvrigor:
~#408 observed failures on the `macOS` test, presumably related to the [recent changes in the `macOS` runner](https://github.com/orgs/community/discussions/102846). This PR will attempt to replicate this issue and fix it.~

**August 2024 update**: As ba0f47a already contains the macOS runner fixes, the scope of this PR has shifted to  caching of Psydac dependencies, e.g. PETSc, Python packages, and Ubuntu `apt` packages.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,7 +23,7 @@ jobs:
           - { isMerge: false, python-version: 3.9 }
           - { isMerge: false, python-version: '3.10' }
         include:
-          - os: macos-14
+          - os: macos-latest
             python-version: '3.10'
 
     name: ${{ matrix.os }} / Python ${{ matrix.python-version }}
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -35,6 +35,10 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip'
+
+      - name: Display CPU architecture used by local runner
+        run: python -c "import platform; print(f\"CPU type: \{platform.machine()\}\")"
 
       - name: Install non-Python dependencies on Ubuntu
         if: matrix.os == 'ubuntu-latest'
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -74,20 +74,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip
 
-#      - name: Get pip cache dir
-#        id: pip-cache-dir
-#        run: |
-#          echo "::set-output name=dir::$(python -m pip cache dir)"
-#
-#      - name: pip cache
-#        uses: actions/cache@v2
-#        id: pip-cache
-#        with:
-#          path: ${{ steps.pip-cache-dir.outputs.dir }}
-#          key: ${{ matrix.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('**/requirements.txt') }}
-#          restore-keys: |
-#            ${{ matrix.os }}-${{ matrix.python-version }}-pip-
-
       - name: Determine directory of parallel HDF5 library
         run: |
           if [[ "${{ matrix.os }}" == "ubuntu-latest" ]]; then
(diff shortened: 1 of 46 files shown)


kvrigor:
## Update 19.06

- [x] Fix to new macOS runner applied to [settings.py](https://github.com/pyccel/psydac/pull/411/files#diff-dafd44e893fc4b0d9d542088285a7bb0fee833d4a2c7690814503cc51ea9278a)
- [x] Implemented PETSc caching in CI. This shaves ~2.5 minutes off the total test time.
- [ ] New CI failure when installing Psydac, caused by recent release of `numpy 2.0.0.` Pending acceptance of PR bugfix: https://github.com/dalcinl/igakit/pull/4

.github/workflows/continuous-integration.yml
```diff
@@ -93,17 +82,38 @@ jobs:
...
+        name: Install PETSc with complex support
         working-directory: ./petsc
         run: |
           export PETSC_DIR=$(pwd)
           export PETSC_ARCH=petsc-cmplx
           ./configure --with-scalar-type=complex --with-fortran-bindings=0 --have-numpy=1         
-          make all check
+          make all
+          echo "PETSC_DIR=$PETSC_DIR" > petsc.env
+          echo "PETSC_ARCH=$PETSC_ARCH" >> petsc.env
```
yguclu, line 108: Do you use a single `>` in order to create a new `petsc.env` file?

kvrigor: yes @yguclu

kvrigor:
As ba0f47a already contains the macOS runner fixes, the scope of this PR is now limited to PETSc caching.

.github/workflows/continuous-integration.yml
```diff
@@ -96,17 +82,38 @@ jobs:
...
+
+      # This step is not really necessary and could be combined with PETSc install
+      # step; however it's good to verify if the cached PETSc installation really works!
+      - name: Test PETSc installation
+        working-directory: ./petsc
+        run: |
+          source petsc.env
```
yguclu, line 115: Are we sure that this file will always exist? It is created in the previous step, if there is no previous cache.

kvrigor: Yes; `./petsc` is the declared cached folder (line 91), so its contents will persist throughout multiple CI runs within a single PR. Basically `petsc.env` will be either in cache or created in the previous step.

.github/workflows/continuous-integration.yml
```diff
@@ -72,20 +72,6 @@ jobs:
...
-#
-#      - name: pip cache
-#        uses: actions/cache@v2
-#        id: pip-cache
-#        with:
-#          path: ${{ steps.pip-cache-dir.outputs.dir }}
-#          key: ${{ matrix.os }}-${{ matrix.python-version }}-pip-${{ hashFiles('**/requirements.txt') }}
-#          restore-keys: |
-#            ${{ matrix.os }}-${{ matrix.python-version }}-pip-
-
```
yguclu, line 88: Installing the Python dependencies takes about 2 minutes. Could it be beneficial to cache them too, as attempted here a long time ago?

yguclu: Actually, the "Install project" step also runs `pip` commands and takes about 2 minutes. The total time saving could be substantial.

.github/workflows/continuous-integration.yml
yguclu, line 1: I think we can use `pip` cache too, and add it to this PR. Apparently the `setup-python` action already supports this feature:
- https://github.com/actions/setup-python#caching-packages-dependencies
- https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

yguclu (changes requested):

.github/workflows/continuous-integration.yml
kvrigor: Last time I've tried this (16c4c5f20d) I've encountered a bunch of errors and hastily concluded that `pip`-caching is fragile. Perhaps at that time too many things were changing at once (_i.e._ new versions of CI runner and Python dependencies). Nevertheless, I also want faster CI tests so I'll try this again 😁

kvrigor:
@yguclu While we're at speeding up the CI workflow, do you think we should enable `fail-fast: true` here? Perhaps it makes sense to immediately abort all tests as soon as one fails.

https://github.com/pyccel/psydac/blob/fe14755a03cfb066b45f0ea81813d384e0c8a7c7/.github/workflows/continuous-integration.yml#L15-L16

yguclu:
> @yguclu While we're at speeding the CI workflow, do you think we should enable `fail-fast: true` here? Perhaps it makes sense to immediately abort all tests as soon as one fails.
> 
> https://github.com/pyccel/psydac/blob/fe14755a03cfb066b45f0ea81813d384e0c8a7c7/.github/workflows/continuous-integration.yml#L15-L16

The `fail-fast` option applies to the entire job matrix. If it is set to `true`, GitHub will cancel all in-progress and queued jobs in the matrix as soon as one job fails. For example, in the case that our tests fail on Ubuntu, it will also stop the macOS jobs... This makes it harder to catch a platform-specific problem IMHO.

.github/workflows/continuous-integration.yml
```diff
@@ -37,14 +37,24 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
+          cache: 'pip' # Enable pip-caching
```
yguclu, line 40: Since we have the file `requirements_extra.txt` which is not automatically checked by `setup-python`, I suspect that we need to specify the `cache-dependency-path` as described in the documentation.

```suggestion
          cache: 'pip' # Enable pip-caching
          cache-dependency-path: |
            pyproject.toml
            requirements.txt
            requirements_extra.txt
```

.github/workflows/continuous-integration.yml
```diff
@@ -37,14 +37,28 @@ jobs:
...
+        if: matrix.os == 'ubuntu-latest'
+        uses: awalsh128/cache-apt-pkgs-action@latest
+        with:
+          packages: gfortran openmpi-bin libopenmpi-dev libhdf5-openmpi-dev
+          version: 1.0
+          execute_install_scripts: true
+
+      # When loading cached apt packages, the default MPI compiler isn't set.
+      # Workaround is to 'reinstall' openmpi-bin, which doesn't actually perform install
+      # installation (since openmpi-bin already exists), but instead reruns `update-alternatives`
```
yguclu, line 56: It looks like the word "install" (before "installation") is not needed

yguclu (changes requested):
Good job! This looks very good to me!

There is just one duplication error in a comment, which I pointed out before.

.github/workflows/continuous-integration.yml
yguclu: Fixed with 03c532d

yguclu (approved):

PR #415, opened 2024-07-04
Matrix-free linear operator

e-moral-sanchez:
This PR provides the class `MatrixFreeLinearOperator`, which allows creating a general `LinearOperator`  only requiring the domain, the codomain and a callable `dot` method.


campospinto:
the classic fork-trap ;)

PR #416, opened 2024-07-19
Test sympde Norm and SemiNorm

campospinto:
Several tests are added for discretised sympde Norm and SemiNorm using real and complex functions and FEM fields.

In various cases the tests fail: they should be fixed.

See also issue ?


campospinto:
this PR has been opened on the fork, see PR https://github.com/campospinto/psydac_dev/pull/11

PR #417, opened 2024-07-26
Test sympde branch terminal_expressions_for_NS

campospinto:
The purpose of this PR is not to be merged, but to trigger the psydac test suite against a specific sympde branch.

In particular, there is nothing to review or approve. If the tests fail for this PR, then:
- either this is due to an error in sympde and the fix should be done in the corresponding sympde branch
- or this is due to a problem in psydac which should be handled in a dedicated PR (not this one).

Here we are verifying that psydac will not be broken by changes in the sympde 
PR https://github.com/pyccel/sympde/pull/161 are breaking psydac.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu-latest ]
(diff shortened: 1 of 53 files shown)


campospinto:
opened in the fork psydac_dev...

Issue #420, opened 2024-07-31
Make Psydac compatible with `mpi4py` 4.0.0


yguclu:
Version 4 of `mpi4py` was released on 28 July 2024 (see [here](https://mpi4py.readthedocs.io/en/stable/)), and apparently it is not backward compatible, as many of our unit tests fail now. We should decide soon if we want to be compatible with both versions 3 and 4, or only support version 4. For the time being we require version 3.

UPDATE as of 26 March 2024: Given that we want to support Python 3.13 (see pyccel/sympde#176 and #475), we must switch to version 4 as soon as possible.

spossann:
I have found the issue and will push a branch with some improvements.

Issue #419, opened 2024-07-31
Make Psydac compatible with SciPy 1.14


yguclu:
In order to use the latest version of SciPy, we should stop passing the deprecated `tol` argument in calls to SciPy's `minres`.

We should be careful as SciPy 1.14 requires Python 3.10 or newer.

PR #421, opened 2024-07-31
 Fixes for CI failures

yguclu:
- Modify Pyccel compilation flags to support both Intel and Apple Silicon processors on macOS
- Do not use new version 4 of mpi4py -- see issue #420
- Do not use new version 1.14 of SciPy -- see issue #419


psydac/api/settings.py
```diff
@@ -41,8 +41,22 @@
...
+    else:
+        # TODO: Support later Apple CPU models. Perhaps the CPU naming scheme could be easily guessed
+        # based on the output of 'sysctl -n machdep.cpu.brand_string', but I wouldn't rely on this
+        # guess unless it has been manually verified. Loud errors are better than silent failures!
+        raise SystemError(f"Unsupported Apple CPU '{cpu_brand}'.")
+else:
+    # Default architecture flags
+    PSYDAC_BACKEND_GPYCCEL['flags'] += ' -march=native -mtune=native'
+    if platform.machine() == 'x86_64':
+        PSYDAC_BACKEND_GPYCCEL['flags'] += ' -mavx'
```
campospinto, line 59: what is a good way to check that these flags are correct ?

campospinto: (do we have unit tests for the pyccelization of python kernels?)

campospinto (commented):
I have a question regarding the api/settings.py, otherwise these changes seem good to me.

psydac/api/settings.py
yguclu: Many of our kernels are pyccelized during the installation process, and there we do not pass any special flags AFAIK.

These flags are for the kernels that are generated automatically, i.e. those which are responsible for assembling a scalar, a vector, or a matrix.

yguclu: Earlier in Psydac all the kernels were generated automatically, so these optimization flags were used everywhere. Now it is clear that we need another approach, maybe a separate command which generates the correct flags and writes them to file. We might have an issue about this already.

yguclu: @campospinto Sorry I did not answer your question directly. Currently there is no unit test which checks the pyccelization of a Python kernel using these flags. It might be a good idea to write one. We will have to run this test separately before everything else, so that we can stop the CI job in case of a failure.

yguclu: In addition to testing these optimization flags, I think it could be useful to use them also for the hardcoded kernels (which are pyccelized at installation).

campospinto: Ok, thanks! I dit not see any issues about testing pyccelization flags, I don't know if you want to open one ?

campospinto (approved):
it may be a good idea to open an issue about testing pyccelization flags, apart from that I'm fine with this PR

PR #424, opened 2024-08-05
Support Python 3.12

### LINKED ISSUE #406: Remove dependency on `distutils`

yguclu:
We make Psydac support Python 3.12 by using NumPy >= 2.0 together with Petsc4py >= 3.21 and the master branch of Igakit (updated on 6 Aug 2024). We also increase the required SymPDE version to 0.19.0. For simplicity, the same requirements are made for older versions of Python.

This closes #406.

Rationale
---------
Python 3.12 does not have a `distutils` module. Therefore, NumPy 2.0 was released on 16 Jun 2024 with a new `meson` backend for the `f2py` tool which is part of the library. As of 6 Aug 2024, Igakit also moved from `distutils` to the `meson-python` build system in combination with NumPy 2.0.

The first version of `petsc4py` which is compatible with Python 3.12 is shipped with PETSc 3.21, hence we use this now instead of PETSc 3.20. Such a version of `petsc4py` also requires NumPy >= 2.0.

The move to NumPy 2.0 also means that some numeric datatypes behave differently. For this we need SymPDE >= 0.19.0.

Summary of changes
------------------
* In `pyproject.toml`, set maximum Python version to 3.12, and require SymPDE 0.19.0 instead of 0.18.3;
* In `README.md`, increase PETSc version to 3.21.4;
* In method `integral` of `TensorFemSpace`, convert any NumPy result to the native Python type to avoid errors with SymPy's `sympify` (after the upgrade to NumPy 2.0);
* Apply some bug fixes to the unit tests after the upgrade of NumPy and SymPDE;
* In `continuous-integration.yml`, add tests with Python 3.12, increase PETSc version to 3.21.4, and add check of parallel h5py.


kvrigor:
Summary of what I understood in this PR:

- Python3.12 support enabled in  `continuous-integration.yml` and new sympde `0.19.0` reflected in `pyproject.toml`
- Diffs in tests are due to bugfixes by @campospinto and @yguclu  
- Running `maxwell_2d_multi_patch.py` gives a passing result:

```sh
$ python3.12 examples/maxwell_2d_multi_patch.py
> Grid          :: [4,4]
> Degree        :: [2,2]
> CG info       ::  {'niter': 829, 'success': np.True_, 'res_norm': 9.976906608695098e-09}
> L2 error      :: 1.59e+00

> Solution time :: 2.72e+01
> Evaluat. time :: 3.61e-02
```

image

kvrigor (approved):

PR #427, opened 2024-08-07
Test Pyccel optimization flags in CI

### LINKED ISSUE #425: Test Pyccel optimization flags
The compiler optimization flags used for pyccelizing the generated kernels are not tested. If these flags are not appropriate for the current compiler and architecture, compilation fails with an error at some point during our test suite.

It could be wise to pyccelize a simple kernel using these flags as a separate step in the CI pipeline, before the main test suite. By testing the flags early, we could catch the problem earlier.

cc: @campospinto

kvrigor:
Fixes #425


yguclu:
Good job @kvrigor! But this PR must wait for #424.

PR #430, opened 2024-09-10
Avoid installing h5py from cache

kvrigor:
From @yguclu:

> If people have a cached version of the h5py library which was compiled using NumPy < 2.0 and try to install Psydac in a clean Python environment, they end up with a broken version. This is because the cached version of h5py tries to call C functions from NumPy 2.0 which are not ABI compatible.

Adding [`--no-cache-dir`](https://stackoverflow.com/questions/45594707/what-is-pips-no-cache-dir-good-for) to `pip` guarantees that `h5py` is always [built during installation](https://docs.h5py.org/en/stable/build.html#custom-installation). Re-using a prebuilt `h5py` (*i.e.* cached `h5py`) could become an issue stale when `h5py` dependencies upgrade.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,9 +19,9 @@ jobs:
         python-version: [ 3.9, '3.10', '3.11', '3.12' ]
         isMerge:
           - ${{ github.event_name == 'push' && github.ref == 'refs/heads/devel' }}
-        exclude:
-          - { isMerge: false, python-version: '3.10' }
-          - { isMerge: false, python-version: '3.11' }
+        #exclude:
+        #  - { isMerge: false, python-version: '3.10' }
+        #  - { isMerge: false, python-version: '3.11' }
         include:
           - os: macos-latest
             python-version: '3.10'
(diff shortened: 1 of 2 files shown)


.github/workflows/continuous-integration.yml
```diff
@@ -19,9 +19,9 @@ jobs:
         python-version: [ 3.9, '3.10', '3.11', '3.12' ]
         isMerge:
           - ${{ github.event_name == 'push' && github.ref == 'refs/heads/devel' }}
-        exclude:
-          - { isMerge: false, python-version: '3.10' }
-          - { isMerge: false, python-version: '3.11' }
+        #exclude:
+        #  - { isMerge: false, python-version: '3.10' }
+        #  - { isMerge: false, python-version: '3.11' }
```
yguclu, line 24: This was just a temporary test I made. It can be reverted now

```suggestion
        exclude:
          - { isMerge: false, python-version: '3.10' }
          - { isMerge: false, python-version: '3.11' }
```

yguclu (approved):

PR #433, opened 2024-09-16
Update calls to `numpy.reshape` to avoid deprecation warnings

### LINKED ISSUE #432: Deprecation Warning for np.reshape
When reinstalling psydac and running the tests, I got this warning
> DeprecationWarning: `newshape` keyword argument is deprecated, use `shape=...` or pass shape positionally instead. (deprecated in NumPy 2.1)`

in the `np.reshape` function. Since this is essentially removing three letters, I figured I could fix this while waiting for the other tests to finish. (I also added `shape=` in the other occurrences for consistency) 
See #433.

FrederikSchnack:
NumPy 2.1 deprecates the `newshape` argument of the `reshape` function. Instead, one should use the positional-only argument `shape`. Every call in Psydac has been updated accordingly. This closes #432.

### DIFF
--- a/psydac/api/postprocessing.py
+++ b/psydac/api/postprocessing.py
@@ -2039,7 +2039,7 @@ class PostProcessManager:
             grid_local= [grid[i][
                 local_domain[0][i] * npts_per_cell[i]:(local_domain[1][i] + 1) * npts_per_cell[i]]
                         for i in range(ldim)]
-            grid_as_arrays = [np.reshape(grid[i], (len(grid[i])//npts_per_cell[i], npts_per_cell[i]))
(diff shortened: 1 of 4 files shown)


campospinto (commented):
So if understand correctly, these changes remove the warning for Numpy 2.1 but are not incompatible with the Python 3.9 tests (and the corresponding Numpy version) ? So we should probably wait for Python 3.9 to be dropped (Fall 2025?)

FrederikSchnack:
Yes exactly, I was a little too fast with the changes, but I don't know if there is any actual benefit of maintaining old versions of Python for long times (especially since all newer NumPy versions will not be compatible). But we could also remove the keywords in `np.reshape` completely, which I don't like conceptually, but it would remove the warnings and pass the tests with all Python versions.

yguclu (approved):

PR #434, opened 2024-09-17
Add sqrt option for diagonal in stencil and block matrices.

maxlin-ipp:
Corresponds to commit [121215d](https://github.com/pyccel/psydac/commit/121215db3f1ee2076e6560d631bb7bf522c8caae) in stefan-psydac.

### DIFF
--- a/psydac/linalg/block.py
+++ b/psydac/linalg/block.py
@@ -777,13 +777,17 @@ class BlockLinearOperator(LinearOperator):
     #--------------------------------------
     # New properties/methods
     #--------------------------------------
-    def diagonal(self, *, inverse = False, out = None):
+    def diagonal(self, *, inverse = False, sqrt = False, out = None):
         """Get the coefficients on the main diagonal as another BlockLinearOperator object.
 
         Parameters
         ----------
         inverse : bool
             If True, get the inverse of the diagonal. (Default: False).
+
+        sqrt : bool
+            If True, get the square root of the diagonal. (Default: False).
+            Can be combined with inverse to get the inverse square root
 
         out : BlockLinearOperator
             If provided, write the diagonal entries into this matrix. (Default: None).
@@ -815,7 +819,7 @@ class BlockLinearOperator(LinearOperator):
         # Store the diagonal (or its inverse) into `out`
         for i, j in self.nonzero_block_indices:
             if i == j:
-                out[i, i] = self[i, i].diagonal(inverse = inverse, out = out[i, i])
+                out[i, i] = self[i, i].diagonal(inverse = inverse, sqrt=sqrt, out = out[i, i])
 
         return out
 
--- a/psydac/linalg/stencil.py
+++ b/psydac/linalg/stencil.py
@@ -1372,7 +1372,7 @@ class StencilMatrix(LinearOperator):
                 self._data[idx_to] += self._data[idx_from]
 
     # ...
-    def diagonal(self, *, inverse = False, out = None):
+    def diagonal(self, *, inverse = False, sqrt = False, out = None):
         """
         Get the coefficients on the main diagonal as a StencilDiagonalMatrix object.
 
@@ -1380,6 +1380,10 @@ class StencilMatrix(LinearOperator):
         ----------
         inverse : bool
             If True, get the inverse of the diagonal. (Default: False).
+
+        sqrt : bool
+            If True, get the square root of the diagonal. (Default: False).
+            Can be combined with inverse to get the inverse square root
 
         out : StencilDiagonalMatrix
             If provided, write the diagonal entries into this matrix. (Default: None).
@@ -1409,6 +1413,9 @@ class StencilMatrix(LinearOperator):
         diagonal_indices = self._get_diagonal_indices()
         diag = self._data[diagonal_indices]
         data = out._data if out else None
+
+        if sqrt:
+            diag = np.sqrt(diag)
 
         # Calculate entries of StencilDiagonalMatrix
         if inverse:


psydac/linalg/stencil.py
```diff
@@ -1410,6 +1414,9 @@ def diagonal(self, *, inverse = False, out = None):
         diag = self._data[diagonal_indices]
         data = out._data if out else None
 
+        if sqrt:
+            diag = np.sqrt(diag)
+
```
yguclu, line 1419: This creates a new NumPy array with its own storage, which should not be necessary at this stage. Instead, you could pass the `out` argument to `np.sqrt` and move this code to **after** the next `if/elif/else` block:
```python
if sqrt:
    data = np.sqrt(data, out=data)
```

psydac/linalg/block.py
```diff
@@ -777,14 +777,18 @@ def __sub__(self, M):
...
         """Get the coefficients on the main diagonal as another BlockLinearOperator object.
 
         Parameters
         ----------
         inverse : bool
             If True, get the inverse of the diagonal. (Default: False).
 
+        sqrt : bool
+            If True, get the square root of the diagonal. (Default: False).
+            Can be combined with inverse to get the inverse square root
```
yguclu, line 790: Missing end period:
```suggestion
            Can be combined with inverse to get the inverse square root.
```

psydac/linalg/block.py
```diff
@@ -777,14 +777,18 @@ def __sub__(self, M):
...
     # New properties/methods
     #--------------------------------------
-    def diagonal(self, *, inverse = False, out = None):
+    def diagonal(self, *, inverse = False, sqrt = False, out = None):
         """Get the coefficients on the main diagonal as another BlockLinearOperator object.
 
         Parameters
         ----------
         inverse : bool
             If True, get the inverse of the diagonal. (Default: False).
```
yguclu, line 786: How about adding the same comment as in the next parameter?
```suggestion
        inverse : bool
            If True, get the inverse of the diagonal. (Default: False).
            Can be combined with sqrt to get the inverse square root.
```

psydac/linalg/stencil.py
```diff
@@ -1381,6 +1381,10 @@ def diagonal(self, *, inverse = False, out = None):
         inverse : bool
             If True, get the inverse of the diagonal. (Default: False).
```
yguclu, line 1382: How about adding the same comment as in the next parameter?
```suggestion
        inverse : bool
            If True, get the inverse of the diagonal. (Default: False).
            Can be combined with sqrt to get the inverse square root.
```

psydac/linalg/stencil.py
```diff
@@ -1381,6 +1381,10 @@ def diagonal(self, *, inverse = False, out = None):
...
 
+        sqrt : bool
+            If True, get the square root of the diagonal. (Default: False).
+            Can be combined with inverse to get the inverse square root
```
yguclu, line 1386: Missing end period:
```suggestion
            Can be combined with inverse to get the inverse square root.
```

yguclu (changes requested):

psydac/linalg/stencil.py
maxlin-ipp: Good point, is it not enough to write `np.sqrt(data, out=data)` (and `np.divide(1, diag, out=data)`) for updating data in place?

yguclu: Yes, the assignment is just syntactic sugar.

yguclu (approved):

psydac/linalg/stencil.py
```diff
@@ -1410,14 +1415,18 @@ def diagonal(self, *, inverse = False, out = None):
         diag = self._data[diagonal_indices]
         data = out._data if out else None
 
+
```
yguclu, line 1418: Minor correction.. We should not use two whitelines inside of a class definition, but only to separate two classes from each other (see PEP-8 for further information).
```suggestion
```

yguclu (approved):

psydac/linalg/stencil.py
maxlin-ipp: Ok!

yguclu (commented):
Is there a unit test for the new option somewhere?

maxlin-ipp:
No, there isn't. Should I include one in `psydac/linalg/tests/test_linalg.py`?

yguclu:
> No, there isn't. Should I include one in `psydac/linalg/tests/test_linalg.py`?

Yes please!

psydac/linalg/tests/test_linalg.py
```diff
@@ -528,8 +528,9 @@ def test_in_place_operations(n1, n2, p1, p2, P1=False, P2=False):
 @pytest.mark.parametrize('n2', n2array)
 @pytest.mark.parametrize('p1', p1array)
 @pytest.mark.parametrize('p2', p2array)
+@pytest.mark.parametrize('sqrt', [False, True])
 
-def test_inverse_transpose_interaction(n1, n2, p1, p2, P1=False, P2=False):
+def test_inverse_transpose_interaction(n1, n2, p1, p2, sqrt, P1=False, P2=False):
```
yguclu, line 533: The Pytest parameter `sqrt` is not used in this test, hence you can remove it.

yguclu (changes requested):

psydac/linalg/tests/test_linalg.py
maxlin-ipp: Done!

psydac/linalg/tests/test_linalg.py
```diff
@@ -618,6 +619,12 @@ def test_inverse_transpose_interaction(n1, n2, p1, p2, P1=False, P2=False):
     ### -1,T & T,-1 --- -1,T,T --- -1,T,-1 --- T,-1,-1 --- T,-1,T (the combinations I test)
     ###
 
+    # Square root test
+    scaled_matrix = B * np.random.random() # Ensure the diagonal elements != 1
+    diagonal_values = scaled_matrix.diagonal(sqrt=False).toarray()
+    sqrt_diagonal_values = scaled_matrix.diagonal(sqrt=True).toarray()
+    assert np.allclose(sqrt_diagonal_values, np.sqrt(diagonal_values))
```
yguclu, line 625: When using `allclose` in a unit test, it is good practice to specify the relative and absolute tolerances.

In this case I wonder if we could use an equality check, given that ultimately we are calling the same function `numpy.sqrt` in both cases. Please try:
```suggestion
    assert np.array_equal(sqrt_diagonal_values, np.sqrt(diagonal_values))
```

yguclu (changes requested):

psydac/linalg/tests/test_linalg.py
maxlin-ipp: Ok, I redid the test with your suggestion, and it works. I'll commit your suggestion.

yguclu (approved):

PR #439, opened 2024-09-18
Fix pdim bug in VectorFemSpace. 

maxlin-ipp:
Corresponds to commit [3ee2acd](https://github.com/pyccel/psydac/commit/3ee2acd83aa448114c8c83fd539fe3a52f7179f7) in `stefan-psydac`.

### DIFF
--- a/psydac/fem/vector.py
+++ b/psydac/fem/vector.py
@@ -340,7 +340,7 @@ class VectorFemSpace( FemSpace ):
         #      - shall we check the bc too?
 
         degree = [V.degree for V in self.spaces]
-        if self.pdim == 1:
+        if self.ldim == 1:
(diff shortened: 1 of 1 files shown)


psydac/fem/vector.py
yguclu, line 1: The property `is_block` is never used in Psydac, and I think it can be safely removed from the library.

What is Struphy using it for?

maxlin-ipp: I don't don't see it used used anywhere in Struphy either. So removing it should be fine.

yguclu: All right, please go ahead and do it in this PR. The title and description of this PR will have to be changed accordingly.

yguclu (approved):

PR #440, opened 2024-09-18
Circumvent `apt-get` download error

yguclu:
Update `apt` index files with `apt-get update` to avoid download error in CI workflow.


campospinto (approved):
Out of curiosity, which download error happens without this change ?

yguclu:
> Out of curiosity, which download error happens without this change ?

This: https://github.com/pyccel/psydac/actions/runs/10924986097/job/30328052043

Issue #431, opened 2024-09-12
pyccel acceleration does not seem to work in C


campospinto:
The pyccelization of some kernels does not seem to work in C.

For instance, with the current version of devel, the command 
`python psydac_accelerate.py --language c ` yields the following errors:

```
  Pyccelize file: /Users/campos/Work/codes/pyccel/psydac/psydac/core/bsplines_kernels.py

ERROR at code generation stage
pyccel:
 |fatal [codegen]: bsplines_kernels.py [326,27]| _print_NumpyMatmul is not yet implemented for language : C
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language. (<pyccel.ast.numpyext.NumpyMatmul object at 0x11e846dc0>)
```
and 
```
  Pyccelize file: /Users/campos/Work/codes/pyccel/psydac/psydac/core/field_evaluation_kernels.py

ERROR at code generation stage
pyccel:
 |fatal [codegen]: field_evaluation_kernels.py [2326,56]| _print_PythonList is not yet implemented for language : C
Pyccel has encountered syntax that has not been implemented yet. Please create an issue at https://github.com/pyccel/pyccel/issues and provide a small example of your problem. Do not forget to specify your target language. ((x_x1, x_x2, x_x3))
```
I am running this on Mac OS but the error seems also present on other machines

campospinto:
Some of these errors seem to be addressed in the Struphy psydac fork https://github.com/spossann/stefan-psydac, see for instance the commits 

[Enable compilation with language C](https://github.com/pyccel/psydac/commit/9a016b1455c7bb7f50be7d91fb9bc3088005e3d8)

and 

[Comment jacobian matrix eval kernels; lists not supported yet for c-c](https://github.com/pyccel/psydac/commit/e4a648807cf83d631f1864e5c86cb70db0313473)

maxlin-ipp:
I replaced the python lists in `field_evaluation_kernels.py`.

I can successfully use `pyccel psydac/core/field_evaluation_kernels.py --language=c` on my branch now (stefan-psydac had simply commented out all the kernels).

For `bsplines_kernels.py`, we need to replace `np.matmul`, `np.sum`, `sum`, `np.min`, and `np.max`. This is done in stefan-psydac in `psydac/core/arrays.py` which does not exist in psydac devel.

yguclu:
[...]

PR #441, opened 2024-09-19
Mention C support in near future in README

yguclu:
Clarify in the `README.md` file that, although the C backend may be selected for accelerating the kernel files with Pyccel, this is not fully working yet. Hence the Fortran backend (which is the default) is the only one available. A future version of Pyccel will certainly provide a C backend as capable as the Fortran one. See issue #431.


campospinto (approved):
Good to me (again) !

PR #442, opened 2024-09-26
Update `stencil2coo` kernels for shifts > 1

maxlin-ipp:
Fix the indexing in all kernels in `stencil2coo_kernels.py` so that the method `toarray` works for `StencilMatrix` objects in 1D/2D/3D with `shift` larger than 1.

The unit tests for `toarray` in `test_stencil_matrix.py` have been parametrized to run with `shift` values of 1 and 2.

### DIFF
--- a/psydac/linalg/tests/test_stencil_matrix.py
+++ b/psydac/linalg/tests/test_stencil_matrix.py
@@ -489,8 +489,8 @@ def test_stencil_matrix_1d_serial_toarray( dtype, n1, p1, s1, P1):
 @pytest.mark.parametrize('n2', [8, 7])
 @pytest.mark.parametrize('p1', [1, 2])
 @pytest.mark.parametrize('p2', [1, 3])
-@pytest.mark.parametrize('s1', [1,2])
-@pytest.mark.parametrize('s2', [1,2])
+@pytest.mark.parametrize('s1', [1, 2])
+@pytest.mark.parametrize('s2', [1, 2])
 @pytest.mark.parametrize('P1', [True, False])
 @pytest.mark.parametrize('P2', [True, False])
 def test_stencil_matrix_2d_serial_toarray( dtype, n1, n2, p1, p2, s1, s2, P1, P2):
--- a/psydac/linalg/tests/test_stencil_matrix.py
+++ b/psydac/linalg/tests/test_stencil_matrix.py
@@ -2191,7 +2191,7 @@ def test_stencil_matrix_2d_serial_backend_switch(dtype, n1, n2, p1, p2, s1, s2, 
 @pytest.mark.parametrize('dtype', [float, complex])
 @pytest.mark.parametrize('n1', [20, 67])
 @pytest.mark.parametrize('p1', [1, 2, 3])
-@pytest.mark.parametrize('sh1', [1])
+@pytest.mark.parametrize('sh1', [1, 2])
 @pytest.mark.parametrize('P1', [True, False])
 @pytest.mark.parallel
 def test_stencil_matrix_1d_parallel_toarray(dtype, n1, p1, sh1, P1):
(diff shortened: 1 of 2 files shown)


psydac/linalg/tests/test_stencil_matrix.py
```diff
@@ -3154,6 +3154,50 @@ def test_stencil_matrix_2d_parallel_backend_dot(dtype, n1, n2, p1, p2, sh1, sh2,
     assert (M + M).backend is M.backend
     assert (2 * M).backend is M.backend
 
+@pytest.mark.parametrize('dtype', [float])#, complex])
+@pytest.mark.parametrize('n1', [4,6,10])
+@pytest.mark.parametrize('p1', [3,4])
+@pytest.mark.parametrize('s1', [1,2,3])
+def test_stencil_matrix_1d_toarray(dtype, n1, p1, s1, P1=True):
```
yguclu, line 3161: There is already a function called `test_stencil_matrix_1d_serial_toarray`. Why not parametrizing that one instead of writing a new function?

yguclu: See [here](https://github.com/pyccel/psydac/blame/7e621bb5dc82b60c8e460680d5262965424e5a45/psydac/linalg/tests/test_stencil_matrix.py#L433-L438):
```python
@pytest.mark.parametrize('dtype', [float])
@pytest.mark.parametrize('n1', [7])
@pytest.mark.parametrize('p1', [1,2])
@pytest.mark.parametrize('s1', [1])
@pytest.mark.parametrize('P1', [True, False])
def test_stencil_matrix_1d_serial_toarray( dtype, n1, p1, s1, P1):
    ...
```

maxlin-ipp: You're right, I didn't see it at first. There is also a test for 2d in `test_stencil_matrix_2d_serial_toarray`, I will parameterize both.

Should I leave the 3d kernels as they are until we have a `test_stencil_matrix_2d_serial_toarray` unit test too?

psydac/linalg/tests/test_stencil_matrix.py
```diff
@@ -433,7 +433,7 @@ def test_stencil_matrix_2d_serial_spurious_entries( dtype, p1, p2, s1, s2, P1, P
 @pytest.mark.parametrize('dtype', [float])
 @pytest.mark.parametrize('n1', [7])
 @pytest.mark.parametrize('p1', [1,2])
-@pytest.mark.parametrize('s1', [1])
+@pytest.mark.parametrize('s1', [1,2])
```
yguclu, line 436: Minor suggestion on style:
```suggestion
@pytest.mark.parametrize('s1', [1, 2])
```

psydac/linalg/tests/test_stencil_matrix.py
```diff
@@ -489,8 +489,8 @@ def test_stencil_matrix_1d_serial_toarray( dtype, n1, p1, s1, P1):
 @pytest.mark.parametrize('n2', [8, 7])
 @pytest.mark.parametrize('p1', [1, 2])
 @pytest.mark.parametrize('p2', [1, 3])
-@pytest.mark.parametrize('s1', [1])
-@pytest.mark.parametrize('s2', [1])
+@pytest.mark.parametrize('s1', [1,2])
+@pytest.mark.parametrize('s2', [1,2])
```
yguclu, line 493: Minor suggestion on style:
```suggestion
@pytest.mark.parametrize('s1', [1, 2])
@pytest.mark.parametrize('s2', [1, 2])
```

yguclu (changes requested):

psydac/linalg/tests/test_stencil_matrix.py
yguclu, line 1: Have you tried to parametrize the shift also in [`test_stencil_matrix_1d_parallel_toarray`](https://github.com/pyccel/psydac/blob/78bffd600b9d6ab7bb63aef372ca0d35c3e20173/psydac/linalg/tests/test_stencil_matrix.py#L2197)?

maxlin-ipp: I just tested it, and it works. I will add the shift there as well.

yguclu (approved):

PR #445, opened 2024-10-16
Use SymPDE version 0.19.1

yguclu:
Update Psydac after the breaking change introduced by version 0.19.1 of SymPDE: domain coordinates are real symbols.

Further, the new version is needed in order to solve #457.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,7 @@ dependencies = [
     'pyevtk',
 
     # Our packages from PyPi
-    'sympde == 0.19.0',
+    'sympde @ https://github.com/pyccel/sympde/archive/refs/heads/real_coord.zip',
     'pyccel >= 1.11.2',
     'gelato == 0.12',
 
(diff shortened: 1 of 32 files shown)


e-moral-sanchez:
This PR is linked to https://github.com/pyccel/sympde/pull/170 .

After the change that coordinates must be real, the symbols used in the tests to write the exact solution must also be declared as real.

pyproject.toml
```diff
@@ -31,7 +30,7 @@ dependencies = [
     'pyevtk',
 
     # Our packages from PyPi
-    'sympde == 0.19.0',
+    'sympde @ https://github.com/pyccel/sympde/archive/refs/heads/real_coord.zip',
```
e-moral-sanchez, line 33: Just as a reminder, we need to revert this line before merging

e-moral-sanchez (changes requested):

pyproject.toml
yguclu: And by the way, yes we should bump up SymPDE version to 0.17.1

campospinto: Oh, I see that I need to review the Sympde PR... I'll do it asap

e-moral-sanchez: Thanks for taking a look. This PR is ready.

campospinto (approved):

pyproject.toml
yguclu: We now use SymPDE version 0.19.1

PR #446, opened 2024-10-16
Fix Greville point collocation matrix for single grid point case

max-models:
Previously, in the case when `self.greville.size == 1`, `collocation_matrix` would return an empty matrix.

Now, if the grid has a single Greville point, the collocation matrix is simplified to a 1x1 matrix with a value of 1.

### DIFF
--- a/psydac/fem/splines.py
+++ b/psydac/fem/splines.py
@@ -171,14 +171,17 @@ class SplineSpace( FemSpace ):
         Greville points.
 
         """
-        imat = collocation_matrix(
-            knots    = self.knots,
-            degree   = self.degree,
-            periodic = self.periodic,
-            normalization = self.basis,
-            xgrid    = self.greville,
-            multiplicity = self.multiplicity
-        )
+        if self.greville.size == 1:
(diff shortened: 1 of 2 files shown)


psydac/fem/splines.py
```diff
@@ -171,14 +171,17 @@ def init_interpolation( self, dtype=float ):
...
+            imat = np.ones((1, 1), dtype=float)
+        else:
+            imat = collocation_matrix(
+                knots    = self.knots,
+                degree   = self.degree,
+                periodic = self.periodic,
+                normalization = self.basis,
+                xgrid    = self.greville,
+                multiplicity = self.multiplicity
+            )
```
yguclu, line 184: I suggest that this case be handled inside the function `collocation_matrix` itself.

yguclu (changes requested):

campospinto (commented):
I agree with Yaman's suggestion

yguclu (approved):

psydac/fem/splines.py
yguclu: Done in df5ed36

PR #448, opened 2024-11-08
Fast Matrix Assembly

jowezarek:
Matrix Assembly in Psydac
--------------------------------------
https://datashare.mpcdf.mpg.de/s/KA1AaHvNHKpDGTQ (now a pdf)

TODO
---------
- [x] Add explanatory comments to the three main methods `construct_arguments_generate_assembly_file`, `read_Bilinearform`, `make_file`
- [x] Undo changes to `.github/workflows/testing.yml`
- [x] rename `fast_assembly` to `sum_factorization`
- [x] Investigate varying multiplicity 
- [x] Investigate `field_atoms`
- [x] Delete `api/README.md`, `api/tests/figures/*`, `api/tests/_test_matrix_assembly_3d`, `api/tests/compare_3d_matrix_assembly_speed`
- [x] possibly shorten unit test (`test_sum_factorization_assembly_3d` - currently ~7:45 min.)
- [x] Detach new code generation process from the old one (assigned to @yguclu) 

PR description
---------------------
This PR implements the sum factorization algorithm as shown in **Sum factorization techniques in Isogeometric Analysis** by Andrea Bressan & Stefan Takacs for **bilinear forms on 3D volumes**.

The old element-by-element assembly can still be employed by passing a corresponding flag
```python
# a some BilinearForm
a_h = discretize(a, domain_h, (Vh, Vh), backend=backend, sum_factorization=False)
```

Currently, the new generated code does not make use of OpenMP.

Main changes
--------------
- Write new module `psydac.api.fem_common` which:
   * Contains the old functions previously in module `psydac.api.fem`.
   * Has functions `construct_test_space_arguments` and `construct_trial_space_arguments` returning also the multiplicity.
   * Has new functions `compute_max_nderiv`, `compute_imports`, `compute_free_arguments`.
- Move `DiscreteSumForm` to a new module `psydac.api.fem_sum_form`.
- Create a new module `psydac.api.fem_bilinear_form` which only contains one new class `DiscreteBilinearForm`. The new module shows all imports clearly. The new class:
   * Generates the code to assemble the given discrete bilinear form into a matrix (a `BlockLinearOperator` object of `StencilMatrix` objects, or just a single `StencilMatrix` object), using sum factorization.
   * Does not inherit from `BasicDiscrete` (and hence `BasicCodeGen`). Accordingly, no PSYDAC abstract syntax tree (an object representing a function, of class `DefNode`) is created by the constructor of `AST`, which stores it in its `expr` attribute. (Both classes `DefNode` and `AST` are defined in module `psydac.api.ast.fem`.)
   * Does not create a `Parser` object from `psydac.api.ast.parser`, which used to convert a PSYDAC `DefNode` to an old-Pyccel `FunctionDef` from `psydac.pyccel.ast.core`.
   * Does not generate the old assembly Python code (in the form of a string) using the function `pycode` from `psydac.pyccel.codegen.printing.pycode`.
- Add a new unit test file `api/tests/test_sum_factorization_assembly_3d`.
- Modify function `discretize` in module `psydac.api.discretization` so that, given a bilinear form in 3D and not asking for OpenMP support, it creates an object of type `DiscreteBilinearForm` from module `psydac.api.fem_bilinear_form`. In all other cases, or if `sum_factorization=False`, it creates an object of namesake type from the old module  `psydac.api.fem`.

Other changes
--------------
- Expand docstring of class `AST` in module `psydac.api.ast.fem`.
- Expand docstrings of function `parse` and class `Parser` in module `psydac.api.ast.parser`. Clean up `Parser.__init__`.
- Minor cleanup in class `BasicCodeGen` in module `psydac.api.basic`.
- Generate random filenames using `random.choice()` instead of `random.SystemRandom().choice()`.

Unrelated to matrix assembly:
- Add method `set_scalar` to `ScaledLinearOperator`.
- Reimplement method `idot` of `LinearOperator` using local storage. This avoids creating unnecessary temporary vectors (especially beneficial for the method `dot` of `SumLinearOperator`).


jowezarek:
I have added a file highlighting a bug located somewhere in `DiscreteBilinearForm.allocate_matrices`.
This function generates empty `StencilMatrices` which we write into when assembling the matrix corresponding to the bilinear form.

It appears that the dimensions of the empty `StencilMatrix` are wrong in a very specific edge case:

Given a 3D de Rham sequence, let $u\in V_0$ and $f\in V_3$.
Now, consider `a = BilinearForm((u, f), integral(domain, u*f))` and `a = BilinearForm((f, u), integral(domain, u*f))`.
The resulting matrices should be identitcal up to transposition (A = B.T or equivalently A.T = B where A, B are the two assembled matrices). This however is not the case.

```
# Go to psydac/psydac/api/tests
>>> python allocate_matrix_bug.py
```
The two bilinear forms $(u, f)\mapsto\int_{\Omega} uf$ and $(f,u)\mapsto\int_{\Omega} uf$ are each assembled three times (no mapping, analytical mapping, Bspline mapping). The **relevant** output looks as follows (I removed the Bspline part and the timings):
```
 >>> Mapping Option : None 
 >>> BF u*f 
 >>> ...
 >>>      Error:            6.63 
 >>> Rel. Error:            0.989 
 >>> Norm      : 6.709671 & 9.471318

 >>> Mapping Option : None 
 >>> BF f*u 
 >>> ...
 >>>      Error:            2.37e-15 
 >>> Rel. Error:            2.47e-16 
 >>> Norm      : 9.586167 & 9.586167

 >>> Mapping Option : analytical 
 >>> BF u*f 
 >>> ...
 >>>      Error:            6.63 
 >>> Rel. Error:            0.989 
 >>> Norm      : 6.709671 & 9.471318

 >>> Mapping Option : analytical 
 >>> BF f*u 
 >>> ...
 >>>      Error:            2.26e-15 
 >>> Rel. Error:            2.35e-16 
 >>> Norm      : 9.586167 & 9.586167
```
`Norm` reports the L2 norm of 1. the assembled matrix using the **old** assembly algorithm and 2. the **new** assembly algorithm. Only in the case `f*u` ($\Leftrightarrow f\in V_3$ is the trial function and $u\in V_0$ is the test function) are the two matrices identitcal. This could be due to the new algorithm being implemented wrong. However, both **old and new** matrices have different norms depending on which function is test and which function is trial function, compare 6.70 vs 9.58 and 9.47 vs 9.58. This should not be the case!

```
# In line 69, change `fix_bug = False` to `fix_bug = True`
>>> python python allocate_matrix_bug.py
```
**This is not an actual bug fix!** See `psydac/api/fem.py` for where I use the variable `fix_bug` to circumvent the problem.

The output now looks as follows:
```
 >>> Mapping Option : None 
 >>> BF u*f 
 >>> ...
 >>>      Error:            2.34e-15 
 >>> Rel. Error:            2.44e-16 
 >>> Norm      : 9.586167 & 9.586167

 >>> Mapping Option : None 
 >>> BF f*u 
 >>> ...
 >>>      Error:            2.37e-15 
 >>> Rel. Error:            2.47e-16 
 >>> Norm      : 9.586167 & 9.586167

 >>> Mapping Option : analytical 
 >>> BF u*f 
 >>> ...
 >>>      Error:            2.46e-15 
 >>> Rel. Error:            2.57e-16 
 >>> Norm      : 9.586167 & 9.586167

 >>> Mapping Option : analytical 
 >>> BF f*u 
 >>> ...
 >>>      Error:            2.26e-15 
 >>> Rel. Error:            2.35e-16 
 >>> Norm      : 9.586167 & 9.586167
```
Great! 9.58 is the true L2 norm of the matrix we are trying to assemble.
@yguclu, could you have a look at the `allocate_matrices` method of `DiscreteBilinearForm` please? Maybe you have a feeling for where the error might be hidden! Also @yguclu, could you please have a look at the new `allocate_matrix_bug.py` file and tell me whether you think the way I use analytical and Bspline mappings is correct?

psydac/api/tests/_test_matrix_assembly_3d.py
```diff
@@ -0,0 +1,603 @@
...
+                elif field_space == 'V2':
+                    A_old = a_h.assemble(F2=field)
+                elif field_space == 'V3':
+                    A_old = a_h.assemble(F3=field)
+                elif field_space == 'Vs':
+                    A_old = a_h.assemble(Fs=field)
+                elif field_space == 'Vvc':
+                    A_old = a_h.assemble(Fvc=field)
+                elif field_space == 'Vvd':
+                    A_old = a_h.assemble(Fvd=field)
```
yguclu, line 499: A bit shorter:
```suggestion
                field_name = field_space.replace('V', 'F')
                kwargs = {field_name : field}
                A_old = a_h.assemble(**kwargs)
```

psydac/api/tests/_test_matrix_assembly_3d.py
```diff
@@ -0,0 +1,603 @@
...
+                elif field_space == 'V2':
+                    A_new = a_h.assemble(F2=field)
+                elif field_space == 'V3':
+                    A_new = a_h.assemble(F3=field)
+                elif field_space == 'Vs':
+                    A_new = a_h.assemble(Fs=field)
+                elif field_space == 'Vvc':
+                    A_new = a_h.assemble(Fvc=field)
+                elif field_space == 'Vvd':
+                    A_new = a_h.assemble(Fvd=field)
```
yguclu, line 543: Also here:
```suggestion
                field_name = field_space.replace('V', 'F')
                kwargs = {field_name : field}
                A_new = a_h.assemble(**kwargs)
```

psydac/api/tests/test_sum_factorization_assembly_3d.py
```diff
@@ -0,0 +1,644 @@
...
+    
+    a_sym_h = discretize(a_sym, domain_h, (V1_h, V1_h), backend=backend)
+    a_fem_h = discretize(a_fem, domain_h, (V1_h, V1_h), backend=backend)
+
+    A_sym = a_sym_h.assemble()
+    A_fem = a_fem_h.assemble(p=p_fem)
+    
+    diff = (A_sym-A_fem).tosparse()
+
+    assert abs(diff.data).max() < 4e-2
```
yguclu, line 585: It is hard for me to understand if this is a large error or not. Maybe you could compute `abs(A_sym.data).max()` and `abs(A_fem.data).max()`, and get a relative error?

psydac/api/discretization.py
```diff
@@ -547,6 +547,20 @@ def discretize(a, *args, **kwargs):
...
+        backend = kwargs.get('backend')# or None
+        assembly_backend = kwargs.get('assembly_backend')# or None
+        assembly_backend = backend or assembly_backend
+        openmp = False if assembly_backend is None else assembly_backend.get('openmp')
+
+        # Set sum_factorization to True if domain.dim==3, a is a symbolic BilinearForm and the code is not openMP parallelized ...
+        if (((isinstance(a, sym_BilinearForm)) and (dim == 3))) and not openmp:
+            # ... unless we already explicitely set sum_factorization to False! (For test cases, or in case of a newly encountered bug)
+            if not (kwargs.get('sum_factorization') == False):
+                kwargs['sum_factorization'] = True
```
yguclu, line 563: The old assembly code used a well-designed OpenMP parallelization of the vanilla assembly algorithm. The parallel scaling was great, but it started from a single-process performance which is much much slower than the new one.

So I think that there are several näive ways to use OpenMP in the sum factorization algorithm, and in most cases the resulting code will be faster than the old one. This should not take much time at all to implement, and I will be happy to help.

psydac/api/ast/parser.py
```diff
@@ -548,6 +558,7 @@ def _visit_DefNode(self, expr, **kwargs):
         imports       = [Import('numpy', numpy_imports)] + \
                         ([Import(math_library, math_imports)] if math_imports else []) + \
                         [*expr.imports]
+        self._imports_string = '\n    '.join([str(i) for i in imports])
```
yguclu, line 561: After a bit of digging, I can confirm that there is no good reason for getting the imports from the abstract syntax tree of the old assembly code for bilinear forms. 

Instead, we should write a new function with the lines 555-561 above after some cleanup. Here `expr` is a `DefNode` object (which represents a function) and the contents of the list `expr.imports` are computed at the [very end](https://github.com/pyccel/psydac/blob/a85e4a8203ca66f68600a66597d562da375e5bbd/psydac/api/ast/fem.py#L1191-L1195) of `AST._create_bilinear_form`:
```python
    # Add Import OpenMP if used
    local_vars = []
    imports    = []
    if add_openmp:
        imports.append(Import('pyccel.stdlib.internal.openmp',('omp_get_thread_num', )))
 ```
As you can see, `expr.imports` is empty by default, and contains just one import if OpenMP is used 😜

File: psydac/api/tests/_test_matrix_assembly_3d.py [hunk 6]
```diff
@@ -0,0 +1,603 @@
...
+                    c1 = fields[0].coeffs
+                    c2 = fields[1].coeffs
+                    equil = c1.inner(A_old @ c2)
+                    print(f'old equil: {equil}')
+                elif field_spaces == ['Vs', 'Vs']:
+                    t0 = time.time()
+                    A_old = a_h.assemble(Fs=fields[0], Fs2=fields[1])
+                    t1 = time.time()
+                else:
+                    raise NotImplementedError('This special case must still be taken care of.')
```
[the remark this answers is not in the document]
jowezarek: No, not at all :)
See for example [line 208](https://github.com/pyccel/psydac/blob/eb79f0b4a5fb5d86abfd24e55303435a4d686894/psydac/api/tests/test_sum_factorization_assembly_3d.py#L208) in `test_sum_factorization_assembly_3d.py`.
There, the four free FemFields (`Fvc1_field`, `Fs1_field`, `Fs2_field`, `Fvd1_field`) (read as vector-curl 1, scalar 1, scalar 2, vector-divergence 1) belong to three different spaces.

psydac/api/tests/test_sum_factorization_assembly_3d.py
jowezarek: @e-moral-sanchez Could you have a look into this please?

psydac/api/ast/parser.py
```diff
@@ -104,29 +104,39 @@ def is_scalar_array(var):
...
 
     """
     psy_parser = Parser(settings, backend)
     ast = psy_parser.doit(expr)
-    return ast
+    if not sum_factorization:
+        return ast
+    else:
+        imports_string = psy_parser._imports_string
+        return imports_string
```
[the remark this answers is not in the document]
jowezarek: Probably not, but I don't now. I certainly tried to avoid this.
The issue is related to `self._math_functions` and `expr`:

I have spent a long time trying to obtain the imports information a different way, but eventually gave up.
I know that the import information is generated in `_visit_DefNode`, which, just as all the other `_visit_...` functions, takes `expr` as argument. 
I tried to understand how this specific `expr` argument looks like, in order to then copy only the relevant part of the `_visit_DefNode`, but the recursive structure of this module made that absolutely impossible.

```python
# This is the relevant part of _visit_DefNode
math_library  = 'cmath' if expr.domain_dtype=='complex' else 'math' # Function names are the same
math_imports  = (*self._math_functions,)
numpy_imports = ('array', 'zeros', 'zeros_like', 'floor')
imports       = [Import('numpy', numpy_imports)] + \
                ([Import(math_library, math_imports)] if math_imports else []) + \
                [*expr.imports]
```
`_math_functions` is computed at an entirely different place (`_visit_ComputeKernelExpr`). And as mentioned before, I did not manage to obtain the right `expr` from which to read `expr.imports` and `expr.domain_dtype`.

psydac/api/ast/parser.py
jowezarek: I should have read this comment before replying to the previous one :')
I would be fantastic if you could write this new function! As mentioned above, I gave up trying to do so...

yguclu: The function that we need already exists, and it is [`math_atoms_as_str`](https://github.com/pyccel/psydac/blob/25b46b2816600bbfc96928896ec04731e4a4a4a5/psydac/api/ast/utilities.py#L1056-L1119) (terrible name!) from `psydac/api/ast/utilities.py`:
```python
def math_atoms_as_str(expr, lib='math'):
    """
    Given a Sympy expression, find all known mathematical atoms (functions and
    constants) that need to be imported from a math library (e.g. Numpy) when
    generating Python code.

    Parameters
    ----------
    expr : sympy.core.expr.Expr
        Symbolic expression for which Python code is to be generated.

    lib : str
        Library used to translate symbolic functions/constants into standard
        Python ones. Options: ['math', 'mpmath', 'numpy']. Default: 'math'.

    Returns
    -------
    imports : set of str
        Set of all names (strings) to be imported.

    """
```

yguclu: I am writing a new `DiscreteBilinearForm` class which uses sum factorization only, and does not inherit from `BasicCodeGen` or `BasicDiscrete`. I will keep you posted.

psydac/api/fem.py
```diff
@@ -358,9 +388,18 @@ def __init__(self, expr, kernel_expr, domain_h, spaces, *, nquads,
...
                        nquads=nquads, is_rational_mapping=is_rational_mapping, mapping=symbolic_mapping,
-                       mapping_space=mapping_space, num_threads=self._num_threads, backend=assembly_backend)
+                       mapping_space=mapping_space, num_threads=self._num_threads, backend=assembly_backend,
+                       sum_factorization=sum_factorization)
+        
+        # If sum_factorization == True, broadcast the import information (sqrt, sin, pi, ...) to all processes
+        if (comm is not None) and (comm.size > 1) and (sum_factorization==True):
+            if comm.rank != 0:
+                self._imports_string = None
+            self._imports_string = comm.bcast(self._imports_string, root=0)
```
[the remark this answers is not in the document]
yguclu: Done with [7424d1d](https://github.com/pyccel/psydac/pull/448/commits/7424d1de8fccc2026274c587a4a06478e6c4754a)

psydac/api/ast/parser.py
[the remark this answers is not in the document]
yguclu: See [7424d1d](https://github.com/pyccel/psydac/pull/448/commits/7424d1de8fccc2026274c587a4a06478e6c4754a)

psydac/api/fem.py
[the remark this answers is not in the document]
yguclu: See PR description for some details.

psydac/api/ast/parser.py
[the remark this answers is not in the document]
yguclu: See PR description for some details.

psydac/api/tests/test_sum_factorization_assembly_3d.py
jowezarek: c16082389effb524b2a59d62006f8da015220811

Issue #451, opened 2024-12-04
GMRES solver fails when solving Ax=0 with default inital guess x0=0


spossann:
The GMRES solver has a division by zero when computing the initial residual in this case.

Minimal example:
```
import numpy as np
from psydac.linalg.solvers import inverse
from psydac.linalg.tests.test_solvers import define_data, define_data_hermitian

n = 5
p = 2

# cg solve of Ax = 0 with default initial guess
V, A, xe = define_data_hermitian(n, p)
solver = 'cg'
solv = inverse(A, solver, tol=1e-13)
b = A.domain.zeros()
x = solv.dot(b)
assert np.max(np.abs(x.toarray())) == 0.

# gmres solve of Ax = 0 with default initial guess
diagonals = [-7,-1,-3]
V, A, xe = define_data(n, p, diagonals)
solver = 'gmres'
solv = inverse(A, solver, tol=1e-13)
b = A.domain.zeros()
x = solv.dot(b)  # fails here
```

Proposed solution:

Handle the case `b=0` separately by immediately returning the solution `x=0`?

PR #452, opened 2024-12-04
fix bug gmres

e-moral-sanchez:
Solves issue https://github.com/pyccel/psydac/issues/451.


spossann (approved):
That looks good to me!

Issue #330, opened 2023-09-20
Change name of `VectorSpace.dot()` into `VectorSpace.inner()`


yguclu:
Align names with Wikipedia and FEniCSx (see [documentation](https://fenics.readthedocs.io/projects/ufl/en/latest/manual/form_language.html#tensor-algebra-operators)). Specifically, rename the method `dot` as `inner`:

- [x] In the base classes `VectorSpace` and `Vector` of module `psydac.linalg.basic`;
- [x] In the derived class `BlockVectorSpace` of module `psydac.linalg.block`;
- [x] In the derived class `StencilVectorSpace` of module `psydac.linalg.stencil`;
- [x] In the derived class `DenseVectorSpace` of module `psydac.polar.dense`.

Moreover:
- [x] In module `psydac.api.ast.linalg`, rename the class `VectorDot` as `VectorInner`.

PR #455, opened 2025-02-03
Small fixes for operators with complex

e-moral-sanchez:
- Make `ScaledLinearOperator` work with a complex scalar;
- Take real part of `r.dot(r)` to avoid complex warning in `sqrt`;
- Add unit test.

### DIFF
--- a/psydac/linalg/basic.py
+++ b/psydac/linalg/basic.py
@@ -546,7 +546,7 @@ class ScaledLinearOperator(LinearOperator):
         assert isinstance(domain, VectorSpace)
         assert isinstance(codomain, VectorSpace)
         assert np.isscalar(c)
-        assert(np.isreal(c) or np.iscomplex(c) == (codomain._dtype == complex))
+        assert np.iscomplexobj(c) == (codomain._dtype == complex)
         assert isinstance(A, LinearOperator)
         assert domain   == A.domain
         assert codomain == A.codomain
--- a/psydac/linalg/basic.py
+++ b/psydac/linalg/basic.py
@@ -291,7 +291,6 @@ class LinearOperator(ABC):
 
         """
         assert np.isscalar(c)
-        assert np.isreal(c)
         if c==0:
             return ZeroOperator(self.domain, self.codomain)
         elif c == 1:
@@ -538,7 +537,7 @@ class IdentityOperator(LinearOperator):
 #===============================================================================
 class ScaledLinearOperator(LinearOperator):
     """
-    A linear operator $A$ scalar multiplied by a real constant $c$. 
+    A linear operator $A$ scalar multiplied by a constant $c$. 
     
     """
 
@@ -547,7 +546,7 @@ class ScaledLinearOperator(LinearOperator):
         assert isinstance(domain, VectorSpace)
         assert isinstance(codomain, VectorSpace)
         assert np.isscalar(c)
-        assert np.isreal(c)
+        assert(np.isreal(c) or np.iscomplex(c) == (codomain._dtype == complex))
         assert isinstance(A, LinearOperator)
         assert domain   == A.domain
         assert codomain == A.codomain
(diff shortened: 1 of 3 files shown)


yguclu:
An important matter to be decided is whether the dot product should always return a positive real value. If we call it `inner` product, then I think yes! See #330

psydac/linalg/basic.py
```diff
@@ -547,7 +546,7 @@ def __init__(self, domain, codomain, c, A):
         assert isinstance(domain, VectorSpace)
         assert isinstance(codomain, VectorSpace)
         assert np.isscalar(c)
-        assert np.isreal(c)
+        assert(np.isreal(c) or np.iscomplex(c) == (codomain._dtype == complex))
```
yguclu, line 549: I am not sure that NumPy's `iscomplex` is the correct function to be used here. According to its docstring, "What is tested is whether the input has a non-zero imaginary part, not if the input type is complex."

yguclu (changes requested):
Are there any unit tests which verify the correctness of the proposed changes?

psydac/linalg/solvers.py
```diff
@@ -946,7 +946,7 @@ def solve(self, b, out=None):
         rp.copy(out=rp0)
 
         # squared residual norm and squared tolerance
-        res_sqr = r.dot(r)
+        res_sqr = r.dot(r).real
```
campospinto, line 949: if `r = 1+1j`, then this will probably give `res_sqr = 0` (assuming that `dot` is a regular product). I don't think this will be correct. An inner product should be used instead.

campospinto (commented):
I agree with Yaman's [comment](https://github.com/pyccel/psydac/pull/455#issuecomment-2653493372),
I think that an `inner` product should be used instead of taking the real part of the `dot` product

psydac/linalg/solvers.py
e-moral-sanchez: At the moment, the `dot` product of Psydac  does the complex conjugate when the underlying space field is complex. 
So in the case `r=v+1j*v`  (`v` is a `Vector` with only ones), then `res_sqr = 2*size(v) + 0j` .
In other to avoid the warning when taking `np.sqrt`, we can take the real part.

campospinto: Ok, so why not put the real conversion in the `dot` product (and call it `inner` for clarity) ?

psydac/linalg/basic.py
```diff
@@ -594,7 +593,7 @@ def tosparse(self):
         return self._scalar*csr_matrix(self._operator.toarray())
 
     def transpose(self, conjugate=False):
-        return ScaledLinearOperator(domain=self.codomain, codomain=self.domain, c=self._scalar, A=self._operator.transpose(conjugate=conjugate))
+        return ScaledLinearOperator(domain=self.codomain, codomain=self.domain, c=np.conjugate(self._scalar), A=self._operator.transpose(conjugate=conjugate))
```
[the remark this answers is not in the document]
e-moral-sanchez: Fixed in https://github.com/pyccel/psydac/pull/455/commits/0b16e3f598425145c5a1630109ef9d82b12d487b

psydac/linalg/basic.py
e-moral-sanchez: Fixed in https://github.com/pyccel/psydac/pull/455/commits/db085e6596e413f5ff7df3f8a6d18a59b29e888d

psydac/linalg/solvers.py
e-moral-sanchez: I think that this is quite a technical change, so in my opinion it would require a different PR. It would impact many places in the code and has to be consistent with Sympde. Maybe we can discuss it in the next Psydac meeting.

psydac/linalg/tests/test_linalg.py
```diff
@@ -522,7 +530,7 @@ def test_in_place_operations(n1, n2, p1, p2, P1=False, P2=False):
     assert isinstance(Z3, StencilMatrix)
     assert isinstance(T, StencilMatrix)
     assert np.array_equal(w2.toarray(), np.dot(np.dot(2, Sa), v_array))
-    
+test_in_place_operations(10, 10, 2, 2, P1=False, P2=False)
```
yguclu: If you really want this line to be run interactively with
```shell
$ python3 test_linalg.py
```
then you should move it to the bottom of the file:
```python
if __name__ == "__main__":
    ...
    test_in_place_operations(10, 10, 2, 2, P1=False, P2=False)
```
I am not sure that this is useful though.

yguclu (changes requested):

psydac/linalg/tests/test_linalg.py
yguclu: Sorry I must have commented on an old commit.

yguclu (approved):

PR #458, opened 2025-02-11
Pyccelize evaluate_dofs to speed up tests

kvrigor:
*TODO*


yguclu:
Pyccelized modules can call Python functions, if these are also Pyccelized...

kvrigor:
> Pyccelized modules can call Python functions, **if these are also Pyccelized**...

That's the catch -- those Python functions (*i.e.* f-evals in our case) should remain black box from `C/ForTran` perspective, and thus should not be `pyccel`-ized. Surely not optimal, but I think in this case preserving the original f-eval is more important. There should be a way to bind Python methods as a function pointers in `C` . I asked this question in the Pyccel repo: https://github.com/pyccel/pyccel/discussions/2213#discussioncomment-12169877

Issue #331, opened 2023-09-20
Make sure that multi-patch algorithms work on single-patch domains


yguclu:
The title says it all. This can be fixed once #457 is done.

yguclu:
@campospinto Related to #457?

campospinto:
> [@campospinto](https://github.com/campospinto) Related to [#457](https://github.com/pyccel/psydac/issues/457)?

Asolutely! Let me change the description.

PR #468, opened 2025-03-05
Improve FEM API (names, properties, constructors...) & plot_field()

### LINKED ISSUE #459: Replace`FemSpace.spaces` and `FemField.fields` with more specific attributes
Let `Vh` be a `FemSpace`. Currently `Vh.spaces` may return different types of spaces, which makes the code not always easy to read, and hampers the single/multi-patch compatibility. Instead of this attribute, we should provide **more specific helper attributes** which always return a `tuple`:

 -   `Vh.patch_spaces`:
     + if `Vh` is a multi-patch space (ie `ProductFemSpace`), return the FEM spaces `(Vh_0, Vh_1, ...)` on each patch
     + if `Vh` is single-patch (anything else), return `(self,)`

 -   `Vh.component_spaces`:
     + if `Vh` is a multi-patch space (ie `ProductFemSpace`), raise Error?
     + if `Vh` is single-patch, vector-valued (ie `VectorFemSpace`), return the FEM spaces for each component `k=0, 1, ...`
     + if `Vh` is single-patch, scalar-valued (ie `TensorFemSpace`), return `(self,)`

 -   `Vh.axis_spaces`:
     + if `Vh` is single-patch, scalar-valued, tensor-product (ie `TensorFemSpace`), return the 1D FEM spaces for each axis `i=0, 1, ...`
     + else raise Error?

-------

A similar problem occurs for a multipatch/vector-valued `FemField` `uh` (`uh.fields` may correspond to patch or component fields). Instead of this attribute, we should provide **more specific helper attributes** which always return a `tuple`:

  -  `uh.patch_fields`:
     + if `uh` is a multi-patch field, return the FEM fields `(Vh_0, Vh_1, ...)` on each patch
     + else return `(self,)`

   - `Vh.component_fields`:
    mimic behavior of `Vh.component_spaces`?

### LINKED ISSUE #460: Rename and improve `ProductFemSpace` class
1. The name [ProductFemSpace](https://github.com/pyccel/psydac/blob/74c1271e17394da0bf0e544c9a9f469c37ccf139/psydac/fem/vector.py#L355) is ambiguous: seems to be used only for multi-patch spaces, but a space of vector-valued fields (a `VectorFemSpace`) is also a product of FEM spaces...

    -> We should rename it as `MultipatchFemSpace`
   
2. The "product" ambiguity is made worse by the flag function [Vh.is_product](https://github.com/pyccel/psydac/blob/74c1271e17394da0bf0e544c9a9f469c37ccf139/psydac/fem/basic.py#L62) which means either multi-patch or vector-valued (see for instance in the `FemField` constructor where its is understood as vector-valued...)

    -> We should replace it by `Vh.is_multipatch` and `Vh.is_vector_valued`
   
3. The `ProductFemSpace` [constructor](https://github.com/pyccel/psydac/blob/74c1271e17394da0bf0e544c9a9f469c37ccf139/psydac/fem/vector.py#L368) may return something other than a `ProductFemSpace`, so the code is not easy to follow.

    -> We should use a factory function instead, with a clear docstring

campospinto:
Improve the FEM API to make it more user-friendly and compatible between single-patch and multi-patch APIs (fix #459 and fix #460):

* Rename abstract property `vector_space` of `FemSpace` as `coeff_space`
* Rename `ProductFemSpace` as `MultipatchFemSpace`
* Do not use class constructors as factories (previously `ProductFemSpace()` could return a `VectorFemSpace` object)
* Provide more specific helper properties for FEM spaces, such as
   - `is_vector_valued` and `is_multipatch` (also remove a method `is_scalar` which was implemented in _some_ FEM spaces but not used)
   - `patch_spaces`, `component_spaces` and `axis_spaces`  which always return tuples of Fem spaces (or an error)
* Provide more specific helper properties for FEM fields, such as
   - `component_fields` and `patch_fields` and `axis_fields` which always return tuples of FEM fields (or an error)
* Improve the `VectorFemSpace` class (in particular its constructor) to better match the `FemSpace` abstract interface.

Since these changes are also motivated by a better compatibility between single-patch and multi-patch APIs (see issue #331), this PR also modifies the `plot_field` function (previously in `psydac.feec.multipatch.plotting_utilities`), which is:

* made compatible with single patch and multi patch fields
* renamed as `plot_field_2d` since it only handles 2d fields
* tested (not the content but the runs) in single and multipatch configurations for scalar and vector-valued FEM fields
* moved with its module `psydac.feec.multipatch.plotting_utilities` to `psydac.fem.plotting_utilities` since it is not specific to FEEC or multipatch

### DIFF
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -503,11 +503,22 @@ def discretize_space(V, domain_h, *, degree=None, multiplicity=None, knots=None,
         new_g_spaces[inter]   = Vh
 
     construct_reduced_interface_spaces(g_spaces, new_g_spaces, interiors, connectivity)
+    spaces = list(new_g_spaces.values())
 
-    Vh = create_product_space(*new_g_spaces.values(), connectivity=connectivity)
+    if len(connectivity) == 0:
+        assert all(isinstance(Wh, FemSpace) for Wh in spaces)
+        if  len(spaces) == 1:
+            Vh = spaces[0]
+        else:
+            Vh = VectorFemSpace(*spaces)
+    else:
+        assert all((isinstance(Wh, FemSpace) and not Wh.is_multipatch) for Wh in spaces)
+        Vh = MultipatchFemSpace(*spaces, connectivity=connectivity)
+    
     Vh.symbolic_space = V
 
     return Vh
+
 
 #==============================================================================
 def discretize_domain(domain, *, filename=None, ncells=None, periodic=None, comm=None, mpi_dims_mask=None):
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -78,13 +78,13 @@ def change_dtype(V, dtype):
             for v in V.spaces:
                 change_dtype(v, dtype)
                 new_spaces.append(v.vector_space)
-            V._vector_space = BlockVectorSpace(*new_spaces, connectivity=V.vector_space.connectivity)
+            V._coeff_space = BlockVectorSpace(*new_spaces, connectivity=V.vector_space.connectivity)
 
         # If the vector_space is a StencilVectorSpace
         else:
             # Recreate the StencilVectorSpace
             interfaces = V.vector_space.interfaces
-            V._vector_space = StencilVectorSpace(V.vector_space.cart, dtype=dtype)
+            V._coeff_space = StencilVectorSpace(V.vector_space.cart, dtype=dtype)
 
             # Recreate the interface in the StencilVectorSpace
             for (axis, ext), interface_space in interfaces.items():
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -37,7 +37,7 @@ from psydac.api.utilities    import flatten
 from psydac.fem.splines      import SplineSpace
 from psydac.fem.tensor       import TensorFemSpace
 from psydac.fem.partitioning import create_cart, construct_connectivity, construct_interface_spaces, construct_reduced_interface_spaces
-from psydac.fem.vector       import ProductFemSpace, VectorFemSpace
+from psydac.fem.vector       import MultipatchFemSpace, VectorFemSpace
 from psydac.cad.geometry     import Geometry
 from psydac.mapping.discrete import NurbsMapping
 from psydac.linalg.stencil   import StencilVectorSpace
@@ -115,7 +115,7 @@ def get_max_degree_of_one_space(Vh):
     elif isinstance(Vh, VectorFemSpace):
         return [max(p) for p in zip(*Vh.degree)]
 
-    elif isinstance(Vh, ProductFemSpace):
+    elif isinstance(Vh, MultipatchFemSpace):
         degree = [get_max_degree_of_one_space(Vh_i) for Vh_i in Vh.spaces]
         return [max(p) for p in zip(*degree)]
 
@@ -504,7 +504,7 @@ def discretize_space(V, domain_h, *, degree=None, multiplicity=None, knots=None,
 
     construct_reduced_interface_spaces(g_spaces, new_g_spaces, interiors, connectivity)
 
-    Vh = ProductFemSpace(*new_g_spaces.values(), connectivity=connectivity)
+    Vh = MultipatchFemSpace(*new_g_spaces.values(), connectivity=connectivity)
     Vh.symbolic_space = V
 
     return Vh
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -37,7 +37,7 @@ from psydac.api.utilities    import flatten
 from psydac.fem.splines      import SplineSpace
 from psydac.fem.tensor       import TensorFemSpace
 from psydac.fem.partitioning import create_cart, construct_connectivity, construct_interface_spaces, construct_reduced_interface_spaces
-from psydac.fem.vector       import MultipatchFemSpace, VectorFemSpace
+from psydac.fem.vector       import MultipatchFemSpace, VectorFemSpace, create_product_space
 from psydac.cad.geometry     import Geometry
 from psydac.mapping.discrete import NurbsMapping
 from psydac.linalg.stencil   import StencilVectorSpace
@@ -504,7 +504,7 @@ def discretize_space(V, domain_h, *, degree=None, multiplicity=None, knots=None,
 
     construct_reduced_interface_spaces(g_spaces, new_g_spaces, interiors, connectivity)
 
-    Vh = MultipatchFemSpace(*new_g_spaces.values(), connectivity=connectivity)
+    Vh = create_product_space(*new_g_spaces.values(), connectivity=connectivity)
     Vh.symbolic_space = V
 
     return Vh
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -37,7 +37,7 @@ from psydac.api.utilities    import flatten
 from psydac.fem.splines      import SplineSpace
 from psydac.fem.tensor       import TensorFemSpace
 from psydac.fem.partitioning import create_cart, construct_connectivity, construct_interface_spaces, construct_reduced_interface_spaces
-from psydac.fem.vector       import MultipatchFemSpace, VectorFemSpace, create_product_space
+from psydac.fem.vector       import MultipatchFemSpace, VectorFemSpace
 from psydac.cad.geometry     import Geometry
 from psydac.mapping.discrete import NurbsMapping
 from psydac.linalg.stencil   import StencilVectorSpace
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -55,7 +55,7 @@ __all__ = (
 #==============================================================================
 def change_dtype(V, dtype):
     """
-    Given a FemSpace V, change its underlying vector_space (i.e. the space of
+    Given a FemSpace V, change its underlying coeff_space (i.e. the space of
     its coefficients) so that it matches the required data type.
 
     Parameters
@@ -64,31 +64,31 @@ def change_dtype(V, dtype):
         The FEM space, which is modified in place.
 
     dtype : float or complex
-        Datatype of the new vector_space.
+        Datatype of the new coeff_space.
 
     Returns
     -------
     FemSpace
         The same FEM space passed as input, which was modified in place.
     """
-    if not V.vector_space.dtype == dtype:
-        if isinstance(V.vector_space, BlockVectorSpace):
+    if not V.coeff_space.dtype == dtype:
+        if isinstance(V.coeff_space, BlockVectorSpace):
             # Recreate the BlockVectorSpace
             new_spaces = []
             for v in V.spaces:
                 change_dtype(v, dtype)
-                new_spaces.append(v.vector_space)
-            V._coeff_space = BlockVectorSpace(*new_spaces, connectivity=V.vector_space.connectivity)
+                new_spaces.append(v.coeff_space)
+            V._coeff_space = BlockVectorSpace(*new_spaces, connectivity=V.coeff_space.connectivity)
 
-        # If the vector_space is a StencilVectorSpace
+        # If the coeff_space is a StencilVectorSpace
         else:
             # Recreate the StencilVectorSpace
-            interfaces = V.vector_space.interfaces
-            V._coeff_space = StencilVectorSpace(V.vector_space.cart, dtype=dtype)
+            interfaces = V.coeff_space.interfaces
+            V._coeff_space = StencilVectorSpace(V.coeff_space.cart, dtype=dtype)
 
             # Recreate the interface in the StencilVectorSpace
             for (axis, ext), interface_space in interfaces.items():
-                V.vector_space.set_interface(axis, ext, interface_space.cart)
+                V.coeff_space.set_interface(axis, ext, interface_space.cart)
 
     return V
 
@@ -417,7 +417,7 @@ def discretize_space(V, domain_h, *, degree=None, multiplicity=None, knots=None,
         else:
             mappings  = [domain_h.mappings[inter.logical_domain.name] for inter in interiors]
 
-        # Get all the FEM spaces from the mapping and convert their vector_space at the dtype needed
+        # Get all the FEM spaces from the mapping and convert their coeff_space at the dtype needed
         spaces    = [change_dtype(m.space, dtype) for m in mappings]
         g_spaces  = dict(zip(interiors, spaces))
         spaces    = [S.spaces for S in spaces]
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -34,6 +34,7 @@ from psydac.api.glt          import DiscreteGltExpr
 from psydac.api.expr         import DiscreteExpr
 from psydac.api.equation     import DiscreteEquation
 from psydac.api.utilities    import flatten
+from psydac.fem.basic        import FemSpace
 from psydac.fem.splines      import SplineSpace
 from psydac.fem.tensor       import TensorFemSpace
 from psydac.fem.partitioning import create_cart, construct_connectivity, construct_interface_spaces, construct_reduced_interface_spaces
@@ -505,15 +506,15 @@ def discretize_space(V, domain_h, *, degree=None, multiplicity=None, knots=None,
     construct_reduced_interface_spaces(g_spaces, new_g_spaces, interiors, connectivity)
     spaces = list(new_g_spaces.values())
 
-    if len(connectivity) == 0:
+    if connectivity:
+        assert all((isinstance(Wh, FemSpace) and not Wh.is_multipatch) for Wh in spaces)
+        Vh = MultipatchFemSpace(*spaces, connectivity=connectivity)
+    else:
         assert all(isinstance(Wh, FemSpace) for Wh in spaces)
         if  len(spaces) == 1:
             Vh = spaces[0]
         else:
             Vh = VectorFemSpace(*spaces)
-    else:
-        assert all((isinstance(Wh, FemSpace) and not Wh.is_multipatch) for Wh in spaces)
-        Vh = MultipatchFemSpace(*spaces, connectivity=connectivity)
     
     Vh.symbolic_space = V
 
--- a/psydac/api/postprocessing.py
+++ b/psydac/api/postprocessing.py
@@ -426,7 +426,7 @@ class OutputManager:
         Adds a vector space to the scope.
         Parameters
         ----------
-        vector_space: psydac.fem.vector.VectorFemSpace or psydac.fem.vector.ProductFemSpace
+        vector_space: psydac.fem.vector.VectorFemSpace or psydac.fem.vector.MultipatchFemSpace  # todo [MCP 12.03.2025]: check these types
             Vector/Product FemSpace to add to the scope.
         """
 
--- a/psydac/api/postprocessing.py
+++ b/psydac/api/postprocessing.py
@@ -511,7 +511,8 @@ class OutputManager:
         # Add field coefficients as named datasets
         for name_field, field in fields.items():
             multipatch = hasattr(field.space.symbolic_space.domain.interior, 'as_tuple')
-            assert multipatch == f.space.is_multipatch   ## [MCP 27.03.2025] check: if OK, then simplify
+            
+            assert multipatch == field.space.is_multipatch   ## [MCP 27.03.2025] check: if OK, then simplify
             
             if multipatch:
                 for f in field.fields:
@@ -537,7 +538,7 @@ class OutputManager:
                             mpi_dd_gp.create_dataset(f'{name_patch}', shape=(size, *local_domain.shape), dtype='i')
                             mpi_dd_gp[name_patch][rank] = local_domain
 
-                    assert f.space.is_product ## [MCP 27.03.2025] check: if OK, then simplify
+                    assert f.space.is_product == f.space.is_vector_valued ## [MCP 27.03.2025] check: if OK, then simplify
 
                     if f.space.is_product:  # Vector field case ([MCP 27.03.2025]: I don't understand this comment... why vector ?)
                         for i, field_coeff in enumerate(f.coeffs):
--- a/psydac/api/postprocessing.py
+++ b/psydac/api/postprocessing.py
@@ -433,7 +433,7 @@ class OutputManager:
         Parameters
         ----------
         vector_space: psydac.fem.vector.VectorFemSpace
-            Vector/Product FemSpace to add to the scope.
+            Vector FemSpace to add to the scope.
         """
         assert isinstance(vector_space, VectorFemSpace)
 
--- a/psydac/api/postprocessing.py
+++ b/psydac/api/postprocessing.py
@@ -511,7 +511,8 @@ class OutputManager:
         # Add field coefficients as named datasets
         for name_field, field in fields.items():
             multipatch = hasattr(field.space.symbolic_space.domain.interior, 'as_tuple')
-
+            assert multipatch == f.space.is_multipatch   ## [MCP 27.03.2025] check: if OK, then simplify
+            
             if multipatch:
                 for f in field.fields:
                     i = self._spaces.index(f.space)
@@ -525,7 +526,7 @@ class OutputManager:
                             size = self.comm.Get_size()
 
 
-                            if f.space.is_product:
+                            if f.space.is_product:  # [MCP 27.03.2025] always true if multipatch... todo: simplify
                                 sp = f.space.spaces[0]
                             else:
                                 sp = f.space
@@ -536,7 +537,9 @@ class OutputManager:
                             mpi_dd_gp.create_dataset(f'{name_patch}', shape=(size, *local_domain.shape), dtype='i')
                             mpi_dd_gp[name_patch][rank] = local_domain
 
-                    if f.space.is_product:  # Vector field case
+                    assert f.space.is_product ## [MCP 27.03.2025] check: if OK, then simplify
+
+                    if f.space.is_product:  # Vector field case ([MCP 27.03.2025]: I don't understand this comment... why vector ?)
                         for i, field_coeff in enumerate(f.coeffs):
                             name_field_i = name_field + f'[{i}]'
                             name_space_i = name_space + f'[{i}]'
@@ -554,7 +557,7 @@ class OutputManager:
                                                             shape=Vi.npts, dtype=Vi.dtype)
                             dset.attrs.create('parent_field', data=name_field)
                             dset[index] = field_coeff[index]
-                    else:
+                    else:                        
                         V = f.space.vector_space
                         index = tuple(slice(s, e + 1) for s, e in zip(V.starts, V.ends))
                         dset = saving_group.create_dataset(f'{name_patch}/{name_space}/{name_field}', shape=V.npts, dtype=V.dtype)
@@ -570,7 +573,7 @@ class OutputManager:
                         rank = self.comm.Get_rank()
                         size = self.comm.Get_size()
 
-                        if field.space.is_product:
+                        if field.space.is_product:  ## MCP 27.03.2025: simplify if above checks pass
                             sp = field.space.spaces[0]
                         else:
                             sp = field.space
@@ -581,7 +584,7 @@ class OutputManager:
                         mpi_dd_gp.create_dataset(f'{name_patch}', shape=(size, *local_domain.shape), dtype='i')
                         mpi_dd_gp[name_patch][rank] = local_domain
 
-                if field.space.is_product:  # Vector field case
+                if field.space.is_product:  # Vector field case (## MCP 27.03.2025: here indeed I would agree)
                     for i, field_coeff in enumerate(field.coeffs):
                         name_field_i = name_field + f'[{i}]'
                         name_space_i = name_space + f'[{i}]'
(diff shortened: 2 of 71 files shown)


psydac/fem/basic.py
```diff
@@ -125,29 +141,54 @@ def eval_field_gradient( self, field, *eta , weights=None):
...
 
         spaces = [*(   a.spaces if    a.is_product else    [a]),
                   *(self.spaces if self.is_product else [self]),]
 
-        space = ProductFemSpace(*spaces)
+        space = create_product_space(*spaces)
 
         if a.symbolic_space and self.symbolic_space:
             space._symbolic_space =  a.symbolic_space * self.symbolic_space
         return space
```
yguclu, line 165: I am quite puzzled by these two concrete methods, which I cannot recognize. Apparently this code was added by @saidctb 4 years ago when adding support for repeated knots.

In principle, since `FemSpace` is an abstract base class, it should not have any concrete methods. In fact, these two methods are inconsistent because there is no guarantee that a `FemSpace` object will have a `spaces` attribute.

Accordingly, the new properties `patch_spaces` and `component_spaces` cannot be defined in this class as concrete. They currently use the attribute `_spaces` which does not exist.

psydac/api/discretization.py
```diff
@@ -504,7 +504,7 @@ def discretize_space(V, domain_h, *, degree=None, multiplicity=None, knots=None,
 
     construct_reduced_interface_spaces(g_spaces, new_g_spaces, interiors, connectivity)
 
-    Vh = ProductFemSpace(*new_g_spaces.values(), connectivity=connectivity)
+    Vh = create_product_space(*new_g_spaces.values(), connectivity=connectivity)
```
yguclu, line 507: Is this the only place in the library where the factory function `create_product_space` is used?

psydac/api/postprocessing.py
```diff
@@ -426,7 +426,7 @@ def _add_vector_space(self, vector_space, name=None, patch_name=None):
         Adds a vector space to the scope.
         Parameters
         ----------
-        vector_space: psydac.fem.vector.VectorFemSpace or psydac.fem.vector.ProductFemSpace
+        vector_space: psydac.fem.vector.VectorFemSpace or psydac.fem.vector.MultipatchFemSpace  # todo [MCP 12.03.2025]: check these types
```
yguclu, line 429: Is this TO-DO for you or posterity? 😅

psydac/api/tests/test_api_2d_fields.py
```diff
@@ -320,14 +326,15 @@ def test_field_collela():
     assert abs(error_2 - expected_error_2) < 1.e-7
 
 #------------------------------------------------------------------------------
-def test_field_quarter_annulus():
+@pytest.mark.parametrize('do_plot', [False, True])
+def test_field_quarter_annulus(do_plot):
```
yguclu, line 330: If you parametrize the test, `pytest` will show the plot when the unit tests are run. Is this what you wanted? If not, I suggest simply
```suggestion
def test_field_quarter_annulus(do_plot=False):
```
which runs the unit test with `do_plot=False`, but allows you to run it manually with `do_plot=True`.

psydac/feec/global_projectors.py
```diff
@@ -659,7 +659,7 @@ class Projector_H1vec(GlobalProjector):
     
     Parameters
     ----------
-    H1vec : ProductFemSpace
+    H1vec : FemSpace
```
yguclu, line 662: I think `H1vec` should be an object of type `VectorFemSpace`. Or not?

psydac/feec/multipatch/examples/mixed_source_pbms_conga_2d.py
```diff
@@ -413,8 +413,9 @@ def P2_phys(f_phys):
     # bc_type = 'pseudo-vacuum'
     source_type = 'dipole_J'
 
-    source_proj = 'P_L2_wcurl_J'
-
+    # source_proj = 'P_L2_wcurl_J'
+    source_proj = 'P_geom'
+
```
yguclu, line 418: What example do you want to be run manually? Is this the one presented in the paper? If yes, we can maybe add a comment here.

psydac/fem/basic.py
```diff
@@ -125,29 +141,54 @@ def eval_field_gradient( self, field, *eta , weights=None):
...
 
+    @property
+    def patch_spaces(self):
+        """
+        Return the patch spaces (self if single-patch) as a tuple.
+        """
+        if self.is_multipatch:
+            return self._spaces
+        else:
+            return (self,)
```
yguclu, line 175:
```suggestion
    @property
    @abstractmethod
    def patch_spaces(self):
        """
        Return the patch spaces (only self if single-patch) as a tuple.
        """
```

psydac/fem/basic.py
```diff
@@ -125,29 +141,54 @@ def eval_field_gradient( self, field, *eta , weights=None):
...
+        """
+        if self.is_vector_valued:
+            if self.is_multipatch:
+                # should we return here the multipatch scalar-valued space?
+                raise NotImplementedError('Component spaces not implemented for multipatch spaces')
+            else:
+                return self._spaces
+        else:
+            return (self,)
+
```
yguclu, line 190:
```suggestion
    @property
    @abstractmethod
    def component_spaces(self):
        """
        Return the component spaces (only self if scalar-valued) as a tuple.
        """
```

psydac/fem/basic.py
```diff
@@ -265,6 +306,26 @@ def coeffs( self ):
     def fields(self):
         return self._fields
 
+    @property
+    def is_vector_valued(self):
+        return self.space.is_vector_valued
+
+    @property
+    def patch_fields(self):
+        """ return the patch fields (self if single-patch) as a tuple """
```
yguclu, line 315:
```suggestion
        """ Return the patch fields (only self if single-patch) as a tuple """
```

psydac/api/postprocessing.py
campospinto: For me 😄

yguclu:
If I understand correctly this PR introduces the properties `patch_spaces` and `component_spaces` for any `FemSpace`. Does it also introduce the property `axis_spaces` as described in issue #459?

psydac/fem/tests/test_product.py
```diff
@@ -30,7 +30,7 @@ def test_product_space_2d():
     Vy = TensorFemSpace(domain_decomposition, V1, V2)
     # ...
 
-    V = ProductFemSpace(Vx, Vy)
+    V = create_product_space(Vx, Vy)
```
yguclu, line 33: This could be replaced with `VectorFemSpace(Vx, Vy)`, correct?

psydac/fem/tests/test_product.py
```diff
@@ -67,7 +67,7 @@ def test_product_space_3d():
     Vz = TensorFemSpace(domain_decomposition, V1, V2, V3)
     # ...
 
-    V = ProductFemSpace(Vx, Vy, Vz)
+    V = create_product_space(Vx, Vy, Vz)
```
yguclu, line 70: Same question

psydac/fem/tests/test_product.py
yguclu, line 1: @campospinto Do you have an idea of what this file is supposed to test exactly? I do not see any `assert` statement in the test functions.

yguclu (changes requested):

psydac/api/discretization.py
campospinto: Apart from the dubious concrete `__mul__` and `__rmul__` functions in the abstract FemSpace, yes.

psydac/api/postprocessing.py
campospinto: todo -> done

psydac/api/tests/test_api_2d_fields.py
campospinto: Well the point was to test the plot. 
As the argument `hide_plot=True` is passed to `plot_field`, I imagine the plot will be saved to file but not shown.

psydac/feec/multipatch/examples/mixed_source_pbms_conga_2d.py
campospinto: both should be fine, I just put the fastest I guess

psydac/fem/basic.py
campospinto: Yes this is now an abstract method

psydac/fem/basic.py
campospinto: same

campospinto:
> If I understand correctly this PR introduces the properties `patch_spaces` and `component_spaces` for any `FemSpace`. Does it also introduce the property `axis_spaces` as described in issue #459?

It did not but it now does

psydac/feec/multipatch/operators.py
```diff
@@ -748,7 +748,7 @@ def get_K0_and_K0_inv(V0h, uniform_patches=False):
     if uniform_patches:
         print(' [[WARNING -- hack in get_K0_and_K0_inv: using copies of 1st-patch matrices in every patch ]] ')
 
-    V0 = V0h.symbolic_space   # VOh is ProductFemSpace
+    V0 = V0h.symbolic_space   # VOh is FemSpace
```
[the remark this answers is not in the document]
campospinto: Well, not necessarily if we want these functions to also work on single patch spaces

psydac/fem/basic.py
campospinto: The tests seem to pass. I suggest to leave these functions raising a NotImplementedError and if we discover that some code is using them, we'll look for a good solution.

psydac/api/discretization.py
campospinto: (they were also used in some tests, but as you noticed they could be replaced by a plain VectorFemSpace constructor)

psydac/feec/global_projectors.py
campospinto: (I changed the description)

psydac/feec/multipatch/operators.py
```diff
@@ -798,7 +798,7 @@ def get_K1_and_K1_inv(V1h, uniform_patches=False):
     if uniform_patches:
         print(' [[WARNING -- hack in get_K1_and_K1_inv: using copies of 1st-patch matrices in every patch ]] ')
 
-    V1 = V1h.symbolic_space   # V1h is ProductFemSpace
+    V1 = V1h.symbolic_space   # V1h is FemSpace
```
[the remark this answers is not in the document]
campospinto: same answer :)

psydac/feec/multipatch/operators.py
```diff
@@ -858,7 +858,7 @@ def get_K1_and_K1_inv(V1h, uniform_patches=False):
 #     """
 #     from pprint import pprint
 #
-#     V = Vh.symbolic_space   # VOh is ProductFemSpace
+#     V = Vh.symbolic_space   # VOh is FemSpace
```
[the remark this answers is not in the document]
campospinto: same answer :D

psydac/fem/tests/test_product.py
campospinto: absolutely! I replaced it

psydac/fem/tests/test_product.py
campospinto: idem: replaced

psydac/fem/tests/test_product.py
campospinto: I'd say, it checks that different FemSpace constructors run without crashing?

psydac/fem/basic.py
yguclu: Good idea!

psydac/api/discretization.py
yguclu: I see. Then please use the class constructor here if possible

psydac/api/tests/test_api_2d_fields.py
campospinto: I guess you're right! I will put this in a separate PR

psydac/api/discretization.py
campospinto: I just removed the factory function

psydac/fem/splines.py
```diff
@@ -316,11 +312,32 @@ def eval_field_gradient( self, field, *eta , weights=None):
...
+    @property
+    def is_vector_valued(self):
+        return False
+
+    # question [MCP 03.2025]: this is not in the FemSpace interface, redundant with is_vector_valued. keep it ?
+    @property
     def is_scalar( self ):
         """ Only scalar field is implemented for now.
         """
         return True
```
yguclu, line 327: How often is `is_scalar` used?

If we wanted to keep it for convenience, I would turn it into a concrete method of the `FemSpace` abstract base class which does simply
```python
@property
def is_scalar_valued(self):
    return not self.is_vector_valued
```

psydac/fem/tensor.py
```diff
@@ -702,6 +710,16 @@ def integral(self, f, *, nquads=None):
...
+        return False
+
+    @property
+    def is_vector_valued(self):
+        return False
+
+    # question [MCP 03.2025]: this is not in the FemSpace interface, redundant with is_vector_valued. keep it ?
     @property
     def is_scalar(self):
         return True
```
yguclu, line 725: See my previous comment in splines.py

psydac/fem/tests/test_product.py
yguclu: Maybe we can check some property of the spaces?

By the way, I see no longer the need to use the word "product" in the file name, nor in the test functions...

psydac/fem/vector.py
```diff
@@ -304,6 +312,17 @@ def integral( self, f ):
     # Other properties and methods
     #--------------------------------------------------------------------------
     @property
+    def is_multipatch(self):
+        return False
+
+    @property
+    def is_vector_valued(self):
+        # question (MCP 03.2025) do we allow scalar-valued VectorFemSpaces ?
```
yguclu, line 320: What do you mean exactly? A vector space with only one component?

psydac/fem/vector.py
```diff
@@ -304,6 +312,17 @@ def integral( self, f ):
...
+        return True
+
+    # question [MCP 03.2025]: this is not in the FemSpace interface,
+    # and almost redundant/inconsistent with is_vector_valued. keep it ?
+    @property
     def is_scalar(self):
         return len( self.spaces ) == 1
```
yguclu, line 327: See my previous comment in splines.py

psydac/fem/vector.py
```diff
@@ -351,27 +370,22 @@ def __str__(self):
...
 
-    def __new__(cls, *spaces, connectivity=None):
-
-        if len(spaces) == 1:
-            return spaces[0]
-        else:
-            return FemSpace.__new__(cls)
-
-    def __init__( self, *spaces, connectivity=None):
+    def __init__( self, *spaces, connectivity={}):
```
yguclu, line 380: It is dangerous to pass a mutable object (like a dictionary) as a default value to a function, because this gets evaluated when the function is defined. Here I would leave the original `None`.

psydac/fem/basic.py
```diff
@@ -259,12 +304,32 @@ def coeffs( self ):
...
-        
+
     # ...
     @property
     def fields(self):
         return self._fields
 
+    @property
+    def is_vector_valued(self):
+        return self.space.is_vector_valued
```
yguclu, line 315: Do we really need this property in the `FemField` class? One can always get the information from the space as clearly shown by this code.

psydac/fem/tensor.py
```diff
@@ -702,6 +710,16 @@ def integral(self, f, *, nquads=None):
...
     # Other properties and methods
     #--------------------------------------------------------------------------
+
+    @property
+    def is_multipatch(self):
+        return False
+
+    @property
+    def is_vector_valued(self):
+        return False
```
yguclu, line 720: These two properties should be moved up a bit, to the section "Abstract interface: read-only attributes".

psydac/fem/vector.py
```diff
@@ -304,6 +312,17 @@ def integral( self, f ):
     # Other properties and methods
     #--------------------------------------------------------------------------
     @property
+    def is_multipatch(self):
+        return False
+
+    @property
+    def is_vector_valued(self):
+        # question (MCP 03.2025) do we allow scalar-valued VectorFemSpaces ?
+        return True
```
yguclu, line 321: These two properties should be moved up, to the section "Abstract interface: read-only attributes".

psydac/fem/vector.py
```diff
@@ -606,15 +634,23 @@ def eval_fields_irregular_tensor_grid(self, grid, *fields, weights=None, overlap
...
     #--------------------------------------------------------------------------
     # Other properties and methods
     #--------------------------------------------------------------------------
+    @property
+    def is_multipatch(self):
+        return True
+
+    @property
+    def is_vector_valued(self):
+        return self.patch_spaces[0].is_vector_valued
```
yguclu, line 652: These two properties should be moved up, to the section "Abstract interface: read-only attributes".

yguclu (changes requested):

psydac/fem/tests/test_product.py
campospinto: I renamed this file into `test_vector_spaces.py`. As for the test I agree but presently I don't see what to test

psydac/api/postprocessing.py
```diff
@@ -2,6 +2,10 @@
 #
 # Copyright 2019 Yaman Güçlü
 
+# TODO [MCP 03.2025]
+# - ambiguous method is_product is used in many places: it should be replaced by is_vector_valued (probably always) or is_multipatch (probably never)
```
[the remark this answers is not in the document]
campospinto: I see 1 place in `psydac/fem/basic.py`, 4 places in `psydac/api/postprocessing.py` and 5 places in `psydac/api/fem.py`. For several of them, I have no idea if the space is multipatch or vector valued. I'm just moving the function to the concrete area and leave a todo

psydac/fem/splines.py
```diff
@@ -316,11 +312,32 @@ def eval_field_gradient( self, field, *eta , weights=None):
     # Other properties
     #--------------------------------------------------------------------------
     @property
+    def is_multipatch(self):
+        return False
+
+    @property
+    def is_vector_valued(self):
+        return False
```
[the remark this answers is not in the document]
campospinto: done

psydac/fem/splines.py
```diff
@@ -316,11 +312,32 @@ def eval_field_gradient( self, field, *eta , weights=None):
...
+    def patch_spaces(self):
+        return (self,)
+
+    @property
+    def component_spaces(self):
+        return (self,)
+
+    @property
+    def axis_spaces(self):
+        return (self,)
```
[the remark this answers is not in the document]
campospinto: done

psydac/fem/splines.py
campospinto: no idea but good idea

psydac/fem/tests/test_product.py
yguclu: I was thinking of very basic consistency checks. I can do this

psydac/api/postprocessing.py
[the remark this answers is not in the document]
yguclu: Ah, I meant 4 places in this file... so it is 10 places in total 😅 

I would definitely try and remove the use of `is_product` from Psydac completely... Do you see a reason to wait for another PR?

psydac/api/tests/test_api_2d_fields.py
campospinto: (I don't really see how to test the content of the plots)

psydac/fem/basic.py
```diff
@@ -56,21 +56,62 @@ def mapping( self ):
...
+        """
+
+    @property
     def is_product( self ):
         """
         Boolean flag that describes whether the space is a product space.
         If True, an element of this space can be decomposed into separate fields.
 
         """
+        return self.is_multipatch or self.is_vector_valued
```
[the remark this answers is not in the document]
campospinto: moved but kept for now (see above)

psydac/fem/vector.py
campospinto: ok, good point!

psydac/fem/basic.py
campospinto: well it's convenient (and natural) to ask directly if a vector field is vector valued. What is the issue ?

yguclu: Well I prefer to keep the codebase at its necessary minimum, or at least close to it, because:

1. Any "convenience attribute" needs to be used in the unit tests if we want to achieve good coverage;
2. Any later change to the code will be harder to propagate everywhere.

If we can achieve 100% coverage, then any inconsistency **should** be catched with a failing test. And then, modern tools like Copilot **should** make the changes easy.

psydac/fem/vector.py
yguclu: Done in 55fd2c9

psydac/fem/vector.py
yguclu: Done in 55fd2c9

psydac/fem/splines.py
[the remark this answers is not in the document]
yguclu: Where was it done?

psydac/fem/tensor.py
yguclu: Thanks for commenting this property out

psydac/fem/vector.py
yguclu: Thanks for commenting this property out

psydac/fem/vector.py
yguclu: Done in 6490fa3

psydac/fem/tensor.py
yguclu: Done in 55fd2c9

psydac/fem/splines.py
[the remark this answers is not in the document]
campospinto: I did a few commits and then pushed a batch -- should be visible now... ?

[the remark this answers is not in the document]
campospinto: ah sorry I was mixing this with the other read-only attributes, read too fast

psydac/fem/basic.py
campospinto: alright. your answer is longer than the changes in the code so you win 😅

psydac/fem/splines.py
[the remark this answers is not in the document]
campospinto: But then I guess that we should also move the other methods like `basis`, `interpolation_grid`, `nbasis`, `degree`... right ?

psydac/fem/basic.py
```diff
@@ -56,21 +56,53 @@ def mapping( self ):
     @abstractmethod
     def vector_space( self ):
         """Topologically associated vector space."""
+        # question [MCP 03.2025]:
+            # rename as coeff_space ?
+            # meaning of docstring? see different one in MultipatchFemSpace
```
yguclu, line 61: I agree, we should rename this property! `coeff_space` sounds good to me!

campospinto: 👍  in this PR?

psydac/fem/splines.py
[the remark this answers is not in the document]
campospinto: no but some should probably be. Let's keep this for another time.

psydac/fem/basic.py
campospinto: Ok I'm doing this here as well -- that's a lot of changes, but very simple (renaming)

campospinto: Ahah, actually I need to be careful: well in some cases `vector_space` actually denotes a vector-valued FemSpace... 🙄

campospinto: There were a lot of changes so I put them in a separate branch that can be merged after this PR

psydac/fem/splines.py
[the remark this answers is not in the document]
yguclu: Have they really been moved? Can you help me with the commit hash?

psydac/fem/vector.py
yguclu: I suppose that this comment has been removed.

psydac/fem/basic.py
yguclu: Sounds good to me.

yguclu:
My simple assert statements have just revealed that our `VectorFemSpace` does not implement the attribute `periodic` according to the docstring of the base class `FemSpace`. The `periodic` attribute of `FemSpace` reads
```python
    @property
    @abstractmethod
    def periodic( self ):
        """
        Tuple of booleans: along each logical dimension,
        say if domain is periodic.

        """
```
while the `periodic` attribute of `VectorFemSpace` reads
```python
    @property
    def periodic(self):
        return [V.periodic for V in self.spaces]
```
which returns a list of tuples of booleans.

psydac/fem/vector.py
```diff
@@ -356,6 +389,9 @@ def get_refined_space(self, ncells):
         return self._refined_space[tuple(ncells)]
 
     def set_refined_space(self, ncells, new_space):
+        # [YG, 27.03.2025]: It appears that this method is assuming that the
+        # refined space has ldim=2, and that the number of cells is the same
+        # along each axis. These two conditions are very strong.
```
campospinto, line 394: good point 👍

psydac/fem/vector.py
```diff
@@ -21,53 +23,82 @@
...
+        # the property `spaces` which is not part of the abstract interface.
+        self.set_refined_space(self.ncells, self)
         for key in self.spaces[0]._refined_space:
-            if key == tuple(self._ncells):continue
-            self.set_refined_space(key, VectorFemSpace(*[V._refined_space[key] for V in self.spaces]))
+            if key != tuple(self.ncells):
+                V_fine = VectorFemSpace(*[V._refined_space[key] for V in self.spaces])
+                self.set_refined_space(key, V_fine)
+        #+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
```
campospinto, line 101: Thanks! The constructor seems much better now 👍 👍 👍

psydac/fem/tensor.py
```diff
@@ -143,7 +143,10 @@ def ldim( self ):
 
     @property
     def periodic(self):
-        return [V.periodic for V in self.spaces]
+        # [YG, 27.03.2025]: according to the abstract interface of FemSpace,
+        # this property should return a tuple of `ldim` booleans. However, the
+        # spaces in self.spaces seem to be returning a single scalar value.
+        return tuple(V.periodic for V in self.spaces)
```
campospinto, line 149: The purpose of this comment is not clear to me: should we change the periodic function in the self.spaces (presumably univariate Spline spaces ?)

psydac/fem/splines.py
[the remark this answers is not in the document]
campospinto: The push was stuck in limbo... Now it seems to be through.

psydac/fem/tensor.py
yguclu: For me this is hard to say now. Usually the return type is defined by the abstract interface and all subclasses should adhere to it. That said, maybe univariate spaces are "kind of special" in Psydac. I am not even sure that we ever create a `FemField` of a `SplineSpace`...

psydac/api/postprocessing.py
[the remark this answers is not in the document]
yguclu: So have you finally removed `is_product`?

psydac/fem/basic.py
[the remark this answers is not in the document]
yguclu: Removed now?

psydac/fem/basic.py
yguclu: Just double-checking... Have you removed it?

psydac/fem/vector.py
yguclu: This will raise an error immediately if we try to use it in 3D, or with different number of cells in each direction. Better than failing silently or giving an undefined behavior, I guess! 😅

psydac/api/postprocessing.py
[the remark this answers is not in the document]
campospinto: yes!

psydac/fem/basic.py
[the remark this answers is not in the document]
campospinto: yes 👍  For some reason this question was hidden

psydac/fem/basic.py
campospinto: Yes, now we only use the space attribute

psydac/fem/vector.py
campospinto: Absolutely. The comment will be useful to clarify the error when it pops up

psydac/fem/tensor.py
campospinto: You mean that SplineSpace should maybe not be a FemSpace ?

campospinto: (in that case maybe just add a question as a comment on top of the fem/splines.py module ?)

yguclu: Better to create an issue about this, otherwise we have to run the tests once again...

psydac/fem/splines.py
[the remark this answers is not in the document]
yguclu: I see it now, thanks!

yguclu (approved):
Great job @campospinto! It's time to merge this PR

psydac/fem/tests/test_product.py
yguclu: Done!

PR #477, opened 2025-03-28
Rename vector space > coeff space

campospinto:
The spaces used for Fem field coefficients are currently called `vector_space` for no clear reason. 

Since this terminology introduces some confusion when dealing with vector-valued Fem fields, this PR renames these spaces (as attributes of Fem space objects) into `coeff_space`.


yguclu:
These changes have been included in PR #468.

PR #478, opened 2025-04-01
Enable mpi4py 4.0

### LINKED ISSUE #420: Make Psydac compatible with `mpi4py` 4.0.0

spossann:
Closes #420.

Fix usage of function `Translate_ranks` which had non-backward-compatible changes in mpi4py version 4.0: https://mpi4py.readthedocs.io/en/stable/changes.html.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ setuptools >= 61, != 67.2.0
 numpy  >= 1.16
 scipy  >= 1.12
 Cython >= 0.25, < 3.0
-mpi4py >= 3.0.0, < 4
+mpi4py >= 3.0.0
 
 # Required to build h5py from source
 pkgconfig
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ setuptools >= 61, != 67.2.0
 numpy  >= 1.16
 scipy  >= 1.12
 Cython >= 0.25, < 3.0
-mpi4py >= 3.0.0
+mpi4py >= 4
 
 # Required to build h5py from source
 pkgconfig
--- a/psydac/ddm/blocking_data_exchanger.py
+++ b/psydac/ddm/blocking_data_exchanger.py
@@ -146,7 +146,7 @@ class BlockingCartDataExchanger(CartDataExchanger):
             rank_source = info['rank_source']
 
             if self._axis is not None:
-                rank_source = comm.group.Translate_ranks(gcomm.group, np.array([rank_source]), comm.group)[0]
+                rank_source = gcomm.group.Translate_ranks(np.array([rank_source]), comm.group)[0]
             
             recv_buf = (array, 1, recv_typ)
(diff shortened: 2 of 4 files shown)


requirements.txt
```diff
@@ -3,7 +3,7 @@ setuptools >= 61, != 67.2.0
 numpy  >= 1.16
 scipy  >= 1.12
 Cython >= 0.25, < 3.0
-mpi4py >= 3.0.0, < 4
+mpi4py >= 3.0.0
```
yguclu, line 6: I doubt that Psydac could work with mpi4py version 3 after these changes. We should ask for `mpi4py >= 4`.

yguclu (changes requested):
Good job, thanks!

max-models:
Is it really neccesary to install the dependencies with

```
python3 -m pip install -r requirements.txt
python3 -m pip install -r requirements_extra.txt --no-build-isolation --no-cache-dir
```

Can we not simply ensure that h5py is installed from source with `pip install -e . --no-binary h5py`

requirements.txt
max-models: Fixed in https://github.com/pyccel/psydac/pull/478/commits/92cb2067b29b78e0cfd22cb38ebc14c3d5dc4f4f

yguclu:
> Is it really neccesary to install the dependencies with
> 
> ```
> python3 -m pip install -r requirements.txt
> python3 -m pip install -r requirements_extra.txt --no-build-isolation --no-cache-dir
> ```
> 
> Can we not simply ensure that h5py is installed from source with `pip install -e . --no-binary h5py`

This installation procedure was set up a long time ago to circumvent problems between mpi4py, h5py, Cython, and igakit. Since then, all of these packages have substantially improved, so I expect that we could simplify things now. Feel free to create an issue, and a corresponding PR if you can fix it 😃

yguclu (approved):

max-models:
Also, thanks @spossann for doing what I [promised](https://github.com/pyccel/psydac/issues/420#issuecomment-2757175466) to do!

yguclu:
There are two additional files which need updating:
- [psydac/ddm/blocking_data_exchanger.py](https://github.com/pyccel/psydac/blob/b2b38c60abe2fc37b7a41afb8064da4e3086951b/psydac/ddm/blocking_data_exchanger.py#L149)
- [psydac/ddm/nonblocking_data_exchanger.py](https://github.com/pyccel/psydac/blob/b2b38c60abe2fc37b7a41afb8064da4e3086951b/psydac/ddm/nonblocking_data_exchanger.py#L135)

PR #481, opened 2025-04-07
Resolve warnings in sequential unit tests

yguclu:
Changes:

* Use raw strings for docstrings with LaTeX to avoid UTF8 syntax warnings on escape sequences;
* Use CSC matrices to avoid SciPy sparse solver warning.

We do not address the NumPy warnings which arise in our MPI unit tests (see #353).

### DIFF
--- a/psydac/feec/global_projectors.py
+++ b/psydac/feec/global_projectors.py
@@ -356,7 +356,7 @@ class GlobalProjector(metaclass=ABCMeta):
         pass
     
     def __call__(self, fun):
-        """
+        r"""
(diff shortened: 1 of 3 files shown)


campospinto (approved):
Thanks Yaman ! 

Out of curiosity, where do the UTF8 syntax warnings appear ?

yguclu:
> Thanks Yaman !
> 
> Out of curiosity, where do the UTF8 syntax warnings appear ?

They appear when LaTeX is included in the docstrings without properly escaping the backslash character. I am going to explain in the PR description.

PR #482, opened 2025-04-08
Run test with pyccel devel

max-models:
This can be closed. Pyccel 2.0 will be supported when #503 is merged.


yguclu:
See PR #503

PR #483, opened 2025-04-11
Pass `out` parameter to `petsc_to_psydac`

e-moral-sanchez:
Allow to pass an `out` parameter to the function `petsc_to_psydac` to avoid allocating new memory every time we call the function. The `out` parameter can be either `None` or a Psydac vector of type `StencilVector` or `BlockVector`.


psydac/linalg/utilities.py
```diff
@@ -84,6 +84,11 @@ def petsc_to_psydac(x, Xh):
     x : PETSc.Vec
       PETSc vector
 
+    Xh: psydac.linalg.basic.VectorSpace | psydac.linalg.block.BlockVectorSpace
+      Space of the coefficients of the Psydac vector.
```
yguclu, line 88: `VectorSpace` is the base class of `StencilVectorSpace` and `BlockVectorSpace`. I believe that you wanted to write as for the return value `u`:
```suggestion
    Xh: psydac.linalg.stencil.StencilVector | psydac.linalg.block.BlockVectorSpace
      Space of the coefficients of the Psydac vector.
```

e-moral-sanchez: I almost agree. I think it should be `psydac.linalg.stencil.StencilVectorSpace`

yguclu: Just corrected 😉

psydac/linalg/utilities.py
```diff
@@ -84,6 +84,11 @@ def petsc_to_psydac(x, Xh):
     x : PETSc.Vec
       PETSc vector
 
+    Xh: psydac.linalg.stencil.StencilVectorSpace | psydac.linalg.block.BlockVectorSpace
+      Space of the coefficients of the Psydac vector.
+
+    out: Psydac vector to store the result.
```
yguclu, line 90: More complete and precise, and consistent with the rest of the docstring:
```suggestion
    Xh : psydac.linalg.stencil.StencilVectorSpace | psydac.linalg.block.BlockVectorSpace
      Space of the coefficients of the Psydac vector.

    out : psydac.linalg.stencil.StencilVector | psydac.linalg.block.BlockVector, optional
      The Psydac vector where to store the result.
``````

yguclu (approved):
Looks good to me!

PR #484, opened 2025-05-01
Rename `dot` as `inner` in `Vector`, `VectorSpace`, and subclasses

### LINKED ISSUE #330: Change name of `VectorSpace.dot()` into `VectorSpace.inner()`

yguclu:
Main changes
--------------
* Rename the method `dot` of the base class `VectorSpace` as `inner`, and make it an abstract method (which must be implemented by the subclasses).

* Rename the method `dot` of the base class `Vector` as `inner`. This is a concrete method which calls `self.space.inner` and does not need to be overridden by the subclasses.

* Add `inner` methods to the classes `StencilVector`, `BlockVectorSpace`, and `DenseVectorSpace`. These methods override the abstract method of the base class as required, and are derived from the former functions `Stencil.dot`, `BlockVector.dot`, and `DenseVector.dot`, which have been removed (see next point).

* Remove the property `dtype` and the method `dot` (now `inner`) from the classes `StencilVector`, `BlockVector`, and `DenseVector`, because the default methods of the base class already provide a sufficient implementation.

This fixes #330.

Necessary additional changes
------------------------------
* Update all linear solvers in `linalg.solvers` with the new method calls;
* Update all unit tests with the new method calls, in files:
   - `api/tests/test_assembly.py`
   - `feec/tests/test_commuting_projections.py`
   - `feec/tests/test_global_projectors.py`
   - `linalg/tests/test_block.py`
   - `linalg/tests/test_linalg.py`
   - `linalg/tests/test_stencil_interface_matrix.py`
   - `linalg/tests/test_stencil_vector.py`

Unrelated additional changes
-----------------------------
* Rename the class `VectorDot` as `VectorInner` in module `api.ast.linalg`, although never used in Psydac.
* Speed up 3D unit tests in:
   - `feec/tests/test_commuting_projectors.py`
   - `feec/tests/test_global_projectors.py`

### DIFF
--- a/psydac/feec/tests/test_commuting_projections.py
+++ b/psydac/feec/tests/test_commuting_projections.py
@@ -1,4 +1,7 @@
 # -*- coding: UTF-8 -*-
+from mpi4py import MPI
+import numpy as np
+import pytest
 
 from psydac.feec.global_projectors import Projector_H1, Projector_L2, Projector_Hcurl, Projector_Hdiv
 from psydac.fem.tensor       import TensorFemSpace, SplineSpace
@@ -11,18 +14,14 @@ from psydac.ddm.cart         import DomainDecomposition
 from psydac.linalg.solvers   import inverse
 from psydac.linalg.basic     import IdentityOperator
 
-from mpi4py import MPI
-import numpy as np
-import pytest
-
 #==============================================================================
 # 3D tests
 #==============================================================================
-@pytest.mark.parametrize('Nel', [8, 12])
-@pytest.mark.parametrize('Nq', [5])
-@pytest.mark.parametrize('p', [2,3])
+@pytest.mark.parametrize('m', [1, 2])
 @pytest.mark.parametrize('bc', [True, False])
-@pytest.mark.parametrize('m', [1,2])
+@pytest.mark.parametrize('p', [2, 3])
+@pytest.mark.parametrize('Nq', [5])
+@pytest.mark.parametrize('Nel', [5, 6])
 def test_3d_commuting_pro_1(Nel, Nq, p, bc, m):
 
     fun1    = lambda xi1, xi2, xi3 : np.sin(xi1)*np.sin(xi2)*np.sin(xi3)
@@ -92,11 +91,11 @@ def test_3d_commuting_pro_1(Nel, Nq, p, bc, m):
     norm2_e1 = np.sqrt(e1.inner(e1))
     assert norm2_e1 < 1e-12
 
-@pytest.mark.parametrize('Nel', [8, 12])
-@pytest.mark.parametrize('Nq', [8])
-@pytest.mark.parametrize('p', [2,3])
+@pytest.mark.parametrize('m', [1, 2])
 @pytest.mark.parametrize('bc', [True, False])
-@pytest.mark.parametrize('m', [1,2])
+@pytest.mark.parametrize('p', [2, 3])
+@pytest.mark.parametrize('Nq', [7])
+@pytest.mark.parametrize('Nel', [5, 6])
 def test_3d_commuting_pro_2(Nel, Nq, p, bc, m):
 
     fun1    = lambda xi1, xi2, xi3 : np.sin(xi1)*np.sin(xi2)*np.sin(xi3)
@@ -184,11 +183,11 @@ def test_3d_commuting_pro_2(Nel, Nq, p, bc, m):
     norm2_e2 = np.sqrt(e2.inner(e2))
     assert norm2_e2 < 1e-12
 
-@pytest.mark.parametrize('Nel', [8, 12])
-@pytest.mark.parametrize('Nq', [8])
-@pytest.mark.parametrize('p', [2,3])
+@pytest.mark.parametrize('m', [1, 2])
 @pytest.mark.parametrize('bc', [True, False])
-@pytest.mark.parametrize('m', [1,2])
+@pytest.mark.parametrize('p', [2, 3])
+@pytest.mark.parametrize('Nq', [7])
+@pytest.mark.parametrize('Nel', [5, 6])
 def test_3d_commuting_pro_3(Nel, Nq, p, bc, m):
 
     fun1    = lambda xi1, xi2, xi3 : np.sin(xi1)*np.sin(xi2)*np.sin(xi3)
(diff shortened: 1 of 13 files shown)


psydac/linalg/basic.py
```diff
@@ -66,10 +66,35 @@ def zeros(self):
 
         """
 
-#    @abstractmethod
-    def dot(self, a, b):
+    @abstractmethod
+    def inner(self, x, y):
         """
-        Evaluate the scalar product between two vectors of the same space.
+        Evaluate the inner vector product between two vectors of this space V.
```
campospinto, line 72: 'this' space: which one ?

psydac/linalg/block.py
```diff
@@ -85,6 +86,44 @@ def zeros(self):
         """
         return BlockVector(self, [Vi.zeros() for Vi in self._spaces])
 
+    # ...
+    def inner(self, x, y):
+        """
+        Evaluate the inner vector product between two vectors of this space V.
```
campospinto, line 92: same question (which space V?)

Btw, this docstring seems to be a copy of the abstract `inner` method, seems redundant.

psydac/linalg/stencil.py
```diff
@@ -214,6 +214,56 @@ def zeros(self):
...
+            The second vector in the scalar product. The inner product is
+            linear w.r.t. this vector.
+
+        Returns
+        -------
+        float | complex
+            The scalar product of the two vectors. Note that inner(x, x) is
+            a non-negative real number which is zero if and only if x = 0.
+
+        """
```
campospinto, line 247: same question: why copy the docstring from the abstract method ?

psydac/linalg/stencil.py
```diff
@@ -424,46 +474,47 @@ def __del__(self):
...
+        if self.space.parallel:
+            if with_pads:
+                return self._toarray_parallel_with_pads(order=order)
+            else:
+                return self._toarray_parallel_no_pads(order=order)
 
-        inner_func = self._space._inner_func
-        inner_args = (self._data, v._data, *self._space._inner_consts)
+        # In serial case, ignore 'with_pads' flag
+        return self.toarray_local(order=order)
```
campospinto, line 506: this should be specified in the docstring, no ?

psydac/polar/dense.py
```diff
@@ -137,14 +137,66 @@ def zeros(self):
...
+        y : Vector
+            The second vector in the scalar product. The inner product is
+            linear w.r.t. this vector.
+
+        Returns
+        -------
+        float | complex
+            The scalar product of the two vectors. Note that inner(x, x) is
+            a non-negative real number which is zero if and only if x = 0.
+
```
campospinto, line 169: same remark as above

psydac/feec/tests/test_commuting_projections.py
```diff
@@ -11,18 +14,14 @@
...
 #==============================================================================
-@pytest.mark.parametrize('Nel', [8, 12])
-@pytest.mark.parametrize('Nq', [5])
-@pytest.mark.parametrize('p', [2,3])
+@pytest.mark.parametrize('m', [1, 2])
 @pytest.mark.parametrize('bc', [True, False])
-@pytest.mark.parametrize('m', [1,2])
+@pytest.mark.parametrize('p', [2, 3])
+@pytest.mark.parametrize('Nq', [5])
+@pytest.mark.parametrize('Nel', [5, 6])
```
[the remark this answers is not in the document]
yguclu: The order of these lines shows up when you run the tests with the `-v` flag. Since the decorators are applied from the bottom up (the first one is the one closer to the function) the original code was printing the parameters in the wrong order.

psydac/linalg/basic.py
yguclu: The one mentioned in the other docstrings? This is a method of the `VectorSpace` class, so there is just one space.

psydac/linalg/block.py
yguclu: Same answer.

It is indeed a copy of the docstring of the abstract class, which will not show up since we have overridden the method. In order to see any docstring when you write `W.inner?` in IPython, or `help(W.inner)` in Python, the docstring of `BlockVectorSpace.inner` must be given explicitly.

psydac/linalg/stencil.py
yguclu: Same answer... we cannot assume that people use an IDE to read the docstrings.

psydac/linalg/stencil.py
yguclu: Here I have not changed anything from the original code (comments included), just moved things around...

psydac/polar/dense.py
yguclu: Same answer

psydac/linalg/block.py
campospinto: Ok, that does not seem very practical, but if that's the usual way...

psydac/linalg/stencil.py
campospinto: Ok, I'll add the comment in the docstring if you agree

campospinto:
One last comment:  I suggest to replace `dot` by `inner` also in the docstring of `VectorSpace(ABC)` in psydac/linalg/basic.py

psydac/linalg/block.py
yguclu: Indeed this could be error prone, but I have no better solution right now...

yguclu:
> One last comment: I suggest to replace `dot` by `inner` also in the docstring of `VectorSpace(ABC)` in psydac/linalg/basic.py

Sure, I must have missed this!

yguclu:
> > One last comment: I suggest to replace `dot` by `inner` also in the docstring of `VectorSpace(ABC)` in psydac/linalg/basic.py
> 
> Sure, I must have missed this!

Done with 1d2ed29.

campospinto (approved):

PR #488, opened 2025-05-07
Update doctring for inner product in polar/dense.py

campospinto:


psydac/polar/dense.py
```diff
@@ -150,12 +150,16 @@ def inner(self, x, y):
         want to reverse this behavior in order to align with the convention
         of FEniCS.
 
+        TODO [MCP 07.05.2025]: Actually there currently no conjugation yet, 
+        since np.dot is being called.
```
yguclu, line 154:
```suggestion
        TODO [MCP 07.05.2025]: Actually, there is currently no conjugation, 
        since `numpy.dot` is being called.
```

yguclu (approved):
Thanks @campospinto!

PR #499, opened 2025-05-29
Install PETSc-3.23.2 (latest version)

yguclu:
This should avoid an incompatibility between the newly released `Cython-3.1.1` and the `petsc4py` package shipped with the sources of PETSc-3.22.2.


kvrigor (approved):
looks good!

PR #519, opened 2025-08-04
Detach new code generation process from the old one

yguclu:
- Create a new `psydac.api.fem2` module which only contains one new class and a some utility functions. The new module shows all imports clearly.
- Create a new `DiscreteBilinearForm2` class which:
    * Does not inherit from `BasicDiscrete` (and hence `BasicCodeGen`). Accordingly, no PSYDAC abstract syntax tree of class `DefNode` is created by `AST.__init__` (both classes from module `psydac.api.ast.fem`).
    * Does not create a `Parser` object from `psydac.api.ast.parser`, which used to convert a PSYDAC `DefNode` to an old-Pyccel `FunctionDef` from `psydac.pyccel.ast.core`.
    * Does not generate the old-assembly Python code (in the form of a string) using the function `pycode` from `psydac.pyccel.codegen.printing.pycode`.


yguclu:
All [tests](https://github.com/pyccel/psydac/actions/runs/16728590938) have passed. Merging now!

PR #526, opened 2025-09-18
Allow `mpi_dims_mask` with geometry file

yguclu:
Add the optional parameter `mpi_dims_mask` to the constructor of class `Geometry`, as well as its class methods `from_discrete_mapping` and `from_topological_domain`. Add unit tests to verify that the domain is correctly decomposed.

### DIFF
--- a/psydac/cad/tests/test_geometry.py
+++ b/psydac/cad/tests/test_geometry.py
@@ -252,6 +252,7 @@ def test_from_topological_domain():
 
     expected_starts = (0, 2 * rank, 0)
     expected_ends   = (3, 2 * rank + 1, 7)
+
     # Create a topological domain
     F = Mapping('F', dim=3)
     domain = F(Cube(name='Omega'))
@@ -260,8 +261,8 @@ def test_from_topological_domain():
     geo_from_domain = Geometry.from_topological_domain(domain, ncells, comm=comm, mpi_dims_mask=mpi_dims_mask)
 
     # Verify that the domain is distributed as expected
-    assert geometry.ddm.starts == expected_starts
-    assert geometry.ddm.ends   == expected_ends
+    assert geo_from_domain.ddm.starts == expected_starts
+    assert geo_from_domain.ddm.ends   == expected_ends
 
 #==============================================================================
 @pytest.mark.parametrize( 'ncells', [[8,8], [12,12], [14,14]] )


psydac/cad/tests/test_geometry.py
```diff
@@ -169,6 +171,101 @@ def test_geometry_2d_4():
...
+    # Create geometry from topological domain using mpi_dims_mask
+    geo_from_domain = Geometry.from_topological_domain(domain, ncells, comm=comm, mpi_dims_mask=mpi_dims_mask)
+
+    # Verify that the domain is distributed as expected
+    check_decomposition(geo_from_domain, ncells, rank, mpi_dims_mask)
+
+# ==============================================================================
+# Check that each process has 2 cells along the axis where mpi_dims_mask is True
+# Only 1 dimension should be decomposed
+def check_decomposition(geometry, ncells, rank, mpi_dims_mask):
```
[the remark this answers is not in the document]
yguclu: This function verifies that each process owns the correct subset of the whole discretized domain. To do so it checks `starts` and `ends` of the underlying `DomainDecomposition` object stored in the attribute `ddm`. The expected values are indeed hardcoded. Would you like us to add the parameters `expected_starts` and `expected_ends` to this function, and remove `ncells`?

psydac/cad/tests/test_geometry.py
```diff
@@ -169,6 +171,101 @@ def test_geometry_2d_4():
...
+# ==============================================================================
+@pytest.mark.parallel
+def test_from_discrete_mapping():
+
+    comm = MPI.COMM_WORLD
+    rank = comm.rank
+    size = comm.size
+    mpi_dims_mask = [False, False, True]  # We swill verify that this has an effect
+    ncells = [4, 8, 2 * size]  # Each process should have two cells along x3
+    degree = [3, 3, 3]
```
yguclu, line 224: @campospinto We could compute the expected values here:
```suggestion
    degree = [3, 3, 3]

    expected_starts = [0, 0, 2 * rank]
    expected_ends   = [3, 7, 2 * rank + 1]
```

psydac/cad/tests/test_geometry.py
```diff
@@ -169,6 +171,101 @@ def test_geometry_2d_4():
...
+
+    # Create a mapping
+    mapping = discrete_mapping('identity', ncells=ncells, degree=degree)
+
+    # Create geometry from the mapping using mpi_dims_mask
+    geo_from_mapping = Geometry.from_discrete_mapping(mapping, comm=comm, mpi_dims_mask=mpi_dims_mask)
+
+    # Verify that the domain is distributed as expected
+    check_decomposition(geo_from_mapping, ncells, rank, mpi_dims_mask)
```
yguclu, line 233: @campospinto We could use simple assert statements here:

```suggestion
    assert geo_from_mapping.ddm.starts == expected_starts
    assert geo_from_mapping.ddm.ends   == expected_ends
```

PR #527, opened 2025-10-07
Improve `Geometry` class

yguclu:
- Add new factory class method `from_file`
- Use `__init__` method in all factory class methods (`from_file`, `from_discrete_mapping`, and `from_topological_domain`)
- Discourage direct use of `__init__` (factory class methods should be preferred)
- Add mandatory `pdim` parameter (number of dimensions of physical domain) to `__init__`
- Clean up and document (with docstrings) the factory class methods
- Add Pytest mark `@pytest.mark.xdist_group('h5py')` to all tests in `psydac.cad.tests.test_geometry`, because `h5py` is not thread-safe

Moreover:

- Use SymPDE version `0.20.0`
- Always provide interface orientation to the `join` constructor of `Domain` objects (from `sympde.topology.domain`)
- Add interface orientation to multipatch geometry files (in the folder `psydac/cad/mesh/multipatch`)
- Use `ubuntu_installations` action in `documentation` workflow
- Fix module name in documentation sources: `cmd.mesh` >> `cmd.psydac_mesh`

TO DO:

- [ ] In `from_file`, make sure that the `domain` argument (an instance of `sympde.topology.Domain`) is compatible with the topological domain defined in the `.yml` section of the HDF5 file.
- [ ] Update unit tests in `cad.tests.test_geometry`:
   - [ ] Add some 1D unit tests
   - [ ] Add proper asserts statements to `test_geometry_2d_1` and `test_geometry_2d_2`
   - [ ] Remove obsolete test function `test_geometry_1`
   - [ ] Test functions from `cad.cad` in new module `cad.tests.test_cad`
- [ ] After release of SymPDE version 0.20.0, update `pyproject.toml` and run all unit tests once again on all platforms
- [ ] Mention SymPDE 0.20.0 in PR description
- [ ] Mention SymPDE 0.20.0 in CHANGELOG.md


psydac/cad/geometry.py
```diff
@@ -8,60 +8,68 @@
...
     Distributed discrete geometry that works for single and multiple patches.
-    The Geometry object can be created in two ways:
-    - case 1 : through a geometry file whos name can be given to the constructor
-    - case 2 : provide the ncells, the periodicity and the mapping objects of each patch.
+
+    The Geometry object can be created in four ways:
+    - case 0 : providing a `Domain` to `__init__` with detailed parameters for each patch.
+    - case 1 : passing the path to a geometry file to `from_file`.
+    - case 2 : passing a `SplineMapping` to `from_discrete_mapping` (single patch).
+    - case 3 : passing a `Domain`, ncells, and periodicity to `from_topological_domain` (single or multi-patch).
```
campospinto, line 51: It's very nice to list these cases!
Maybe we should also specify here which cases should be suggested to users ?
My feeling is that case 0 should only be used "internally", i.e. called by  existing functions

edit 1: ok I see now that this is also explained in the PR description. then I suggest to specify it in this docstring as well.

edit2: it would be useful to indicate also here that case 1 and 3 are called by the `discretize_domain` interface (or just `discretize` called on a SymPDE Domain)

psydac/cad/geometry.py
campospinto, line 261: here I suggest checking that domain is of correct type (a symPDE `Domain` I guess)

psydac/cad/geometry.py
campospinto: In particular, I would suggest to remove or rewrite the direct constructions of `Geometry` objects in `cad/cad.py`, to avoid giving bad ideas

psydac/api/discretization.py
```diff
@@ -578,7 +578,7 @@ def discretize_domain(domain, *, filename=None, ncells=None, periodic=None, comm
         raise ValueError("Cannot provide both 'filename' and 'ncells'")
 
     elif filename:
-        return Geometry(filename=filename, comm=comm, mpi_dims_mask=mpi_dims_mask)
+        return Geometry.from_file(filename, comm=comm, mpi_dims_mask=mpi_dims_mask)
```
campospinto, line 582: thanks!

psydac/cad/geometry.py
yguclu: Done in 349eee8c

psydac/cad/geometry.py
yguclu: Do you mean the script-like code on the bottom of `cad/cad.py`? I believe that code does not work right now, so we could either remove it or try to create a unit test out of it

campospinto: yes, I guess I was referring to these calls to the Geometry constructor. Since these are not advised the best is probably to remove them

yguclu: Excluding some unit tests (which naturally have to test also the `__init__` function), I can find those calls only at the bottom of `cad/cad.py`. That is why I was asking

PR #529, opened 2025-10-14
Fix method `plot_2d_decomposition` of `TensorFemSpace`

yguclu:
Fix bug in method `plot_2d_decomposition` of `TensorFemSpace`, which was failing when run in parallel with a distributed spline mapping.

- Create a new test file `psydac/fem/tests/test_tensor.py` with a unit test which fails on the `devel` branch. This compares the generated PNG images with "reference" ones which are known to be correct, within a 2 % relative tolerance on each of the RGB channels.
- Only evaluate mapping in local subdomain owned by process
- Gather global mapping information on root process
- Add optional parameters `fig`, `ax`, and `mpi_root`
- Add docstring
- Update `examples/poisson_2d_mapping.py` to pass the correct mapping (i.e. also a distributed spline mapping if that is used in the computations) to `plot_2d_decomposition`.

## Note

The function `plot_2d_decomposition` does not have to be a method of `TensorFemSpace`. See issue #532.

## TO DO

- [x] Test optional parameters
- [x] Add assert statements to unit test
- [x] Reactivate all unit tests

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -192,52 +192,50 @@ jobs:
           mkdir pytest
           cp mpi_tester.py pytest
 
-#      - name: Run coverage tests on macOS
-#        if: matrix.os == 'macos-14'
-#        working-directory: ./pytest
-#        run: >- 
-#          python -m pytest -n auto
-#          --cov psydac
-#          --cov-config $GITHUB_WORKSPACE/pyproject.toml
-#          --cov-report xml
-#          --pyargs psydac -m "not parallel and not petsc"
-#
-#      - name: Run single-process tests with Pytest on Ubuntu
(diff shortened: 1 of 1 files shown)


campospinto (commented):
Thanks Yaman! The new function plot_2d_decomposition looks good to me and I checked that it works well, both for discrete and analytical mappings (which could also be tested now by passing a different mapping to the function, with the same tensorspace V). 

I also find that the function should be a member of the (new) Geometry class.

campospinto (approved):
Thanks for adding the analytical mapping. It looks good now and can be improved after merging the improved Geometry class (#527 I guess?)

(I'm approving the PR but of course the todo's above should be done first)

alisa-kirkinskaia (approved):

Issue #545, opened 2025-12-02
Make `DiscreteBilinearForm` in `psydac.api.fem_bilinear_form` work on 3D multi-patch domains


yguclu:
The sum factorization algorithm introduced in PR #448 only works for the assembly of bilinear forms in 3D single-patch domains. When discretizing over a 3D multi-patch domain, the constructor of the new `DiscreteBilinearForm` fails in multiple places. The constructor logic is designed to handle multi-patch domains, but it contains some bugs which should be fixed. This problem has not been caught so far because of limitations in our 3D unit tests.

PR #546, opened 2025-12-02
Fix some bugs in multi-patch domains

### LINKED ISSUE #551: raise a warning when discretizing a multipatch domain with periodic argument
When discretizing a multipatch domain, passing a `periodic` argument does not make much sense in general -- and when it does, care should be taken since it is applied on (all) the logical patches. We should probably raise a warning when a user tries to do so.

### LINKED ISSUE #552: discretizing a ScalarFunctionSpace on a disconnected multipatch domain returns a VectorFemSpace
Calling the `discretize` function on a sympde `ScalarFunctionSpace` defined on a connected multipatch domain correctly returns a `MultipatchFemSpace`. On a disconnected domain it incorrectly returns a `VectorFemSpace`. This should be fixed.

Example: 

```
import numpy as np

from sympde import Domain
from sympde.topology import Square, IdentityMapping, ScalarFunctionSpace, VectorFunctionSpace
from sympde.utilities.utils import plot_domain

def build_space(connect=True):
    F = IdentityMapping('I', dim=2)

    om1 = F(Square('1', bounds1=(0., 1.), bounds2=(0., 1.)))
    om2 = F(Square('2', bounds1=(0., 1.), bounds2=(1., 2.)))

    patches = [om1, om2]
    if connect:
        connectivity = [((0,1,-1),(1,1,-1), 1)]
    else:
        connectivity = []    
    Omega = Domain.join(patches, connectivity, 'domain')
    # plot_domain(Omega, isolines=True)

    V = ScalarFunctionSpace('V', Omega, kind=None)
    Omega_h = discretize(Omega, ncells=[10, 10])
    Vh = discretize(V, Omega_h, degree=[2, 2])
    print(type(Vh))
```
then, 
`build_space(True)`
returns `<class 'psydac.fem.vector.MultipatchFemSpace'>`
but 
`build_space(False)`
returns `<class 'psydac.fem.vector.VectorFemSpace'>`

yguclu:
- Do not remove `sum_factorization` argument from auxiliary equation for L2 boundary projection
- Do not use sum factorization for assembly over 3D multi-patch domains (see issue #545)
- Make 2D plots work for multi-patch domains
- Set correct multipatch FEM space type on disconnected multipatch domains (fix #552)
- Raise a warning when discretizing a multipatch domain with a periodic flag (fix #551)

### DIFF
--- a/psydac/cad/geometry.py
+++ b/psydac/cad/geometry.py
@@ -186,7 +186,9 @@ class Geometry:
         else:
             if len(interior) > 1 and True in periodic:
                 msg = "Discretizing a multipatch domain with a periodic flag is not advised -- continue at your own risk."
-                warnings.warn(msg, Warning)
+                # [MCP 18.12.2025] the following line may be causing a strange error in the CI (MPI tests for macos-14/Python 3.10)
+                # warnings.warn(msg, Warning)  
+                warnings.warn(msg, UserWarning)
 
 
         if isinstance(periodic, (list, tuple)):


psydac/fem/plotting_utilities.py
```diff
@@ -235,8 +241,8 @@ def get_patch_knots_gridlines(Vh, N, mappings, plotted_patch=-1):
     F = [M.get_callable_mapping() for d, M in mappings.items()]
 
     if plotted_patch in range(len(mappings)):
-        grid_x1 = Vh.patch_spaces[plotted_patch].spaces[0].breaks[0]
-        grid_x2 = Vh.patch_spaces[plotted_patch].spaces[0].breaks[1]
+        grid_x1 = Vh.patch_spaces[plotted_patch].spaces[0].breaks
+        grid_x2 = Vh.patch_spaces[plotted_patch].spaces[0].breaks
```
campospinto: why define both grids as the same one?

yguclu:
@campospinto Thanks for fixing #552! Your one-liner change looks good to me.

As soon as you approve this PR, I think we can merge it

PR #558, opened 2025-12-14
Fix sum factorization

yguclu:
Avoid failures in the new 3D assembly algorithm based on sum factorization:

* Do not use it on boundaries and interfaces
* Initialize `self._func` to `None`
* Handle complex quantities

### DIFF
--- a/psydac/api/discretization.py
+++ b/psydac/api/discretization.py
@@ -592,8 +592,8 @@ def discretize(a, *args, **kwargs):
         domain  = domain_h.domain
         dim     = domain.dim
 
-        # The current implementation of the sum factorization algorithm does not support openMP parallelization
-        # It is not properly tested, whether this way of excluding openMP parallelized code from using the sum factorization algorithm works.
+        # The current implementation of the sum factorization algorithm does not support OpenMP parallelization
+        # It is not properly tested, whether this way of excluding OpenMP parallelized code from using the sum factorization algorithm works.
         backend = kwargs.get('backend')# or None
         assembly_backend = kwargs.get('assembly_backend')# or None
         assembly_backend = backend or assembly_backend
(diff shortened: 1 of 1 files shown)


yguclu:
> Hey @yguclu, the implemented changes look good to me, I only have two comments:
> 
> 1. In `api/discretization.py::discretize`, there's this piece of code:
> 
>    ```python
> 
>       if isinstance(a, sym_BilinearForm):
> 
>           if kwargs.pop('sum_factorization'):
> 
>               return DiscreteBilinearForm_SF(a, kernel_expr, *args, **kwargs)
> 
>           else:
> 
>               return DiscreteBilinearForm(a, kernel_expr, *args, **kwargs)
> 
>    ```
> 
>    Don't you think we could, just like you've done in `api/fem_sum_form.py`, already check here whether `a` is an interior expression? Somethink along the lines of
> 
>    ```python
> 
>       if isinstance(a, sym_BilinearForm):
> 
>           # Currently sum factorization can only be used for interior domains
> 
>           from sympde.expr.evaluation import DomainExpression
> 
>           is_interior_expr = isinstance(e, DomainExpression)
> 
>           if kwargs.pop('sum_factorization') and is_interior_expr:
> 
>               return DiscreteBilinearForm_SF(a, kernel_expr, *args, **kwargs)
> 
>           else:
> 
>               return DiscreteBilinearForm(a, kernel_expr, *args, **kwargs)
> 
>    ```

This suggestion makes sense to me, but it should be tested! Could you please push a commit so we can check if it works?

> 2. In `api/fem_bilinear_form.py`, parts of the code are apparently not tested. Between line 534 and line 655 we want to use `StencilInterfaceMatrix`, `knot_insertion_projection_operator` and `ComposedLinearOperator` multiple times, but these two classes and this one function are never imported.

Good catch! What do you propose exactly?

PR #565, opened 2026-01-22
Expand editable install info in README.md

yguclu:
**Relevant changes**
- Explain that `devel` branch is cloned by default
- Explain that `build/` folder should be removed before attempting another editable install

**Additional changes**
- Run documentation workflow whenever `README.md` is modified
- Run testing workflow on PRs only when set to "ready for review"
- Add new `Unreleased` section to `CHANGELOG.md`


campospinto (approved):
This is very useful, thanks!

PR #567, opened 2026-01-27
Improve `psydac test` command

yguclu:
### Main changes

- Allow selecting a specific test function in a module by passing `--mod test_<MODULE>::test_<FUNCTION>`
- Pass `--dist loadgroup` to group test execution by the `xdist_group` mark (see the documentation of [pytest-xdist](https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus)), which allows running non-thread-safe tests sequentially on a single runner
- Remove local `.pytest_cache` folder before running tests
- Move `pytest.ini` to package root, so it can be used after installation
- If not present, copy `pytest.ini` to local working directory before running tests

### Additional changes

- Fix the parallel creation of the folder `__psydac__` in `psydac.api.fem_bilinear_form`: only root process creates it


campospinto (approved):
I tried different testing commands, the changes look good to me!

PR #569, opened 2026-01-29
Fix CPU flags in `psydac/api/settings.py`

### LINKED ISSUE #568: wrong cpu flag for Apple M4 chips
The current value of `cpu_flag` in `psydac/api/settings.py` is set to `apple-m4` on Apple M4 computers, which is not supported by current compilers such as GCC 15.2.0.

campospinto:
Clean up platform-specific flags for GCC compilers. In particular, do not use the plausible but inexistent flag `-mcpu=apple-m4` on Apple M4 computers (which use ARM chips). Instead, always use `-march=native`, and additionally `-mavx` on `x86_64` architectures. Further, do not use `-mtune=native` as this is not always supported. These changes fix #568


psydac/api/settings.py
```diff
@@ -61,6 +61,7 @@
     if cpu_brand.startswith("Apple M"):
         # Example: "Apple M3 Pro (virtual)" --> " -mcpu=apple-m3"
         cpu_flag = '-'.join(cpu_brand.lower().split()[:2])
+        cpu_flag = cpu_flag.replace('4', '3')
```
max-models: But does this actually work? 😅 Looks like we are just pretending to be on M3 while actually being on M4.

campospinto: It worked on my Macbook with an M4 chip. Maybe not the optimal solution, but already better than without this hack.

max-models: Ok! It will also convert `m14` to `m13`, but by then I hope there will be a better solution! :) 

I'll approve it, but maybe we should rethink the strategy a bit in general.

max-models: Actually, I am a bit confused about this line, it seems like I tested it on a M4 mac but definitely not with GCC  15 https://github.com/pyccel/psydac/pull/480

I don't really understand the comment "Apple silicon requires architecture-specific flags (see https://github.com/pyccel/psydac/pull/411) which are only available on GCC version >= 14". What would happen if you just use `-march=native`? Could you try this on your Mac?

```python
if platform.system() == "Darwin" and platform.machine() == 'arm64':
    PSYDAC_BACKEND_GPYCCEL['flags'] += ' -march=native'
else:
    # Default architecture flags
    PSYDAC_BACKEND_GPYCCEL['flags'] += ' -march=native -mtune=native'
    if platform.machine() == 'x86_64':
        PSYDAC_BACKEND_GPYCCEL['flags'] += ' -mavx'
```

campospinto: I just tried it on two Macbooks (with M1 and M4 chips), it works fine. @yguclu what do you think ?

campospinto: @max-models I'm committing your suggestion to see if the CI tests still pass

yguclu: I have looked into this a bit more. In response to @max-models's comment, I think that `-mcpu=CPU` is appropriate when one is cross-compiling, e.g. to produce wheels for different target architectures. This is not what we are doing here, of course.

Passing `-march=native` tells the GCC compiler to use the appropriate instruction set for the host CPU, which sounds like what we should do. In addition to that, I would also pass `-mtune=native`, as in the code under "Default architecture flags". Hence, probably, we could simplify the code greatly.

psydac/api/settings.py
yguclu: Does this work?
```suggestion
                       'flags'   : '-O3 -ffast-math -march=native -mtune=native',
```

yguclu, line 27: I am not sure if we need to check the compiler version at all. But here we could simply add the following lines:

```suggestion

if platform.machine() == 'x86_64':
    PSYDAC_BACKEND_GPYCCEL['flags'] += ' -mavx'

```

campospinto: I just tried: on the M3 and M4 macs, with
`   'flags'   : '-O3 -ffast-math -march=native -mtune=native',`
some tests don't pass. 
The error message is the same: 

```
cmd = ['/opt/homebrew/bin/gfortran', '-O3', '-ffast-math', '-march=native', '-mtune=native', '-O3', ...], verbose = False

    @staticmethod
    def run_command(cmd, verbose):
        """
        Run the provided command and collect the output.
    
        Run the provided compilation command, collect the output and raise any
        necessary errors if the file does not compile.
    
        Parameters
        ----------
        cmd : list of str
            The command to run.
        verbose : int
            Indicates the level of verbosity.
    
        Returns
        -------
        str
            The exact command that was run.
    
        Raises
        ------
        RuntimeError
            Raises `RuntimeError` if the file does not compile.
        """
        cmd = [os.path.expandvars(c) for c in cmd]
        if verbose > 1:
            print(' '.join(cmd))
    
        with subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE,
                universal_newlines=True) as p:
            out, err = p.communicate()
    
        if verbose and out:
            print(out)
        if p.returncode != 0:
            err_msg = "Failed to build module"
            err_msg += "\n" + err
>           raise RuntimeError(err_msg)
E           RuntimeError: Failed to build module
E           f951: Error: unknown value 'apple-m1' for '-march'
E           f951: note: valid arguments are: armv8-a armv8.1-a armv8.2-a armv8.3-a armv8.4-a armv8.5-a armv8.6-a armv8.7-a armv8.8-a armv8.9-a armv8-r armv9-a armv9.1-a armv9.2-a armv9.3-a armv9.4-a armv9.5-a native
E           f951: note: did you mean '-mcpu=apple-m1'?
```

campospinto: out of curiosity I also tried with 
`   'flags'   : '-O3 -ffast-math -march=native',`
and then the tests pass (both on the M3 and M4 chips)

yguclu: So the flag which is not working is `-mtune=native`?

campospinto: yes, it seems so. (in these tests I disabled the Platform-dependent flags)

psydac/api/settings.py
campospinto: [here](https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html) I see that "Specifying -march=cpu-type implies -mtune=cpu-type, except where noted otherwise."

but I don't know if this is specific to x86 architectures

yguclu: There is a layer of mystery in the GCC documentation, but I would expect this to be true on all architectures 😅

psydac/api/settings.py
```diff
@@ -53,26 +61,31 @@
 gfortran_version = Version(gfortran_version_string)
```
yguclu: These lines are not used any longer, so I suggest to comment them out. Then we will not need to import `subprocess`, `re`, and `packaging.version.Version`.

Moreover, with the next version of Pyccel we will be able to use the C code as the backend language, so it would not make sense to only check the `gfortran` version

psydac/api/settings.py
```diff
@@ -53,26 +61,31 @@
...
-
-else:
-    # Default architecture flags
-    PSYDAC_BACKEND_GPYCCEL['flags'] += ' -march=native -mtune=native'
-    if platform.machine() == 'x86_64':
-        PSYDAC_BACKEND_GPYCCEL['flags'] += ' -mavx'
+# if platform.system() == "Darwin" and platform.machine() == 'arm64':
+#     PSYDAC_BACKEND_GPYCCEL['flags'] += ' -march=native'
+
+# [MCP] version before PR #569, delete if OK:
```
yguclu: We could remove these four new lines, and just keep the commented-out code below

yguclu (changes requested):
Thanks again for this PR!

This looks good to me now, I just have two minor comments

psydac/api/settings.py
campospinto: (was committed by hand)

psydac/api/settings.py
campospinto: (done)

psydac/api/settings.py
campospinto: (done)

psydac/api/settings.py
```diff
@@ -20,11 +20,18 @@
...
-                       'flags'   : '-O3 -ffast-math',
+                       'flags'   : '-O3 -ffast-math -march=native',
                        'folder'  : '__gpyccel__',
                        'tag'     : 'gpyccel',
                        'openmp'  : False}
 
+# [MCP 10.02.2026] this flag was used previously for intel mac but probably not needed
+# if not platform.system() == "Darwin":
+#     PSYDAC_BACKEND_GPYCCEL['flags'] += ' -mtune=native'
+
```
yguclu: I do not understand this comment. These four lines have been added in this PR and then commented out, hence I recommend removing them

psydac/api/settings.py
```diff
@@ -47,32 +54,32 @@
...
+# if platform.system() == "Darwin" and platform.machine() == 'arm64' and gfortran_version >= Version("14"):
+#
+#     # Apple silicon requires architecture-specific flags (see https://github.com/pyccel/psydac/pull/411)
+#     # which are only available on GCC version >= 14
+#     cpu_brand = subprocess.check_output(['sysctl','-n','machdep.cpu.brand_string']).decode('utf-8').strip() # nosec B603, B607
+#     if cpu_brand.startswith("Apple M"):
+#         # Example: "Apple M3 Pro (virtual)" --> " -mcpu=apple-m3"
+#         cpu_flag = '-'.join(cpu_brand.lower().split()[:2])
+#         cpu_flag = cpu_flag.replace('4', '3')
```
yguclu: This was the first hack I suggested, but it is not used now after commenting the whole code block. Hence I recommend removing this line

yguclu (changes requested):

psydac/api/settings.py
campospinto: ok -- I'm removing only the line then

psydac/api/settings.py
campospinto: indeed, removed

yguclu (approved):

PR #571, opened 2026-02-03
Reactivate sum factorization algorithm

jowezarek:
Fix an issue that prevented the use of the sum factorization for matrix assembly: in the function `discretize`, the variable `terminal_expr` is not a scalar but rather a tuple with a single element.

**Detailed list of changes**

- Check whether `terminal_expr[0]` is a `DomainExpression`, not `terminal_expr`
- In the folder `examples/performance/`, rename the file `README.md` to `matrix_assembly_speed_log.md`
- Clarify purpose and usage of the aforementioned markdown file

### DIFF
--- a/examples/performance/README.md
+++ b/examples/performance/README.md
@@ -67,7 +67,7 @@ Test cases
 Data
 ----
 
-2025-09-09 17:28:19 (added by Julian O. - ThinkPad T14 on performance mode)
+2025-09-09 17:28:19 (added by Julian O. - ThinkPad T14 in performance mode)
 ----------
 
 | Test case | old assembly | new assembly | old discretization | new discretization |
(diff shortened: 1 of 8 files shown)


examples/performance/README.md
```diff
@@ -84,3 +84,37 @@ Data
...
+| 2.3 | 43.744 | 3.914 | 21.244 | 23.042 |
+| 3.1.1 | 0.242 | 0.043 | 1.713 | 1.798 |
+| 3.1.2 | 0.561 | 0.045 | 1.701 | 1.522 |
+| 3.1.3 | 0.897 | 0.161 | 4.085 | 3.296 |
+| 3.2 | 1.043 | 0.046 | 1.364 | 1.691 |
+| 3.3 | 0.57 | 0.082 | 1.045 | 2.024 |
+| 3.4 | 1.656 | 0.073 | 1.775 | 2.204 |
+| 3.5 | 0.563 | 0.074 | 1.788 | 2.274 |
+| 3.6 | 0.768 | 0.211 | 3.307 | 1.865 |
+
```
yguclu: I am not sure that we should report all of these numbers. In particular, the timings seem to oscillate quite a bit. Maybe one should run these benchmarks a few times and report the average value?

psydac/api/discretization.py
```diff
@@ -659,7 +659,10 @@ def discretize(a, *args, **kwargs):
     if isinstance(a, sym_BilinearForm):
         # Currently sum factorization can only be used for interior domains
         from sympde.expr.evaluation import DomainExpression
-        is_interior_expr = isinstance(kernel_expr, DomainExpression)
+        if isinstance(kernel_expr, tuple):
+            is_interior_expr = isinstance(kernel_expr[0], DomainExpression)
```
yguclu: In practice, is `kernel_expr` anything else than a tuple? Could it be a list?

yguclu (changes requested):
Good catch for finding this bug!

I have a couple of comments. Also, please update the file `CHANGELOG.md`.

examples/performance/README.md
jowezarek: I would argue that it makes sense the way it is.
The purpose of these timings is not to compare 25.953s (measurement day 1) with 30.908s (measurement day 2) and conclude that the code got slower. This would have been useful a year ago when we were still deciding on how to implement the sum-factorization algorithm best.
Now, the purpose is rather to check that for all the test cases the new assembly algorithm remains significantly faster then the old algorithm. Also, in principle anyone on any machine can contribute timings, and in that case comparing timings (obtained from different devices) makes no sense.

However, I would agree that these timings need not be visible in a README.md file, but could be more hidden in a matrix_assembly_speed_log.md file.

psydac/api/discretization.py
jowezarek: I don't know. I just run these tests and checked whether `kernel_expr` was always a `tuple`:
- `test_api_1d_compatible_spaces.py`
- `test_2d_complex.py`
- `test_equation.py`
- `test_api_feec_3d.py`
- `test_api_feec_2d.py`
- `test_api_feec_1d.py`
- `test_sum_factorization_assembly_3d.py` (three times, with no mapping, a B-spline mapping, and an analytical mapping)

and in all cases `kernel_expr` was always a `tuple`. Do you suggest I change the above code simply to
```python
is_interior_expr = isinstance(kernel_expr[0], DomainExpression)
```
?

examples/performance/README.md
yguclu: All right, I see your point, and I agree with changing the name of this file.

Please expand the explanation at the beginning to make it clear that we only care about comparing the new and old assembly algorithms.

I am still not convinced that we need to keep track of the timings at different dates, though. This log file is definitely useful for cathing bugs, but I would not add a new table in this PR unless there is a need for it

psydac/api/discretization.py
yguclu: It is probably safe to assume that `kernel_expr` is always a `tuple`, but it might contain more than one object on which to assemble. E.g., there could be an additional domain boundary, or multiple patches in a multipatch domain

examples/performance/README.md
jowezarek: Alright, agreed!

psydac/api/discretization.py
jowezarek: I just did some more tests: At this point in the code the length of `kernel_expr` can only be 1, so I propose to indeed write
```python
is_interior_expr = isinstance(kernel_expr[0], DomainExpression)
```
The case where `kernel_expr` has a length greater than 1 is handled by the class `DiscreteSumForm`, which itself correctly applies the sum factorization algorithm by checking for `dim == 3` and `is_interior_expr`.

jowezarek: 7c67b25f39c93143142659f5dc9be6a76c0a6c5d

examples/performance/README.md
jowezarek: 7c67b25f39c93143142659f5dc9be6a76c0a6c5d

yguclu (approved):
Looks good to me!

Please update the CHANGELOG now. Also, the PR description can be shortened, so it can be used for the PR commit message

PR #572, opened 2026-02-10
Fix a bug in BlockLinearOperator

alisa-kirkinskaia:
- Fix a bug in `_dot()` of `BlockLinearOperator`. The bug appears when calling the method multiple times due to stale data from previous blocks. The fix clears the output buffer before calling `dot` of the current block.
- Add a unit test which constructs block operators that only partially overwrite the output buffer.

### DIFF
--- a/psydac/linalg/block.py
+++ b/psydac/linalg/block.py
@@ -694,12 +694,15 @@ class BlockLinearOperator(LinearOperator):
 
         if n_rows == 1:
             for (_, j), L0j in blocks.items():
+                inc *= 0.0
                 out += L0j.dot(v[j], out=inc)
         elif n_cols == 1:
             for (i, _), Li0 in blocks.items():
+                inc[i] *= 0.0
                 out[i] += Li0.dot(v, out=inc[i])
         else:
             for (i, j), Lij in blocks.items():
+                inc[i] *= 0.0
                 out[i] += Lij.dot(v[j], out=inc[i])
 
     # ...
--- a/psydac/linalg/tests/test_block.py
+++ b/psydac/linalg/tests/test_block.py
@@ -7,6 +7,7 @@ import pytest
 import numpy as np
 from scipy.sparse import csr_matrix
 
+from psydac.linalg.basic import LinearOperator
 from psydac.linalg.direct_solvers import SparseSolver
 from psydac.linalg.stencil        import StencilVectorSpace, StencilVector, StencilMatrix
 from psydac.linalg.block          import BlockVectorSpace, BlockVector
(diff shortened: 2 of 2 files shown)


psydac/linalg/block.py
```diff
@@ -694,12 +694,15 @@ def _dot(blocks, v, out, n_rows, n_cols, inc):
...
+                inc *= 0.0
                 out += L0j.dot(v[j], out=inc)
         elif n_cols == 1:
             for (i, _), Li0 in blocks.items():
+                inc[i] *= 0.0
                 out[i] += Li0.dot(v, out=inc[i])
         else:
             for (i, j), Lij in blocks.items():
+                inc[i] *= 0.0
                 out[i] += Lij.dot(v[j], out=inc[i])
```
yguclu, line 706: Looking at this code, I must admit that I cannot really understand where the bug is, and why these changes should fix it 😅 

The `dot` method of any `LinearOperator` object may accept the optional argument `out`, but the method must overwrite `out` with the result regardless of the data it contains. Hence there should be no reason to set `inc` to zero here.

Every subclass of `LinearOperator` should behave according to this rule, please see the docstring of the abstract method [LinearOperator.dot](https://github.com/pyccel/psydac/blob/e66b28e28b82baf9064b034fe5199d2fb94db120/psydac/linalg/basic.py#L292-L315). Maybe this is the real problem?

psydac/linalg/tests/test_block.py
```diff
@@ -733,6 +734,93 @@ def test_block_linear_operator_serial_dot( dtype, n1, n2, p1, p2, P1, P2  ):
...
+@pytest.mark.parametrize( 'npts', [[6, 8], [10, 5]] )
+@pytest.mark.parametrize( 'p', [[1,1], [2,3]] )
+@pytest.mark.parametrize( 'P1', [True, False] )
+@pytest.mark.parametrize( 'P2', [True] )
+
+def test_block_linear_operator_partial_write(dtype, npts, p, P1, P2):
+    """
+    We test that dot method of BlockLinearOperator produces correct results when
+    dot methods of blocks (linear operators) only partially overwrite the output.
+    """
```
yguclu, line 749: As I explained in my comment to the changes in the module `psydac.linalg.block`, the `dot` method of any `LinearOperator` object must completely overwrite the `out` vector if this is provided. In practice we should set `out` to zero before doing anything else

psydac/linalg/tests/test_block.py
```diff
@@ -733,6 +734,93 @@ def test_block_linear_operator_serial_dot( dtype, n1, n2, p1, p2, P1, P2  ):
...
+            return self._codomain
+
+        @property
+        def dtype(self):
+            return self._domain.dtype
+
+        def dot(self, x, out=None):
+            assert isinstance(x, StencilVector)
+            if out is None:
+                out = self.codomain.zeros()
```
yguclu, line 771: This is needed here:
```suggestion
            if out is None:
                out = self.codomain.zeros()
            else:
                assert isinstance(out, StencilVector)
                assert out.space is self.codomain
                out *= 0.0
```

yguclu (commented):
I am not sure that there was really a bug in `BlockLinearOperator`, but we definitely had one [here](https://github.com/pyccel/psydac/blob/polar_splines_alisa/psydac/feec/polar/conga_projections.py) when `out` was given to the linear operators (as it happens in a `BlockLinearOperator`)

PR #573, opened 2026-02-11
Update CI installation of `h5py` and `petsc4py` after release of `setuptools` 81.0

campospinto:
On 06 Feb 2026 [`setuptools v81.0.0`](https://setuptools.pypa.io/en/stable/history.html#v81-0-0) was released with breaking changes, i.e. the support for the parameter `--dry-run` was removed from `setup.py`. As a consequence, our installed versions of `h5py` and `petsc4py` have stopped working. As a temporary fix for our CI workflows, we install the latest package versions from their respective repositories, where the issue has been fixed.

**Detailed list of changes**

- Download `h5py` from GitHub (`master` branch) and build it from sources, instead of installing the latest release from PyPI.
- Change the PETSc `git clone` command to download the 'release' branch instead of a broken version. Hence we install the latest package from the official GitLab repository. (When a new version number is available, we can point the `git clone` command to it.)
- Trigger documentation workflow when PR is set to "Ready for review".

### DIFF
--- a/.github/actions/parallel_h5py/action.yml
+++ b/.github/actions/parallel_h5py/action.yml
@@ -20,9 +20,18 @@ runs:
       run: |
         export CC="mpicc"
         export HDF5_MPI="ON"
-        pip install 'setuptools<81.0'
-        pip install h5py --no-cache-dir --no-binary h5py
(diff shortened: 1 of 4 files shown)


yguclu (approved):
LGTM

PR #579, opened 2026-03-22
Apply some important fixes

yguclu:
- Use `pyccel >= 2.2.3`: this allows compiling all kernels with C
- Require `h5py >= 3.16` to allow for standard install with `setuptools >= 81.0`
- Require `numpy >= 2.1` to support Python >= 3.10
- Use `pytest.toml` instead of `pytest.ini` for Pytest configuration (requires `pytest >= 9.0`)
- Move coverage configuration from `pyproject.toml` to `psydac/pytest.toml`
- Don't run postprocessing unit tests with `pytest-xdist` because `h5py` is not thread-safe: add Pytest mark `@pytest.mark.xdist_group('h5py')` to all unit tests in `psydac/api/tests/test_postprocessing.py`
- Return error code on failure of the `psydac test` command; this was not correctly returned before, hence the CI could not detect when the unit tests were failing in a subprocess
- Return error code on failure of the `psydac compile` command; this fix does not directly affect the CI, but it is included for correctness.
- Check correct reporting of failure for `psydac test` command in CI testing
- Use correct configuration file in coverage CI tests
- For clarity, rename temporary folder where all CI tests are run: `pytest` -> `scratch`
- Clean up `.gitignore`; specifically, stop ignoring files which don't exist anymore


kvrigor (approved):

CHANGELOG.md
```diff
@@ -10,15 +10,24 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+-   #579 : Require `h5py>=3.16` which installs correctly with `setuptools>=81.0`
+-   #579 : Don't run postprocessing unit tests with `pytest-xdist` because `h5py` is not thread-safe
```
campospinto, line 14: out of curiosity, where is this enforced ? in the pytest.toml file ?

campospinto (approved):
This PR has a lot of improvements, thanks ! I can only do a shallow review but it seems good to me. I just have one question mostly to clarify a point.

CHANGELOG.md
yguclu: No, it is enforced by adding `@pytest.mark.xdist_group('h5py')` to the unit tests in `test_postprocessing.py`, in combination with passing the flag `--dist loadgroup` to the `pytest` command. See the [docs](https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus) of the `pytest-xdist` plugin

yguclu: Adding `@pytest.mark.xdist_group('h5py')` defines a group of tests called `h5py`, which has to be executed by a single process sequentially

PR #592, opened 2026-08-11
Vectorized version of topetsc

max-models:
I noticed there are a lot of python loops in the topetsc functions, I think it's possible to improve the performance of these functions. I will try it out in this branch.

### DIFF
--- a/psydac/linalg/tests/bench_topetsc.py
+++ b/psydac/linalg/tests/bench_topetsc.py
@@ -1,0 +1,102 @@
+#---------------------------------------------------------------------------#
+# This file is part of PSYDAC which is released under MIT License. See the  #
+# LICENSE file or go to https://github.com/pyccel/psydac/blob/devel/LICENSE #
+# for full license details.                                                 #
+#---------------------------------------------------------------------------#
+
+import sys
+import time
+
+import numpy as np
+from mpi4py import MPI
+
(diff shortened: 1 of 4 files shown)


yguclu:
> I noticed there are a lot of python loops in the topetsc functions, I think it's possible to improve the performance of these functions. I will try it out in this branch.

I thought those functions were accelerated with Pyccel..

e-moral-sanchez:
Hi, thanks for the PR. The conversion of matrices was pyccelized (within each block, see https://github.com/pyccel/psydac/blob/746d0c9a532575dbd5f432e59de3c9253e1b9f08/psydac/linalg/topetsc.py#L510-L518), but not the conversion of vectors.

yguclu:
Hi @max-models,

Thanks for this PR! What is its current status? Do you want us to review it, or should we wait for the "Ready for review" green light?

BTW, I have just merged `devel` into this branch, as I think it should use the latest version of PETSc.

max-models:
> Hi @max-models,
> 
> Thanks for this PR! What is its current status? Do you want us to review it, or should we wait for the "Ready for review" green light?
> 
> BTW, I have just merged `devel` into this branch, as I think it should use the latest version of PETSc.

Hi, the status is that I unfortunately didn't have time to take a second look at it yet.

Once I have done that, I will mark it as ready for review :)

